PR #20, opened 2026-06-11
Workspace environments: write-only encrypted env vars injected into attached runs

jorgensandhaug:
## Summary

Adds **workspace environments**: named, per-workspace sets of environment variables whose values are secret. Environments attach to sessions (fixed at creation), scheduled tasks, and capability pack installations; values are injected into the sandbox at materialization time only for runs whose session carries an attachment.

### Invariants

- **Write-only values.** No API response, session event, log, span, or audit record ever contains a variable value. Reads return names + metadata (version, timestamps); rotation is `PUT` with a new value.
- **No attachment, no injection.** Unattached runs keep today's behavior byte-for-byte (deployment allowlist + git identity + run-scoped GitHub auth).
- **Agents cannot self-attach.** The worker's first-party MCP token never carries `environments:use`; the MCP scheduled-task tools reject `environmentId` (and instruction edits to environment-attached tasks) without it.
- **Forced RLS isolation** identical to other workspace tables.
- **Encryption at rest**: AES-256-GCM under `OPENGENI_ENVIRONMENTS_ENCRYPTION_KEY` (base64, 32 bytes), held outside Postgres. A DB dump alone does not reveal values.

### What's in here

- Migration `0004_workspace_environments.sql` (additive, idempotent, RLS + grants per existing convention) and drizzle schema: two new tables plus `environment_id` on `sessions` (`SET NULL`) and `scheduled_tasks` (`RESTRICT`).
- REST surface under `/v1/workspaces/:id/environments` with new `environments:manage` / `environments:use` permissions, caps (25 envs/workspace, 100 vars/env), reserved-name policy (platform-managed names and loader-injection vectors), audit events (names only).
- Attachments: `CreateSessionRequest.environmentId`, scheduled task create/update (null detaches; 409 for tasks with a live reusable session), pack enable (`environmentId` validated against the pack's declared `requiredVariables`; template tasks inherit the attachment).
- Worker: values decrypted only in the run activity, layered `allowlist < git identity < workspace env < GitHub run auth`; attached runs fail closed on missing key or deleted environment; echoed values redacted from session events as defense-in-depth (`[redacted:<NAME>]`).
- Deletion semantics: blocked while scheduled tasks or non-terminal sessions reference the environment; idle/terminal sessions detach (documented honestly in `docs/environments.md`).
- Config: `OPENGENI_ENVIRONMENTS_ENCRYPTION_KEY` validated at boot (32 bytes), required for managed mode outside local/test, mirrored in the deployment package's managed env contract; 503 guards when unset elsewhere.
- Docs: new `docs/environments.md` (including the deliberate contrast with the `credentialRef` rule in `docs/packs.md`, the rotation caveat, and residual-exposure notes), `docs/deployment.md`, `.env.example`.

### Tests

- Unit: AES-GCM round-trip/tamper/fail-closed, redactor behavior, reserved-name policy, key decode + managed-mode boot validation, deployment env contract.
- Integration: full CRUD with write-only assertions (raw `value_encrypted` checked, no plaintext in responses/events/audit rows), 503 without key, session/task/pack attachment paths, permission matrix (manage vs use vs neither, including instruction edits on secret-bearing tasks), MCP self-attach rejection with the exact first-party permission set, delete-protection (tasks 409, active sessions 409, idle detach), reusable-session divergence backstop, end-to-end redaction through `runAgentSegment`, fail-closed run without worker key, scheduled dispatch propagation, RLS isolation for both new tables, cross-workspace attachment indistinguishable from missing (422).

### Deployment note

Managed deployments outside local/test must add `OPENGENI_ENVIRONMENTS_ENCRYPTION_KEY` (generate with `openssl rand -base64 32`) to their secret set **before** rolling an image containing this change, or boot validation will fail.

---

> [!NOTE]
> **High Risk**
> Introduces secret storage, decryption in the worker, and broad attachment surfaces (sessions, scheduled tasks, packs); misconfiguration or permission gaps could leak or inject credentials into agent sandboxes.
> 
> **Overview**
> Introduces **workspace environments**: per-workspace named sets of secret env vars (AES-256-GCM in Postgres under `OPENGENI_ENVIRONMENTS_ENCRYPTION_KEY`), exposed via new REST routes and `environments:manage` / `environments:use` permissions. Values are **write-only** in APIs and events; attachment is optional on **session create**, **scheduled tasks**, and **pack enable**, with validation, caps, reserved-name rules, and delete guards when tasks or active sessions still reference an environment.
> 
> The **worker** decrypts values only at sandbox build time (layered under deployment allowlist / git identity, with GitHub run auth last), **redacts** echoed secrets in published session events, and **fails closed** if the key is missing or the environment was deleted. **MCP** scheduled-task tools block agents from attaching environments without `environments:use`. Managed deployments outside local/test must set the new encryption key at boot.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3c384534ed479dd766ad0718cf71aef781c1d96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jorgensandhaug:
Re: the CloudGeni Review finding on `countActiveSessionsUsingEnvironment` missing `idle`:

The specific consequence described — a reusable task re-dispatching its idle session silently without the workspace environment — cannot occur, because the deletion is blocked transitively by the task's own attachment:

1. A reusable session only ever gets its attachment from its task (`dispatchScheduledTaskRun` creates it with `environmentId: task.environmentId`), and the API refuses to change the attachment of a task with a live reusable session (409).
2. So while the task exists, `scheduled_tasks.environment_id` still references the environment, and `DELETE /environments/:id` is rejected by the scheduled-task guard (with `ON DELETE RESTRICT` as the database backstop) regardless of the session being idle.
3. If the task is deleted first, nothing re-dispatches the orphaned idle session; messaging it directly afterwards is the explicitly documented idle-detach case in `docs/environments.md`.

Adding `idle` to the guard would make an environment effectively undeletable after any session ever used it (sessions have no deletion API and remain `idle` indefinitely), which is why the guard deliberately covers only `queued`/`running`/`requires_action` and the idle-detach behavior is documented honestly instead.

Added in 1c9fae9: a regression test asserting the reviewer's exact scenario stays blocked (idle reusable session + attached task → DELETE returns 409 citing the scheduled task), plus a docs clarification of the transitive protection.

PR #66, opened 2026-06-14
Slash-command palette in the chat composer (/clear, /compact, /goal, /help) — rebased on main, click-dispatch fix

jorgensandhaug:
Brings the slash-command palette (originally PR #64, merged into the now-superseded `codex/provider-aware-compaction` stack) cleanly onto current `main`, plus the adversarial-review and Bugbot fixes below. The palette commits rebase onto `main` with no conflicts; the diff is palette-only (the compaction work it stacked on already landed on main via #62/#65).

## What this adds
A filterable slash-command palette in `ChatComposer` (`@opengeni/react`): type `/` for a popover of session/operator controls (`/help`, `/clear-view`, `/goal pause|resume`, `/compact`, `/clear`), with arg hints, keyboard nav, an extensible typed registry, permission gating, and a danger-confirm affordance for destructive commands. Slash commands are session/UI actions, never a structured channel to the agent. Backed by new SDK/API surface for `/clear` (audited context clear) and `/compact` (manual compaction trigger).

## Adversarial-review fixes (assigned finding)

**1. Clicking the harmless `/clear-view` row ran the destructive `/clear`.** The click path routed through `runHighlighted`, whose round-2 exact-match override re-resolved the typed token (`/clear`) to its exact command — so clicking the safe `/clear-view` row dispatched the destructive `/clear` and popped its confirm bar.

Fix: separate explicit selection from token resolution. Extracted the shared autocomplete-or-execute core (`runResolved`) and added `runAt(index)` — the click path runs the clicked row directly, bypassing the exact-match heuristic and the name-only autocomplete step (the operator already chose the row), while still guarding missing required args.

**2. `/clear-view` was unreachable by keyboard once `/clear` was fully typed.** ArrowDown to the clear-view row + Enter still resolved to the destructive `/clear` (exact-match hijack). Fix: track whether the operator has arrow-navigated the highlight for the current draft; when they have, Enter runs the highlighted row directly (like a click). With no navigation the exact-match override is preserved.

## Bugbot review fixes (this PR)

- **Command drafts can't be sent as chat.** A draft that matches a registered command now blocks the composer's send path (Enter handler + send button), even after Escape dismisses the popover, surfacing a nudge instead. Exposed via `isCommandDraft` from `useSlashCommands`. (Fixes "Send bypasses open slash palette" + "Escape then Enter sends slash".)
- **Canceling `/clear` keeps the draft.** The canceled-confirm path returns `{ status: "ok", keepDraft: true }` and `execute` honors `keepDraft`, so the operator who backs out doesn't lose the typed `/clear`. (Fixes "Canceling clear wipes draft".)
- **Case-insensitive Enter resolution.** The exact-match override and `runResolved`'s name check lowercase the token (matching `matchCommand`), so `/Clear` resolves to the destructive clear, not the highlighted prefix near-match. (Fixes "Enter exact match case sensitive".)
- **`/clear-view` latches on an empty stream.** The view-cleared marker is `number | null` instead of a `0` sentinel, so clearing with no events still suppresses the initial-message fallback and hides later events. (Fixes "Clear-view no-op at zero".)

## Gate
`bun run typecheck` (all packages), `bun run test:unit` (575 pass / 0 fail), SDK/react/demo/web builds — all green on this branch against current main. gitleaks clean across all commits.

### DIFF
--- a/apps/api/src/routes/sessions.ts
+++ b/apps/api/src/routes/sessions.ts
@@ -1,17 +1,22 @@
 import {
+  ClearSessionContextRequest,
   ClientSessionEvent,
+  CompactSessionContextRequest,
   ReorderSessionTurnsRequest,
   UpdateSessionGoalRequest,
   UpdateSessionTurnRequest,
 } from "@opengeni/contracts";
+import { resolveContextCompactionMode } from "@opengeni/config";
 import {
   cancelQueuedSessionTurn,
+  clearSessionContext,
   getSession,
   getSessionGoal,
   listSessionEvents,
   listSessions,
   listSessionTurns,
   reorderQueuedSessionTurns,
+  requestSessionCompaction,
   requireSession,
   setSessionGoalStatus,
   updateQueuedSessionTurn,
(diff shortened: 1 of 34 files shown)


jorgensandhaug:
Addressed all 5 Bugbot findings in commit 7c72da4 (force-pushed):

- **Send bypasses open slash palette** / **Escape then Enter sends slash** — the composer now blocks its send path (Enter handler + send button) whenever the draft matches a registered command, even after Escape; `useSlashCommands` exposes `isCommandDraft` for this. A nudge is surfaced instead of delivering the command as chat.
- **Canceling clear wipes draft** — the canceled `/clear` confirm returns `{ status: "ok", keepDraft: true }` and `execute` honors `keepDraft`, so the `/clear` draft survives a cancel.
- **Enter exact match case sensitive** — the exact-match override and `runResolved` now lowercase the token (matching `matchCommand`/`filterCommands`), so `/Clear` resolves to the destructive command.
- **Clear-view no-op at zero** — the view-cleared marker is now `number | null`, so clearing an empty stream still latches (suppresses the initial-message fallback and hides later events).

Each is covered by a test; full gate is green (typecheck, 575 unit tests, SDK/react/demo/web builds), gitleaks clean.

PR #72, opened 2026-06-14
Session workflow: continueAsNew to bound Temporal history

jorgensandhaug:
## Problem (CRITICAL)

The session Temporal workflow ran its `while (true)` turn loop on a **single workflow run** for the life of the session. Every signal, activity schedule/complete, and timer appends to that run's Temporal **event history**, which grows without bound. Temporal force-terminates a run at its hard history limit (~51,200 events / 50MB), so a long-lived session — a weeks-long manager goal — was **guaranteed to be killed** once history overflowed.

Verified at `apps/worker/src/workflows/session.ts` (the `while (true)` loop), origin/main base `2131583`.

## Fix

`continueAsNew` the session workflow at a **safe boundary**: the top of the loop, **before** `claimNextQueuedTurn`. This is the only point where no turn is mid-flight (every path that reaches a new iteration completed or re-queued its turn first), the approval queue has been drained inside `runTurn`, and a pending interrupt can be checked and deferred.

- **Trigger:** `workflowInfo().continueAsNewSuggested` (the real-world primary — the server raises it well before the hard cap) **OR** a deterministic turn-counter backstop (`TURNS_PER_RUN_BACKSTOP = 2000`, conservatively low vs the event budget) that guarantees a CAN even if the suggestion never arrives.
- **Guard:** only continueAsNew when `interruptedEventId === null` **and** `approvalQueue.length === 0`. A pending interrupt / buffered approval is unbacked in-memory state the new run could not reconstruct, so the boundary defers to the loop to handle it instead of losing it.
- **Determinism:** the whole block is gated by `patched("session-continue-as-new")`, so in-flight multi-day histories recorded before this branch existed never emit the new command on replay and stay deterministic. The workflowId is already deterministic (`session-<sessionId>`).

## No-strand guarantee

`continueAsNew` carries only the self-contained `SessionWorkflowInput` (`accountId`, `workspaceId`, `sessionId`, and `maxTurnsPerRun` when set) — deliberately **without** `initialEventId`. The fresh run claims from the **durable Postgres queue** (`session_turns where status='queued'`, `FOR UPDATE SKIP LOCKED`), it does not replay a seed event:

- A **queued turn** survives because it lives in Postgres.
- A buffered `userMessage` / `queueChanged` signal only bumps an in-memory `wakeups` counter; its turn was written to Postgres **before** the signal was sent, so the fresh run re-claims it. Losing the counter strands nothing.

## Tests (`test/integration/temporal-workflow.integration.ts`)

- **continues-as-new at the turn boundary, carrying state and stranding no queued turn** — two turns sit in the durable queue, backstop is 1, so the workflow continues-as-new after each turn; the **second** turn can only be dispatched by the **second** run. Asserts the recorded `WorkflowExecutionContinuedAsNew` event carries the self-contained input forward with **no `initialEventId`**.
- **a queueChanged signal buffered at the continueAsNew boundary is not stranded** — the follow-up turn is enqueued and a `queueChanged` signal sent only after the first turn has run (i.e. while the workflow is poised to continue-as-new); the continued run must still dispatch it from the durable queue even though the in-memory wakeup counter is dropped.

Both continueAsNew tests use a generous 120s bound (they legitimately span a continueAsNew chain plus a possible 5s idle re-claim window) so they are not flaky under CI load when run last in the suite.

## Verification

- Full repo `typecheck` GREEN.
- Worker unit tests GREEN.
- `temporal-workflow.integration.ts` GREEN: 19 tests, 0 fail (incl. both continueAsNew tests; pre-existing interrupt / preempt / worker-death / goal-continuation tests all still pass).
- gitleaks clean; no customer/staging specifics in code, tests, or this PR.

Based on latest `origin/main` (`2131583`). Does not touch the same lines/functions as the concurrent manager-robustness work (`last_input_tokens` trigger in `agent-turn.ts`, `context-compaction.ts`, the `user.interrupt` path, the `session_events` MCP tool).

---

> [!NOTE]
> **High Risk**
> Changes core long-running session orchestration; incorrect boundary handling could strand turns or wedge sessions, though durable-queue design and extensive integration tests mitigate that.
> 
> **Overview**
> Long-lived **session** Temporal workflows could hit Temporal’s hard event-history limit and be force-terminated. This PR **rolls the workflow forward with `continueAsNew`** at the **top of the turn loop** (before `claimNextQueuedTurn`), gated by `patched("session-continue-as-new")` for replay safety.
> 
> **Triggers:** `workflowInfo().continueAsNewSuggested` or a per-run turn counter backstop (**2,000** turns in prod; optional `maxTurnsPerRun` for tests). **Defer** continue when a pending **interrupt** is in memory. **Clear stale `approvalDecision` entries** at the boundary (do not require an empty approval queue—that would wedge the guard). The new run receives only `accountId` / `workspaceId` / `sessionId` (no `initialEventId`); work is reclaimed from the **Postgres turn queue**.
> 
> Integration tests cover multi-run turn draining, `queueChanged` at the boundary, and stale approvals; continueAsNew tests use a **120s** timeout to reduce CI flakiness.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ecdb67e8f8d7a6a2e4154e447af49ec9e5c5127. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/worker/src/workflows/session.ts
+++ b/apps/worker/src/workflows/session.ts
@@ -82,23 +82,38 @@ export async function sessionWorkflow(input: SessionWorkflowInput): Promise<void
   while (true) {
     // History-overflow guard. The top of the loop is the only safe
     // continueAsNew boundary: no turn is mid-flight (every path that reaches a
-    // new iteration completed or re-queued its turn first), the approval queue
-    // is drained inside runTurn before it returns, and a pending interrupt is
-    // unbacked in-memory state that the new run could not reconstruct — so we
-    // refuse to continueAsNew while one is set and let the loop handle it
-    // first. A buffered userMessage/queueChanged signal only bumps `wakeups`,
-    // and its turn was written to Postgres BEFORE the signal was sent, so the
-    // fresh run re-claims it on its first claimNextQueuedTurn — losing the
-    // counter strands nothing. The queue living in Postgres is the safety net:
-    // continueAsNew carries only the self-contained SessionWorkflowInput (no
-    // initialEventId — the new run claims from the queue, it does not replay a
-    // seed event). patched() keeps in-flight histories (recorded before this
-    // branch existed) deterministic: they never see the new command.
+    // new iteration completed or re-queued its turn first), and a pending
+    // interrupt is unbacked in-memory state that the new run could not
+    // reconstruct — so we refuse to continueAsNew while one is set and let the
+    // loop handle it first. A buffered userMessage/queueChanged signal only
+    // bumps `wakeups`, and its turn was written to Postgres BEFORE the signal
+    // was sent, so the fresh run re-claims it on its first claimNextQueuedTurn
+    // — losing the counter strands nothing. The queue living in Postgres is the
+    // safety net: continueAsNew carries only the self-contained
+    // SessionWorkflowInput (no initialEventId — the new run claims from the
+    // queue, it does not replay a seed event). patched() keeps in-flight
+    // histories (recorded before this branch existed) deterministic: they never
+    // see the new command.
+    //
+    // The approval queue is NOT a boundary condition, and is cleared here. A
+    // genuinely-pending approval keeps the workflow blocked INSIDE runTurn (the
+    // `await condition(...)` in the requires_action branch), so it never
+    // reaches the top of the loop — meaning every approvalQueue entry observed
+    // here is necessarily STALE (an approvalDecision signal for a turn that
+    // already settled; the API guard only checks status==='requires_action', so
+    // two decisions submitted while requires_action both land in the queue and
+    // the surplus is left behind once the turn completes without re-blocking).
+    // Coupling continueAsNew to `approvalQueue.length === 0` would let one such
+    // stale entry wedge the guard forever, re-introducing the exact
+    // history-overflow termination this branch prevents. Dropping the surplus
+    // is safe: a real pending approval also leaves the session in
+    // requires_action with the turn re-dispatchable from Postgres.
     if (patched("session-continue-as-new")) {
       const info = workflowInfo();
       const maxTurnsPerRun = input.maxTurnsPerRun ?? TURNS_PER_RUN_BACKSTOP;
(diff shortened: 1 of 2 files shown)


jorgensandhaug:
Addressed the review finding **"continueAsNew can be permanently wedged by a stale approval in approvalQueue"** (major).

Verified valid against the code: the boundary guard `if (shouldContinue && interruptedEventId === null && approvalQueue.length === 0)` is only reached after `runTurn` returns, and `runTurn` never returns while legitimately awaiting an approval (it blocks at the `requires_action` `await condition(...)`). So any `approvalQueue` entry observed at the boundary is necessarily **stale**. The API guard at `apps/api/src/routes/sessions.ts` only checks `status === 'requires_action'`, so two `approvalDecision` signals submitted while the session is `requires_action` both signal and both land in `approvalQueue` (the handler pushes unconditionally); the first re-runs the turn to completion and the second is orphaned. The queue is never drained elsewhere, so `approvalQueue.length === 0` stays false forever and `continueAsNew` never fires — the session grows to the Temporal ~51,200-event hard cap and is force-terminated, the exact failure this PR exists to prevent.

**Fix** (commit `2ecdb67`): dropped `&& approvalQueue.length === 0` from the guard and clear the queue (`approvalQueue.length = 0;`) at the boundary before `continueAsNew`. Kept the `interruptedEventId === null` gate (a pending interrupt is real in-memory-only state). Safe because a genuinely-pending approval keeps the workflow blocked inside `runTurn` and never reaches the boundary, and a real pending approval also leaves the session in `requires_action` with the turn re-dispatchable from Postgres.

**Test added:** `a stale approval left in the queue does not wedge the continueAsNew boundary` — enqueues a stale approval, proves `continueAsNew` still fires at the boundary and the continued run dispatches the follow-up turn (no stranding). Confirmed it fails against the pre-fix guard (`continuedEvent` is `undefined`) and passes with the fix.

Rebased on latest `origin/main`. Full local gate green: monorepo typecheck, 609 unit tests, and the full `temporal-workflow.integration.ts` suite (20 pass).

PR #319, opened 2026-07-08
fix(worker): re-dispatch a turn the dying attempt stamped `cancelled`

jorgensandhaug:
## Problem
A worker that dies mid-turn (activity **heartbeat timeout**) can have its zombie attempt mark the turn `cancelled` in the generic `CancelledFailure` cleanup. `requeueTurnAfterWorkerDeath` then saw a terminal turn and returned `{action:"stale"}` — **dropping it**. If a goal was `paused` awaiting that worker, the session went idle and **stalled indefinitely** until a human hit "continue".

Real incident: geni/Vern manager session `76e2f2ee`, ~16h stall on 2026-07-07→08. Durable proof: Temporal run `117068a9` history shows `ACTIVITY_TASK_TIMED_OUT{HEARTBEAT}` → `requeueTurnAfterWorkerDeath` → `{action:"stale"}` → `markSessionIdle`.

## Why the fix is safe
Reaching `requeueTurnAfterWorkerDeath` **proves** the workflow classified the failure as a worker death (heartbeat / schedule-to-start timeout). A deliberate user interrupt resolves the workflow's *interrupt* branch and runs `interruptActiveTurn` — it never reaches here. So a `cancelled` turn at this point is unambiguously the zombie's **death artifact**, not a real settle → re-dispatch it. A genuinely settled `completed`/`failed` turn still stays `stale`. Bounded by the existing `WORKER_DEATH_MAX_REDISPATCHES` cap (deterministic re-death → `failSession`, a loud escalation, not a silent stall).

## Changes
- **session-state** `requeueTurnAfterWorkerDeath`: treat a death-artifact `cancelled` turn as re-dispatchable.
- **db** `requeuePreemptedTurn`: optional `fromStatuses` so the reset can match a `cancelled` row (default unchanged for every existing caller).
- **activities** `heartbeatTimeout` **30s → 2min**: the 10s heartbeat timer shares the turn's event loop and the SDK throttles forwarded heartbeats to ~80% of timeout, so 30s left only ~6s slack — a GC pause / large-context request assembly tripped **false** worker-deaths. A real dead worker is still caught within 2 min (immaterial for turns that run for days).

## Tests
`session-state-worker-death.test.ts`: re-dispatches a `cancelled` turn (with cancelled-inclusive `fromStatuses` + `turn.preempted{worker_death}`), keeps `completed`/`failed`/missing as `stale`. Typecheck + adjacent worker unit suites green.

## Follow-ups (not in this PR)
- Un-starvable heartbeat (emit off the work event loop) + context-size discipline (root of the false-death).
- Paused-goal reconciler (weaker second net; this race fix already covers the incident).

https://claude.ai/code/session_01B4tzoxgyduFnVme3sAtceZ

---

> [!NOTE]
> **Medium Risk**
> Changes turn lifecycle, worker-death recovery, and Temporal heartbeat semantics; incorrect fencing could still orphan turns or clobber re-queued work, though behavior is bounded and covered by new tests.
> 
> **Overview**
> Fixes **indefinite session stalls** when a worker dies mid-turn: the zombie’s `CancelledFailure` cleanup could mark the turn `cancelled`, and worker-death recovery then treated that as a real terminal outcome (`stale`) instead of re-queuing.
> 
> **Worker-death recovery** (`requeueTurnAfterWorkerDeath`) now treats `cancelled` as a death artifact (only on this path—user interrupts never reach it), requeues via `requeuePreemptedTurn` with `fromStatuses` including `cancelled`, and **rethrows** on persistence errors while the turn is still requeueable instead of silently going stale.
> 
> **Zombie vs recovery races**: `cancelTurnFromDyingDispatch` fences dying-attempt cancels on unchanged `workerDeathRedispatches` and live turn status; `runAgentTurn` records the counter at dispatch and uses this instead of unconditionally `finishTurn` on generic `CancelledFailure`.
> 
> **False worker deaths**: agent activity `heartbeatTimeout` is **30s → 2 minutes** to absorb GC / large-context assembly on the shared event loop.
> 
> Regression tests cover death-artifact requeue and DB fencing orderings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5db3a2a05d67beaf3ada18d345e20e92ab399c67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/worker/src/activities/agent-turn.ts
+++ b/apps/worker/src/activities/agent-turn.ts
@@ -3,6 +3,7 @@ import {
   createTurn,
   applyCreditDebitUpToBalance,
   finishTurn,
+  cancelTurnFromDyingDispatch,
   getBillingBalance,
   getSandbox,
   readActiveSandbox,
@@ -753,6 +754,11 @@ export function createRunAgentTurnActivity(services: () => Promise<ActivityServi
     let turnMetricOutcome: TurnOutcome | null = null;
     let activityError: unknown;
     let turnId: string | undefined;
+    // Worker-death redispatch counter observed when THIS dispatch claimed the
+    // turn. If a dying-attempt cancel later fences on this and the turn's
+    // current value differs, recovery already re-queued/re-dispatched the turn
+    // and the zombie must not clobber it.
+    let redispatchesAtDispatch = 0;
     let heartbeatTimer: ReturnType<typeof startActivityHeartbeat> | undefined;
     // P1.2 ownership inversion: when sandboxOwnershipEnabled, the turn resolves
     // the one box by id from the group lease and injects it NON-OWNED into the
@@ -1072,6 +1078,7 @@ export function createRunAgentTurnActivity(services: () => Promise<ActivityServi
         throw new Error(`Session turn not found for trigger: ${input.triggerEventId}`);
       }
       turnId = turn.id;
+      redispatchesAtDispatch = Number((turn.metadata as { workerDeathRedispatches?: number } | null)?.workerDeathRedispatches ?? 0);
       turnLifecycleMetricsFor(observability).start(turnId);
       // Canonical codex-billed predicate (codex/<slug> + feature enabled + active
       // workspace credential). Computed once and threaded through every billing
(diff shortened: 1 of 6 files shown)


jorgensandhaug:
Fixed in `5db3a2a`. Bugbot's catch is right: the error-recheck still classified `cancelled` as settled, so a *real* persistence error during `requeuePreemptedTurn` on a still-`cancelled` row would return `stale` and drop the turn — the same idle-stall via the error path. The recheck now treats running/requires_action/cancelled as still-re-dispatchable (rethrow to retry) and only genuinely-settled states as stale, with unit tests for both branches.

PR #357, opened 2026-07-10
perf(codex): honest per-model context window + compact at 90% of it

jorgensandhaug:
## The most consequential wrong number in the system

Proactive context compaction has **never fired** for codex-subscription turns. The trigger is `ratio × contextWindowTokens`, and the codex models inherited the 1.05M global default window → trigger ≈ 645K. But the subscription's gpt-5.6 models hard-reject with `context_length_exceeded` at **~334–348K observed input tokens** (prod, 2026-07-10). Result: histories grew straight to the reject cliff — slow prefills on every call, worker memory pressure, and reject/retry loops on long autonomous sessions. This is a direct contributor to the token-streaming sluggishness under fleet load.

## Fix — two honest numbers

1. **Honest window**: codex provider models declare `CODEX_MODEL_CONTEXT_WINDOW_TOKENS = 320_000` (just under the measured cliff), and the worker budgets client-path compaction against the **resolved model's** window instead of the global default (mirroring what the SDK path already did).
2. **Late trigger**: default compaction threshold ratio moves **0.6 → 0.9** in both places it's defined (config zod default + runtime fallback). Retained context beats early headroom now that the window base is honest; the fail-closed reactive compact-on-reject ladder absorbs any overshoot as one retried call, never a dead session.

Net for gpt-5.6: trigger moves from 645K (unreachable) to **288K**, with ~46K margin below the cliff for tokenizer-estimate skew.

## Verification

- `codex-overlay` worker tests updated + passing (11/11): overlay models carry the declared window; compaction settings resolve per-model.
- Full `packages/config` + `packages/runtime` suites: 730 pass.
- Typecheck clean on worker/config/runtime.
- No deployment pins `OPENGENI_COMPACTION_THRESHOLD_RATIO`, so the new default takes effect on deploy.

https://claude.ai/code/session_011YmyyQ8HaSHarkfteTwcQq

### DIFF
--- a/.changeset/codex-context-window-honest.md
+++ b/.changeset/codex-context-window-honest.md
@@ -1,0 +1,7 @@
+---
+"@opengeni/codex": patch
+"@opengeni/config": patch
+"@opengeni/runtime": patch
+---
+
+Proactive context compaction now actually fires on the codex-subscription path: codex models declare their real (empirically measured) context window instead of inheriting the 1.05M global default, and the default compaction trigger moves from 60% to 90% of the declared window — compact as late as possible now that the window base is honest, with the reactive compact-on-reject ladder absorbing any overshoot.
--- a/apps/worker/src/activities/agent-turn.ts
+++ b/apps/worker/src/activities/agent-turn.ts
@@ -2383,6 +2383,13 @@ export function createRunAgentTurnActivity(services: () => Promise<ActivityServi
                 promptCacheKey,
               })
           : undefined;
+      // Client-path compaction must budget against the RESOLVED model's real
+      // context window (the codex subscription window is far smaller than the
+      // 1.05M global default), mirroring the SDK path above. Without this the
+      // threshold (window * ratio) uses 1.05M and proactive compaction never
+      // fires for codex turns — histories grow to the ~340k model cliff.
+      const compactionContextWindowTokens =
+        resolvedModel?.configured.contextWindowTokens ?? runSettings.contextWindowTokens;
       // Pre-turn client-side context compaction (Azure path). When the
       // resolved mode is "client" and the single Codex-parity threshold is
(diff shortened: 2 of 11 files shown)


jorgensandhaug:
**OPE-12 context-policy BLOCK on exact head `08f4ce7b9f3ffbf5bf9cb70263b4af630e0b0c74`: the new 90% trigger crosses the declared usable-input budget.**

This is separate from the already-posted stale-`testSettings` finding. `clientCompactionThresholdTokens` accepts `contextReservedOutputTokens` but does not use it (`packages/runtime/src/context-compaction.ts:114-122`). With the repository default reserve of 128,000 tokens:

- Codex: 320,000 total window − 128,000 reserve = **192,000 usable input**, but the proposed 90% trigger is **288,000**.
- Built-in 1,050,000 window − 128,000 reserve = **922,000 usable input**, but the proposed 90% trigger is **945,000**.

So the global 0.6→0.9 change can defer proactive compaction until after the policy’s own input budget, even before accounting for estimation skew. The current 0.6 Codex result happens to equal the 192k usable budget; that is why the stale test’s `<230k` assertion contradicts the new behavior.

Please keep the per-model Codex window fix, but either retain the existing threshold default or define/test the trigger against the usable-input budget (and provider-evidenced output limits) rather than raw total context. Add an exact production-default test that does not inherit `testSettings`’ pinned 0.6. OPE-12 will freeze window, input budget, and compaction threshold independently per accepted turn; this prerequisite must not establish a contradictory global default.

jorgensandhaug:
Bugbot's catch was real and is fixed in 03de5883: the shared testing fixture pinned the old 0.6 ratio, so the overlay test validated a 192K trigger while production would run at 288K. The sweep also caught one more derived assertion (capabilities server-path compact_threshold, 630K → 945K). Full affected suites re-run green: 911 tests, 0 fail.

PR #362, opened 2026-07-10
perf: make the dev loop fast (tests 7-10x, typecheck 4.3x)

jorgensandhaug:
## Make the dev loop fast

The OpenGeni test/typecheck loop was dominated by two Docker-gated costs that silently ran on every `bun test` (and every PR CI unit run) whenever Docker was present:

1. **`desktop-image.test.ts` built the full desktop OCI image** (Ubuntu + apt xvfb/chrome + noVNC clones) — minutes per run.
2. **Every DB-backed test file replayed the full 55-migration chain** into its own fresh database (~29 such files across `packages/db`, `apps/api`, `apps/worker`).

Every other worker test file runs in under a second. This PR removes both costs and parallelizes typecheck.

### Before / after (same host, back-to-back, load-inflated but ratios hold)

| Lane | Before | After | Speedup |
|---|---|---|---|
| `bun test apps/worker` | 296s | 28s | **10.7x** |
| `bun test apps/worker packages/runtime packages/config packages/testing` | 245s | 34s | **7.2x** |
| `bun run typecheck` | 19s | 4.4s | **4.3x** |

Rebased-tree re-verification (on `0f104135`): the deliverable command is **1050 pass / 0 fail across 61 files**, typecheck **4.77s** all clean, desktop-image still excluded from `bun test`, template harness active.

### Commits (one lever each, before/after in each message)

1. **perf(typecheck)** — run the 20 independent tsgo projects through a bounded parallel pool (was strictly serial). Failures are collected and all reported instead of dying on the first. `OPENGENI_TYPECHECK_CONCURRENCY` overrides.
2. **perf(testing)** — the shared-pg harness migrates **once** into a template database (`og_test_template`, lock-guarded, `datistemplate=true` sentinel, self-healing) and each test file's DB is a `CREATE DATABASE ... TEMPLATE` clone (~100ms) instead of replaying 55 migrations. Same schema, same per-file isolation, same grants; benefits every DB-backed test file.
3. **perf(worker-tests)** — rename `desktop-image.test.ts` → `.e2e.ts` (100% pure rename; the `.e2e.ts` suffix is excluded from the default `bun test` by repo convention), wired into `test:e2e`.
4. **perf(ci)** — cache `~/.bun/install/cache` in the test + deployment jobs (`setup-bun` doesn't cache).
5. **ci(desktop)** — a dedicated `desktop-e2e` workflow that runs the desktop-image e2e on PRs touching `docker/desktop*` / the e2e file, plus a weekly backstop — so the desktop image build still runs automatically, just not on every unrelated PR.

The whole functional diff is infrastructure-only: no test body or assertion is modified, and desktop-image is a 100% rename — every test still fails for exactly the bugs it caught before.

### Note: full-repo `bun test` under a saturated host

Running the entire repo (all 174 files) on a heavily-loaded dev box shows ~30 `apps/api` DB-write 500s (`failed to append initial user event`) — the throwaway Postgres getting starved. A controlled A/B on the identical multi-package subset proves this is **pre-existing host contention, not the template change**: old migrate-per-file harness = 31 fail / 498s, new template-clone = 30 fail / 269s (old fails *more* and is 1.85x slower). It vanishes running `apps/api` alone (379 / 0), the 4-target set (1050 / 0), or the two files in isolation. CI's dedicated runner won't hit it. Follow-up hardening (not in this PR): bound the per-file connection pools / cap DB-file concurrency in the shared-pg harness.

https://claude.ai/code/session_011YmyyQ8HaSHarkfteTwcQq


jorgensandhaug:
### Same-load before/after on the rebased tree

Ran the deliverable command back-to-back on `origin/main` (`0f104135`) vs this branch, same load window, so the ratio is apples-to-apples on current main:

| `bun test apps/worker packages/runtime packages/config packages/testing` | wall | tests |
|---|---|---|
| **before** (origin/main: old migrate-per-file harness + desktop image build) | 315.3s | 1052 across 62 files |
| **after** (this branch) | **28.9s** | 1050 pass / 0 fail across 61 files |

**≈10.9x**, no regressions. The two fewer tests / one fewer file are the desktop-image case that moved to the `test:e2e` lane (now covered by the new `desktop-e2e` workflow). Numbers are load-dependent on this shared host, but the pair is contemporaneous so the ratio holds.

PR #368, opened 2026-07-10
fix(worker): stop child completions enqueueing parent chat work

jorgensandhaug:
## Summary
- default `OPENGENI_CHILD_COMPLETION_PARENT_WAKE_ENABLED` to `false`
- return before every parent DB read/write, event publish, or workflow signal when disabled
- preserve child session status, durable events, goal/evidence, and the explicitly enabled legacy wake path
- document the temporary compatibility flag and lifecycle behavior

## Verification
- `bun run typecheck` — all 20 projects clean
- `bun run lint` — 0 errors (existing warnings only)
- `bun test packages/config/test/config.test.ts apps/worker/test` — 357 pass, 0 fail
- `bun test apps/worker/test/parent-wake.test.ts` — 1 pass, proves no DB/bus/workflow/observability side effect on the disabled path
- `bun run format:check` — pass
- `bun run check:docs-refs` — pass
- focused real-Postgres/NATS integration added to `worker-activity.integration.ts`; local execution was infrastructure-blocked before test startup because this old gVisor sandbox cannot `unshare` Docker image layers, so exact-head CI must pass it before merge

## Release contract
Merge only the exact independently reviewed head with green CI, then build/deploy staging and promote the identical artifact to production. Production acceptance must show a child completion remains durable on the child without adding a parent `user.message`, queued turn, or workflow wake; a normal human parent message must still queue normally.


jorgensandhaug:
INDEPENDENT EXACT-HEAD REVIEW — BLOCK

Reviewed exact SHA `0cb766eba06676b55e830b9347bab9d647931ed1` (checkout, fetched `refs/pull/368/head`, live remote ref, and GitHub `headRefOid` all matched).

The code-side gate is correctly the first executable statement in `notifyParentOfChildTerminal`: default-disabled calls return before parent DB reads/writes, events/turns, bus publish, workflow signal, or observability; all callers settle the child status/events first, goal evidence remains in the child goal row, ordinary human-message admission is untouched, and the explicit `true` legacy path retains broad integration coverage.

**Blocking finding:** `docs/manager-session-robustness.md:142-154` is current-tier documentation but still says #60 auto-wakes parents and directs managers to “go idle and await the wakeup.” With the new default `false`, that guidance can strand a manager indefinitely and directly contradicts the updated lifecycle/architecture docs. Update it to describe default-disabled behavior, explicit compatibility opt-in, and a safe non-wakeup coordination strategy.

Checks: `bun run typecheck` (20/20 projects); focused Bun tests (61 pass, 0 fail); changed-file format check; docs-reference guard; `git diff --check`; targeted lint (0 errors, 10 warnings on unchanged lines). Worker integration was inspected but not executed because this sandbox has no Docker daemon (`/var/run/docker.sock` absent).

jorgensandhaug:
INDEPENDENT EXACT-HEAD REVIEW — APPROVE

Renewed review of exact `b62a33e322ea8f6f314bfe63d80c79d48816caf5`: live PR metadata/ref, fetched ref, and detached checkout all match. I re-reviewed the full original base `1a841303…` → head range; the only follow-up is docs commit `b62a33e…` changing `docs/manager-session-robustness.md`.

The corrected lines 139–158 now match the default-disabled contract: child status/events/goal evidence remain durable, no synthetic parent message/turn is promised, legacy wake behavior requires explicit opt-in, and managers are directed to durable external or bounded one-shot scheduled/event-driven reconciliation—not polling or idle-and-assume coordination. The original code findings remain sound: disabled mode returns before parent DB/event/turn/bus/workflow effects; human messages and explicit legacy behavior are unchanged and covered.

Current `main` drift (`8fef5000…`, observability only) has no changed-path overlap; merge-tree is clean and the refreshed synthetic merge ref has parents current main + this exact head. Checks: exact-head CI 3/3 green (typecheck/unit, deployment artifacts, workload images); docs-reference guard and `git diff --check` pass; correction confirmed docs-only and worktree pristine.

PR #405, opened 2026-07-15
fix(db): stream cutover proofs via COPY

jorgensandhaug:
## Summary
- replace production relation-proof row decoding with a backpressured PostgreSQL COPY byte stream
- hash only canonical, content-free proof fields without materializing JavaScript row objects
- advance the cutover proof contract to v3

## Evidence
- `bunx tsc -p packages/db/tsconfig.json --noEmit`
- `OPENGENI_REQUIRE_REAL_DB=1 bun test packages/db/test/session-control-cutover-audit.test.ts`
- `bunx oxlint packages/db/src/session-control-cutover-audit.ts packages/db/test/session-control-cutover-audit.test.ts`
- `ubs --diff --skip=5,7 --fail-on-warning` (clean; skipped categories are unchanged false positives for non-secret equality checks and pre-existing narrowing heuristics)

## Production failure addressed
The v7 disposable PITR audit remained read-only and failed closed at the 1 GiB pod limit while decoding the 5,040,146-row event relation. Explicit GC at every cursor batch did not bound the Bun/postgres.js decoded-row path. COPY removes that decoding/ownership path entirely while preserving ordered SHA-256 relation proofs.

### DIFF
--- a/packages/db/src/session-control-cutover-audit.ts
+++ b/packages/db/src/session-control-cutover-audit.ts
@@ -1,7 +1,7 @@
 import { createHash } from "node:crypto";
 import type postgres from "postgres";
 
-export const SESSION_CONTROL_CUTOVER_CONTRACT = "opengeni/session-control-cutover/v2" as const;
+export const SESSION_CONTROL_CUTOVER_CONTRACT = "opengeni/session-control-cutover/v3" as const;
 
 export type CutoverPhase = "baseline" | "migrated" | "final";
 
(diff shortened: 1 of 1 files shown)


jorgensandhaug:
Additional deterministic memory reproduction after locating the retained-substring ownership bug:

- Same PostgreSQL COPY shape and exactly **5,040,146 rows** as production cardinality
- Before flattening durable map keys: **947 MiB observed peak RSS**
- After flattening the one durable key per session: **122 MiB observed peak RSS**, 5,040,146/5,040,146 rows hashed
- Mandatory real-PostgreSQL migration/reconciliation test rerun green after the fix

The old cursor and initial COPY implementations both retained backing database chunks through session-id substring map keys. The current head copies only one independent key per session and leaves all streamed substrings transient.

PR #512, opened 2026-07-20
fix(runtime): bound display stack startup

jorgensandhaug:
## Summary

- poll yielded sandbox exec sessions through `writeStdin` until real process completion instead of treating the provider output-yield window as a deadline
- apply both a JavaScript wall deadline and an independent in-box GNU `timeout` that terminates the full lock/waiter process group
- supervise outer and canonical launcher locks with command-mode `flock --close` so detached Xvfb/XFCE/x11vnc/websockify processes cannot retain startup lock descriptors
- add bounded, secret-safe stage/caller/provider telemetry while preserving lazy desktop startup and the existing paint-readiness gate

## Root cause

Production `ensureDisplayStack` calls taking 202–777 seconds were not slow Modal sandbox creation. Modal's `yieldTimeMs` only bounds how long exec waits before returning output; it can return `Process running with session ID N` while the process continues. The runtime neither polled that session nor enforced a real wall/in-box deadline. Separately, file-descriptor inspection proved detached display descendants inherited the startup lock, allowing later callers to remain serialized after the launcher exited.

## Validation

- `bun test packages/runtime/test/display-stack.test.ts apps/worker/test/display-stack-gating.test.ts --timeout 5000` — 21 pass, 0 fail
- `bun test apps/api/test/sandbox-desktop-stream.test.ts --timeout 30000` — 8 pass, 1 explicitly gated live-Modal skip, 0 fail
- `bun run typecheck` — 20 projects clean
- `bun run lint` — 0 warnings/errors across 825 files
- touched-file `oxfmt --check`, `bash -n docker/desktop/opengeni-desktop-up.sh`, `git diff --check`, and launcher mode check — clean
- current-main compatibility — disjoint path audit plus conflict-free temporary-index synthetic merge

Disposable live Modal canary through the real `ModalSandboxClient`:

- two concurrent cold calls completed in 6,984 ms with one launcher and two paint-ready results
- already-ready follow-up completed in 580 ms
- inherited startup-lock FDs: 0
- forced deadline remained typed as `timeout`, launched no delayed process, and both locks were reacquirable afterward
- X display, :5900 RFB, :6080 noVNC, XTEST mouse movement, framebuffer capture, and browser noVNC connection all verified
- browser status: `Connected (encrypted) to modal:0`

OPE-46 is not deployed or production-verified and remains In Progress.

### DIFF
--- a/.changeset/calm-desktops-start.md
+++ b/.changeset/calm-desktops-start.md
@@ -3,4 +3,4 @@
 "@opengeni/api-router": patch
 ---
 
-Bound Modal display startup ownership, poll yielded provider processes to completion, and prevent detached desktop processes from retaining startup locks.
+Bound Modal display startup ownership, parse terminal state only from trusted provider metadata, poll yielded processes to completion, and prevent detached desktop processes from retaining startup locks.
--- a/.changeset/calm-desktops-start.md
+++ b/.changeset/calm-desktops-start.md
@@ -1,0 +1,6 @@
+---
+"@opengeni/runtime": patch
+"@opengeni/api-router": patch
+---
+
+Bound Modal display startup ownership, poll yielded provider processes to completion, and prevent detached desktop processes from retaining startup locks.
--- a/apps/api/src/sandbox/viewer.ts
+++ b/apps/api/src/sandbox/viewer.ts
@@ -648,7 +648,13 @@ export async function mintDesktopStream(
     // Idempotent display stack (flock-guarded; a no-op when already up). A box
     // that genuinely can't run the stack degrades to transport:null, not a throw.
     try {
-      await ensureDisplayStack(established.session);
+      await ensureDisplayStack(established.session, {
+        telemetryContext: {
+          callerKind: "viewer",
+          ...(lease.instanceId ? { sandboxId: lease.instanceId } : {}),
+          leaseEpoch: lease.leaseEpoch,
+        },
+      });
     } catch (error) {
       if (error instanceof DisplayStackUnsupportedError) {
         return null;
--- a/docker/desktop/opengeni-desktop-up.sh
+++ b/docker/desktop/opengeni-desktop-up.sh
@@ -10,6 +10,13 @@ W="${DESKTOP_W:-1280}"; H="${DESKTOP_H:-800}"; DPI="${DESKTOP_DPI:-96}"
 PORT="${STREAM_PORT:-${OPENGENI_DESKTOP_STREAM_PORT:-6080}}"
 export DISPLAY=:0
 RUN=/tmp/opengeni-desktop; mkdir -p "$RUN"
+OPENGENI_DESKTOP_STAGE_STARTED_MS="${OPENGENI_DESKTOP_STAGE_STARTED_MS:-$(date +%s%3N)}"
+desktop_stage() {
+  local now_ms
+  now_ms="$(date +%s%3N)"
+  echo "OPENGENI_DESKTOP_STAGE stage=$1 elapsed_ms=$((now_ms-OPENGENI_DESKTOP_STAGE_STARTED_MS))"
+}
+desktop_stage launcher_entry
 
 # ---- SESSION ENVIRONMENT (idempotent; fixes the "Input/output error" + app breakage) ----
 # These export into the XFCE session because XFCE (stage 2 below) inherits this shell's
@@ -31,6 +38,7 @@ if [ ! -S /run/dbus/system_bus_socket ]; then
   mkdir -p /run/dbus
   dbus-daemon --system --fork >/dev/null 2>&1 || true
 fi
+desktop_stage session_environment_ready
 
 # (d) Re-assert the DEFAULT BROWSER config for THIS session's HOME. The image bakes the
 #     config system-wide (/etc/xdg) and into the /workspace skel, but a fresh /workspace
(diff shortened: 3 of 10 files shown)


jorgensandhaug (commented):
**BLOCK_CANDIDATE** — independent repair-delta review at head `182dc0f4dfb97b22df8d4af8241f202d1b08b8ed`, tree `d355b00b72c80ea314807145c7b7d4d764e8d432`, parent `86430e324465eb12da7cfa7d56c730ba8f6015d4`.

**P0: none. P1: 1 blocking defect. P2: 2 attribution gaps.**

### P1 — the provider operation is still neither canceled nor fully drained

The new fence tracks the JavaScript provider-method promise, not the sandbox process. The locked `@openai/agents-core@0.13.3` `ExecCommandArgs`/`WriteStdinArgs` contracts contain no `signal`, and locked `@openai/agents-extensions@0.13.3` Modal `execCommand`/`writeStdin` ignore the extra property. Modal may resolve `execCommand` with `Process running…` while the process continues; `Promise.allSettled(state.inFlight)` then drains and a retry may issue while the stale process is still live. A delayed `sandbox.exec` can still launch after the outer deadline.

There are two additional concrete recurrence paths:

- `await waitForFencedOperation(state)` is outside any deadline. If the timed-out adapter promise never settles, the next same-session call hangs indefinitely despite its `timeoutMs`.
- State is a `WeakMap` keyed by the transient session object. API viewer requests re-establish a new session handle for the same lease/box, so a later request bypasses the old fence entirely.

The added regression resolves the late call with a terminal result; it does not gate real dispatch, return a yielded session, prove process cleanup, exercise a never-settling drain, or use a new wrapper object for the same lease. This does not satisfy the prior required proof that no command/launch can occur after timeout.

### P2 attribution gaps

1. Host budgets now use `performance.now()`, but emitted sandbox stage elapsed values still use `date +%s%3N` in both `display-stack.ts` and `opengeni-desktop-up.sh`, so those required per-stage timings remain non-monotonic.
2. The wrapper now emits `OPENGENI_DISPLAY_TIMEOUT` only for rc 124. A genuine TERM→KILL escalation may return 137 and will be reported `unknown`; test `(8e)` injects a timeout marker that this wrapper cannot generate on rc 137. External 137 is no longer falsely labeled, but genuine watchdog attribution is not proved.

### Compatibility and checks

Current main `5a9bc1cb7dddc87612627669b74d129116d2cf96` has zero path overlap with the 10-path cumulative PR delta and zero overlap with the two repair paths; three-way merge inspection found no conflict. Exact-head CI currently shows typecheck/lint/format/tests and browser gates passed within the still-running unit job; deployment artifacts, workload images, and desktop stack passed; current-base source admission is failed. This no-rig reviewer lacked Bun/Node, so local execution was unavailable; `git diff --check`, exact-object verification, full static audit, locked SDK inspection, and compatibility checks were performed.

**Merge condition:** replace this repair with a stable sandbox/lease-epoch single-flight fence, provider/process-aware cancellation or cleanup that proves no post-deadline launch, a deadline-bounded drain, monotonic sandbox elapsed telemetry, and honest genuine-vs-external 137 evidence; add deterministic regressions for each path; then obtain a fresh independent exact-head review and green current-base admission/required CI.

jorgensandhaug (commented):
**BLOCK_CANDIDATE** — GitHub does not permit this authenticated account to request changes on its own PR, so this single `COMMENTED` review records the blocking verdict.

Pinned review: head `270c05993e5d1e6b27af75175d874cdad942e37d`, tree `d0af3605eca3d49aa70f2a4f0effdb6ba4f5fc93`, base/current main/merge-base `fcd8b3d18288658e6b31ef30502df54c16473d12` (4 commits, 10 changed paths).

### P1 — fence every active same-epoch operation before declaring the sandbox quiescent

`fenceOperation()` aborts and records cleanup only for the operation that notices the timeout (`display-stack.ts:583-608`). With two concurrent calls sharing one sandbox and epoch, if A times out while B is still awaiting provider work, A increments the shared epoch but B's controller is not aborted and B's marker is not added to `pendingCleanup`. Cleanup of A alone therefore makes `reconcilePendingCleanup()` set `physicallyQuiescent = true` (`:759-779`), allowing retry C through the gate at `:927-945` while stale B remains physically active. B does not notice the epoch mismatch until its provider await returns; until then its command can acquire the in-box flock and mutate display state after A's timeout/alongside C.

A deterministic harness against the real exported `ensureDisplayStack` reproduced the order:

```text
work-1(A), work-2(B), abort-a, cleanup-a,
work-3(C succeeds), resolve-stale-b, abort-b, cleanup-b
```

At C completion B was unresolved, un-aborted, and still active. This is not covered by `(8a)`/`(8a2)` (one active timed-out operation) or `(3b)` (concurrent success).

**Required repair:** enforce true single-flight ownership per stable physical sandbox, or register every active operation's controller and marker so an epoch advance tombstones/aborts and positively marker/PGID-cleans *all* operations in that generation before `physicallyQuiescent` can become true or any retry/degrade path can issue provider work. Add a deterministic regression with two active same-epoch calls where one times out first and prove no retry issues and no stale sibling can produce effects until both are physically stopped. The bounded provider-promise drain may remain advisory, but cannot substitute for physical ownership/cleanup.

Validation: isolated display-stack suite 26/26; channel-a suite 58/58; root typecheck 22/22 projects; `git diff --check`; launcher `bash -n`. Natural exact-head CI currently has current-base admission, deployment artifacts, desktop assertions, and workload builds successful; typecheck/unit tests remain in progress. Current main is the exact reviewed base, so no compatibility delta exists.

jorgensandhaug (commented):
SHIP_CANDIDATE

Focused independent exact-head repair review for OPE-46.

- P0: 0
- P1: 0
- P2: 0
- Candidate: `12e3e049fad7b496d9dcd8272d224ff5e5cc63fb`
- Tree: `3c1114641a9c78a69d69220cba1e88bef46320ca`
- PR base: `fcd8b3d18288658e6b31ef30502df54c16473d12`
- Live main audited: `2d21362975cc58f64dfd4ec07878fdd53654a909`

Reviewed only the focused `270c0599…12e3e049` admission repair plus current-main compatibility. The per-physical-sandbox owner queue serializes ordinary callers, invalidates queued same-epoch siblings before timeout cleanup releases the owner, preserves independent-sandbox concurrency, and keeps late provider completion from authorizing a retry. Budgeting is monotonic, and exit 137 remains unknown without the explicit timeout marker.

Checks:
- `bun test --timeout 30000 packages/runtime/test/display-stack.test.ts`: 28 pass, 0 fail.
- `bun run typecheck`: all 22 projects clean.
- `oxfmt --check` on the two repair files: clean.
- `git diff --check` for both intermediate→head and base→head: clean.
- Natural CI: typecheck/unit, desktop image + stack assertions, deployment artifacts, and workload image builds succeeded.

Compatibility: live main is descended from the PR base; base→live-main and base→candidate have zero path overlap; read-only `git merge-tree` has zero conflict markers. The sole natural-CI failure says `event base SHA differs from current main`, so it is stale-base admission policy mechanics rather than a candidate correctness failure.

Residual risk: some provider transports may ignore `AbortSignal` or never settle; safety therefore relies on the unique in-box cancellation tombstone and positive marker/PGID cleanup proof, which this repair retains and tests. Merge recommendation: merge this exact head; no substantive rebase or full rereview is required for the audited disjoint main delta.

PR #560, opened 2026-07-23
test(ogtool): keep version assertion release-safe

jorgensandhaug:
Fix the version-PR-only regression exposed by the regenerated release tree.

The CLI already reads its version from the package manifest, but its test hard-coded the pre-release `0.0.0`. Assert against that same manifest so every generated package version remains covered without weakening the behavior.

Verification:
- `bun test packages/ogtool/test/cli.test.ts`
- `git diff --check`


jorgensandhaug:
Reopening to retrigger the missing pull_request CI suite; source admission already passed for the unchanged head.

jorgensandhaug:
Superseded by #561, which implements the same manifest-derived version assertion with an equivalent focused proof. Cancelling this duplicate full CI run.

PR #565, opened 2026-07-23
chore: version packages

jorgensandhaug:
## Stable release candidate

Generated from the current `main` Changesets plan. The candidate tree contains only consumed changesets, package versions, dependency propagation, and changelogs.


jorgensandhaug:
Superseded so the canonical Changesets workflow can recreate the release PR under its bot identity for independent review admission.

PR #459, opened 2026-07-18
fix(goals): guarantee durable autonomous continuation wake

jorgensandhaug:
## Summary

- persist a goal-owned monotonic wake/observed revision and arm the continuation obligation atomically with terminal turn settlement
- transactionally materialize one logical continuation as one typed internal update, timeline event pair, usage fact, observed revision, session transition, and repairable workflow wake
- preserve authoritative human queue, Steer, approval, recovery, capacity, Pause, and cancellation ordering without duplicate goal turns
- expose repeatable-read `inactive | scheduled | running | blocked | invariant_broken` truth through contracts, SDK, React, and web UI
- make goal set/update/status/clear and their timeline facts atomic, with receipted ambiguous-success reconciliation and OPE-63's canonical event-writer lock prefix

## Failure semantics

PostgreSQL owns liveness; Temporal signals and workflow runs are repairable nudges. The stable identity is:

```
goal-continuation:<goalId>:wake:<wakeRevision>
```

A lost commit response cannot create a second update, event pair, usage row, or logical goal turn. Exhausted continuation-check retries persist a delayed outbox wake and close; `signalWithStart` reconstructs work without a human message or model polling. Worker death re-dispatches the same fenced logical turn. Goal mutation receipts bind caller/session/turn/attempt/execution-generation identity, so an ambiguous committed result is replayed rather than double-applied or allowed to overwrite newer goal truth.

## Schema and lock integration

- adds rolling migrations `0067_durable_goal_wake.sql` and `0068_goal_update_idempotency.sql`
- classifies legacy state without re-arming already materialized, human-queued, paused, or provider-blocked work
- consumes merged OPE-63's canonical lock order: `workspace_inference_controls FOR SHARE` → `workspaces FOR KEY SHARE` → session row lock for every goal event writer

## Exact-head verification

Candidate identity:

- base: `46744272e69f96329f47a0d3b1d6f93183d1d962`
- head: `674cd3036dd97dc9274773c6562dad2b543bd3fb`
- tree: `04be7adb798569f64d634680529001f3adf8df87`

Green evidence on that exact head:

- full repository typecheck: 20 projects; lint: 0 warnings/errors; format/docs/schema/generated-font checks green
- real PostgreSQL goal/OPE-63 fault and race integration: **50 pass / 0 fail / 676 assertions**, including concurrent evaluators, lost COMMIT response, six rollback boundaries, Send/Steer races, backpressure, pause/resume, corrupt-state projection, and active-turn event writes racing all receipted goal mutations
[...]

PR #630, opened 2026-07-26
feat(goals): add durable autonomous wake protocol

jorgensandhaug:
## Summary

- make active-goal continuation a durable database obligation rather than an in-memory/Temporal timing assumption
- atomically materialize goal evaluation, the internal update, audit/usage facts, and the coalescing session-workflow wake outbox
- reconcile ambiguous commits idempotently and project scheduled, running, blocked, held, and invariant-broken continuation truth through API/SDK/UI
- preserve human Send/Steer/Pause authority and prevent autonomous continuation from overtaking queued human work

This is the current-main successor to stale PR #459; that PR and its branch were not modified.

## Durable protocol

- reserve migrations `0120_durable_goal_wake.sql` then `0121_goal_update_idempotency.sql`
- claim due wake revisions with bounded retry/failure state and stale-acknowledgement fencing
- recover workflow loss from durable wake state, including cancellation and recursive Pause/Resume semantics
- make goal-update mutation idempotent across lost responses and attempt recovery
- expose truthful goal continuation state without presenting scheduled recovery as live execution

## Immutable candidate

- head: `33edf951cce7c87b0ea6b82b0e587da66c3ed38a`
- tree: `702be63f3e50bbe3df69882cb45ff66573b77589`
- base: `fe8c89071e8a03435deed3f97e58ece5df7f9dcc`
- authoritative patch SHA-256: `e9440b1891a12ec3871e09e7a7c7aea0ba2db89e93a43da60dff084244ecd17e`

## Migration reservation

- `0120_durable_goal_wake.sql`: `f3d63dd60c8d933d8729a597d947be940a0483a148665ecca47c8f05909b2577`
- `0121_goal_update_idempotency.sql`: `e90f030e9dfb3c2dc040b2192cdd874fad264020f06b9c82f1c3dcd30a9769ca`
- schema contract: version 2, 103 files, latest `0121_goal_update_idempotency.sql`, SHA-256 `d238b1ee30693a7567a1bf89be55a1611a5b6a88c78edc59d21f67055e7cb9e5`

A final fetch confirmed current `main` still ends at migration 0108, so 0120–0121 remain collision-free.

## Validation

- [x] `bun install --frozen-lockfile`
- [x] `bun run build`
- [x] full unit suite against fresh disposable PostgreSQL clusters: **4,339 pass, 4 skip, 0 fail**, 21,878 assertions, 4,343 tests / 357 files
- [x] real PostgreSQL + NATS + Temporal durable queue suite: **6 pass, 0 fail**, 90 assertions
- [x] focused FORCE-RLS suites: **51 pass, 0 fail**, 682 assertions
- [x] Temporal goal integration subset: **2 pass, 27 filtered, 0 fail**
- [x] both real local-process cancellation tests, including signal-ignoring process and lifecycle-command escalation
- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run check:public-hygiene`
- [x] `bun run release:schema-contract`
- [x] `bun run typecheck` (23 projects)
- [x] `git diff --check`

## Known limitation / release state

`test/integration/api.integration.ts` was attempted but cannot start in this sandbox because its fixture hardcodes Docker Compose and no Docker daemon is available; it failed in `beforeAll` before product assertions. No product or test bypass was introduced.

OPE-59 must remain **In Progress** after implementation/review/CI. Completion requires production deployment and live injected-failure verification of the intended wake/recovery behavior.


jorgensandhaug (commented):
Independent exact-head review of `b0dee970f1814d76b32b19779b588d277eb7015a` (tree `124ee65a09314f2257ac76151d3e346b0df2128c`): **changes requested**.

### P1 — the rollout backfill is a no-op for supported non-superuser table-owner migration roles

`packages/db/drizzle/0120_durable_goal_wake.sql` starts its cross-workspace data repair at line 54, but never establishes tenant RLS GUCs and never temporarily removes FORCE RLS. The backfill reads/writes FORCE-RLS tables including `session_goals`, `session_system_updates`, `sessions`, `workspace_inference_controls`, `codex_capacity_waiters`, and `session_workflow_wake_outbox`. `runMigrations(adminConnection, targetSchema)` does not require SUPERUSER/BYPASSRLS. Under a normal NOBYPASSRLS table owner, FORCE RLS therefore applies with no workspace context and these statements match zero rows.

The repository already documents and solves this exact migration posture in `0009_goal_sessions_first_party_goals_manage.sql`: migrations run as table owner, FORCE RLS otherwise matches zero rows, and the migration wraps its backfill in `NO FORCE ROW LEVEL SECURITY` / `FORCE ROW LEVEL SECURITY` in the same transaction. The new migration test uses the privileged admin URL, so it cannot catch the supported owner-role failure.

Impact: legacy active idle goals remain at `0/0` with no migration-created wake. If their workflow already closed—the defect OPE-59 is intended to repair—they remain orphaned until unrelated external activity.

Please make the backfill work under the supported NOBYPASSRLS owner posture while keeping the app role policy-bound, restore FORCE in the same transaction, and add a real migration regression executed by a non-superuser/NOBYPASSRLS table owner.

### P1 — one malformed legacy continuation poisons projection and materialization forever

The migration deliberately preserves a matching malformed pending update (`goalVersion: "not-an-integer"`) and arms that goal at wake `1`, observed `0` (`migration-0120-durable-goal-wake.test.ts:209-225, 381, 417-419`). Runtime then directly casts the untrusted JSON field in both `getSessionGoalWithContinuation` (`packages/db/src/index.ts:22375`) and `materializeGoalContinuation` (`packages/db/src/index.ts:23633`):

```sql
(payload ->> 'goalVersion')::integer = goal.version
```

PostgreSQL raises `invalid input syntax for type integer`. The API projection consequently 500s. The workflow reaches `maybeContinueGoal`, the materializer throws before claim-time validation can cancel the bad update, and the catch path only schedules another delayed retry that hits the same cast. `mapSessionSystemUpdate` also hard-parses the payload, so a quarantine/cancel step must precede mapping.

This contradicts the acceptance requirement to expose or quarantine malformed continuation state rather than silently orphan it. Please validate typed JSON without an exception-raising cast, atomically quarantine/cancel malformed matching rows under the canonical session/goal locks with visible invariant evidence, then materialize exactly one valid obligation. Add migration-to-runtime materialization and API-projection regressions using this existing fixture.

### Evidence and compatibility

- Exact-head CI is terminal: deployment artifacts succeeded; typecheck, lint, format, unit tests, browser gates, recovery integrations, publish/consumer checks, and client builds all succeeded before the shared web-bundle budget step failed. Both the unit job and workload image job report `direct session gzip graph: 554506 > 553984`; current main also fails that gate (`554042 > 553984`). Current-base admission failed only because the PR event base metadata is stale.
- Remote `main` moved to `99709248dba838b7df60becf5b79af113d4ef085`. Its three commits are descendants of the integrated first parent `5c1d50ae…`, touch four disjoint release-automation/docs paths, and a synthetic merge has no conflict markers or overlapping paths.
- Schema contract independently reproduces as version 2 / 104 files / latest `0121_goal_update_idempotency.sql` / `e09923184e59b6b5b22022949680d549c1c75fe0a3e9098cbb74e84d5ae90ab0`.
- This no-rig reviewer environment has no Bun, PostgreSQL, Docker, or Temporal CLI, so no local real-service execution was possible. CI does not cover either role-sensitive migration behavior or migration-to-runtime malformed-payload behavior above.

OPE-59 must remain In Progress; production deployment and live injected-failure canary are not complete.

jorgensandhaug (commented):
OPE59_DELTA: CHANGES_REQUESTED

Exact bounded delta reviewed: `b0dee970f1814d76b32b19779b588d277eb7015a` / `124ee65a09314f2257ac76151d3e346b0df2128c` → `761baee2af484757c3f835f379463c329dceac25` / `a1974082b0e3b10abe9a03bc8ec4da6999307379`.

Blocking findings in the accepted malformed-obligation repair:

1. **P1 — the new quarantine UPDATE is not executable and breaks every eligible goal materialization.** In `packages/db/src/index.ts` around lines 23585–23626, `goalRead.version` is interpolated as untyped bind parameters inside `jsonb_build_object`. Exact-head CI run [30209313585](https://github.com/Cloudgeni-ai/opengeni/actions/runs/30209313585), job [89812758544](https://github.com/Cloudgeni-ai/opengeni/actions/runs/30209313585/job/89812758544), fails with `PostgresError: could not determine data type of parameter $4`. This UPDATE is issued before pending-turn/Steer/capacity checks even when it will match zero rows, so PostgreSQL rejects ordinary `materializeGoalContinuation` calls too. CI shows 16 durable-goal failures from this query, including the new quarantine test. Cast every JSONB value bind to its intended PostgreSQL type and require the real-PostgreSQL suite to pass.

2. **P2 — a missing `goalVersion` is not quarantined.** At lines 23637–23641, the selector is `not (A and B and C)`. If the JSON key is absent, all three terms are SQL `NULL`, so the negation is also `NULL` and the WHERE clause skips the malformed pending obligation. Use an explicit two-valued invalidity predicate such as `(validity_expression) IS NOT TRUE` (or equivalent) and cover absent, JSON-null, wrong-type, non-positive/fractional, and mismatched versions.

3. **P2 proof gap — the new migration regression aborts before its post-run assertions.** `runMigrations()` returns under the `NOSUPERUSER NOBYPASSRLS` owner, but cleanup then fails with PostgreSQL `2BP01` because the role still owns `USAGE` privileges on schema `public`; therefore the FORCE-after and backfill-data assertions at lines 435+ never execute. Revoke/drop owned privileges before dropping the role and get this deterministic real-PostgreSQL test green.

Migration static review is otherwise favorable: `0120` relaxes and restores FORCE RLS for all seven required relations within the runner's single multi-statement call, and its JSON-number/positive-decimal/current-version textual matching avoids casts. `git diff --check` passes; the delta is exactly the six authorized files; `0120` SHA-256 is `5c24fb49679513e2a7cc387e738b2bbdc9b5b3f465c023ea97921882f035d983`; independently reconstructed schema contract is 104 files, latest `0121_goal_update_idempotency.sql`, hash `bd4c8f8bb509ee2bad6eb5669f83d52f928aa0e097b183b118a4541ae2c98da3`.

CI is terminal: deployment artifacts pass; typecheck/unit has 4,377 pass / 17 fail; current-base admission is stale-base metadata (`event base SHA differs from current main`); the web-image failure is the pre-existing direct-session gzip budget and outside this six-file delta. Main advanced during review to `cb58031ebab6b22ccb1fa88c81a7ee881db25470`; that descendant's release-version/changelog delta has no overlap with these six files and the synthetic merge is clean. No local Bun/PostgreSQL was available, so I do not claim local execution.

**Merge recommendation: do not merge this head.**

jorgensandhaug (commented):
OPE59_REVIEW: APPROVE

Independent exact-candidate delta review of `815c49426b7b0dda47633c44a1680847a535da8c` (tree `ef5dbb17ef108752508d199b9ac3d36cf15d1550`; parents `17ff8095…` + current main `8d2ac94e…`). **No blocking P0–P2 finding remains.**

Evidence:
- The durable protocol preserves one revisioned PostgreSQL obligation for an active idle goal, atomically materializes the typed update/event/usage/workflow wake, keeps human Send/Steer/Pause/approval/provider-capacity authoritative, avoids fake user messages, and only marks a live goal attempt as running.
- The prior SQL defects are repaired: generated `goalVersion` values are explicitly `bigint`; quarantine uses a strict positive-integer JSON-number predicate and `(validity) IS NOT TRUE`, containing missing, JSON-null, string, zero/non-positive, fractional, and mismatched versions. Exact CI passed all 18 real-PostgreSQL durable-goal-wake tests.
- Migration 0120 ran and passed under a real `NOSUPERUSER NOBYPASSRLS` owner, including the injected NO-FORCE probe and intentional rollback proof. The migration then passed unmodified and arms only truly idle legacy goals.
- Schema integration is exact: 113 SQL files, latest `0121_goal_update_idempotency.sql`, release contract `f427b98a02f6ed3feaf59be9ea1b89b709ad6fc42d9b3648f0e240ffea42ff67`; 0118 equals main, while protected 0109–0116 and 0119–0121 are byte-identical to their authorized sources. `git diff --check` passes.
- Stable scheduled-task producer-key replay reuses the committed run/session and idempotent usage/event identities without a duplicate user message; this path is unchanged from integrated main.
- Exact-head CI run 30213468370: workload images and deployment artifacts passed; changeset, typecheck, lint, and format passed; tests totaled 4476 pass / 4 unrelated opt-in skips / 1 fail. The sole failure, `claim did not block on the session row`, is in main-identical `packages/db/test/codex-capacity-waiters.test.ts` (blob `4024a6b4…`), has no candidate path attribution, and is tracked under OPE-21. The later missing visual-evidence upload is a consequence of the early test stop, not OPE-59.
- Remote PR head/tree/base were rechecked immediately before this review; GitHub reports the exact candidate mergeable and main remains `8d2ac94e…`.

Recommendation: admit and merge this exact SHA when release authority clears draft/policy gates. Keep OPE-59 In Progress until production deployment and live acceptance. Required canaries: inject one transient continuation-evaluation failure and verify exactly one autonomous recovery; exercise Pause/Resume and a racing human/Steer input with no duplicate continuation; verify malformed-version quarantine plus one update/event/usage/wake obligation and idle-only legacy migration behavior.

PR #658, opened 2026-07-27
fix: recover retained release head checks

jorgensandhaug:
Adds a fail-closed recovery mode for the narrow partial-success case where an immutable reviewed-head tag/release exists but the provider retention check was never completed.

Recovery proves the merged PR identity and tree, provider merge event, exact source-admission check, current-main ancestry, and unchanged GitHub Actions-owned immutable artifacts. It cannot create a tag or release after merge.

Focused evidence: `bun test scripts/check-release-pr-automation.test.ts` (42 pass); focused formatting and `git diff --check` pass.


jorgensandhaug (commented):
Maintainer review: PASS. The recovery mode is intentionally non-constructive after merge: it requires the existing immutable tag/release and only writes the exact provider check after provider-bound PR, merge/tree, source-admission, and current-main ancestry proofs. Focused suite: 42/42.

PR #659, opened 2026-07-27
fix: recover historical source admission

jorgensandhaug:
Extends the already-merged retained-head recovery path to reconstruct an absent `Current-base source admission` check from immutable historical PR/base/head provider data.

The recovery recomputes the complete direct tree delta and provider file projection, proves the accepted merge and current-main ancestry, and double-reads the existing immutable tag/release before the first check mutation. It still cannot create retained artifacts post-merge.

Focused evidence: `bun test scripts/check-release-pr-automation.test.ts` (42 pass); focused formatting and `git diff --check` pass.


jorgensandhaug (commented):
Maintainer review: PASS. Historical source admission is recomputed from provider-frozen base/head commits, recursive trees, compare topology, and the PR file projection. The immutable retained artifacts are double-read before any check write; no post-merge artifact creation is possible. Focused suite: 42/42.

PR #686, opened 2026-07-28
feat(codex): add native GPT-Live WebRTC path

jorgensandhaug:
## Summary

Implements OPE-11 as a native, server-mediated connected-Codex GPT-Live V3 WebRTC SDP path. It reuses OpenGeni's encrypted connected-Codex account selection, refresh, advisory-lock/CAS lifecycle, and account binding; no Codex app-server runtime is introduced.

Upstream protocol evidence: `openai/codex` commit `bd2de422aa287b97b06ca6425a10935bcf1b3731` (Codex 0.145.0).

## Compatibility contract

- Provider call: `POST https://chatgpt.com/backend-api/codex/realtime/calls?intent=quicksilver&architecture=avas`
- JSON request: `{ sdp, session: { instructions, audio: { output: { voice } }, delegation: { type: "client" }, model: "gpt-live-1-boulder-alpha" } }`
- Protocol: Frameless/V3 only (`version: "v3"`), default voice `cove`, audio SDP required.
- Subscription header contract: server-only bearer auth and conditional account/FedRAMP binding; `originator: codex_cli_rs`, `User-Agent: codex_cli_rs/0.145.0`, `version: 0.145.0`; plus `content-type: application/json`, `openai-alpha: quicksilver=v2`, and server-owned `session-id`/`thread-id`.
- Response: bounded audio SDP answer plus compatible internal `Location`; only `{ sdp, version, model }` crosses the OpenGeni API boundary.

## Security and lifecycle

- Browser receives no bearer/refresh token, account ID, credential ID, API key, provider call ID, or provider body.
- Selection remains connected session pin, then connected workspace-active account.
- One forced refresh and one replay are allowed only after provider 401; a second 401 becomes `reconnect_required`.
- No replay after ambiguous network/timeout/cancellation or 403/404/429/5xx outcomes.
- No public API-key fallback and no Codex WebSocket fallback.
- SDK helper negotiates audio + `oai-events`, fails closed on version/model drift, and closes tracks/data channel/peer transport on cancellation or failure.

## Validation

- Focused OPE-11 suite: **131 passed, 0 failed**.
- Affected package typechecks: codex, contracts, API, SDK, worker passed.
- Workspace typecheck: **23 projects clean**.
- SDK build and full production build (publishable packages, API, worker/workflow bundle, web, bundle budgets) passed.
- `format:check`, lint (zero warnings/errors), docs reference guard, public repository hygiene, changeset status, and `git diff --check` passed.
- Full repository test attempt: **4,508 passed, 4 skipped, 39 failed across 4,551 tests**. Failures were limited to unavailable Docker/PostgreSQL/Node subprocess infrastructure, two unchanged pre-existing `@opengeni/react/session` graph assertions on the base, and lifecycle fixtures contaminated by parallel full-suite credential-path state; that lifecycle file passes targeted with the three injected Git credential-path variables unset. No OPE-11 focused test failed.

## Canary

No provider call was attempted. The native route is not deployed yet, and this managed sandbox has no normal authenticated OpenGeni API base/session bearer or database access to select an eligible connected production session through the server broker. Bypassing the broker with cloud credentials or copying a production Codex token into a harness would violate the required credential boundary. The deterministic SDP-only broker/adapter tests are the prepared safe pre-deployment artifact; live execution requires the serial release authority to deploy this commit and invoke the route with an authorized existing session.

Base: `2fd2687e600ff038ef3a3270c1f8c7a776aed2c4`  
Head: `bd5f766b63fc2d5bfbd8ede932436bd5e9f6cf16`  
Tree: `150e9d258cab204826a3504fae69356af52532b9`


jorgensandhaug:
## Natural CI reconciliation

Exact head `bd5f766b63fc2d5bfbd8ede932436bd5e9f6cf16` reached terminal CI:

- Current-base source admission: **success**
- Deployment artifacts: **success**
- Workload image builds: **success**
- Typecheck and unit tests: **failure**, solely from the unchanged `packages/react/test/session-entry.test.ts` assertions at lines 52 and 92
- Automation version/report jobs: expected **skipped**

The current PR base `2fd2687e600ff038ef3a3270c1f8c7a776aed2c4` has the same terminal-red job and byte-identical two React annotations. `packages/react` is not changed by this PR. Locally, the focused OPE-11 suite remains 131/131 green; all affected package/workspace typechecks and production builds pass.

No workflow was restarted, cancelled, or bypassed, and no unrelated React fix is being folded into this focused OPE-11 PR.

jorgensandhaug (commented):
Independent automated source review of exact head `336ea1d83fc0bfc162a1ca335fd778da185d0713` found the following blocking defects.

1. **High — host-managed activation is uncallable.** `apps/api/src/routes/sessions.ts:831-870` mounts `POST .../realtime/:realtimeId/connections/:connectionId/activate`, but the fail-closed host classifier at `apps/api/src/routes/sessions.ts:2881-2920` has no activation case (and `apps/api/test/session-authorization.test.ts:8-62` omits it). A host-managed client can negotiate a ready connection, then authorization returns `null` before activation, so voice never starts. Classify this exact route as `session.realtime.control` (or a consistently authorized explicit operation) and add a host-managed route test.

2. **High — local SCTP enqueue is fabricated into a durable provider ACK.** In `packages/sdk/src/codex-realtime-v3.ts:142-148`, `sendOutbound()` calls local `RTCDataChannel.send()` and immediately adds the ledger sequence to `providerAckSequences`; `packages/db/src/session-realtime-ledger.ts:1562-1604` persists that as `provider_acked_at`. The exact pinned upstream V3 protocol has no acceptance event for `delegation.context.append` or `session.context.append`. If the local send succeeds and the peer dies before receipt, the next sync marks the row provider-ACKed and reconnect/rotation permanently excludes an undelivered result/update. Do not ACK on local send. Correlate ACK to a real upstream receipt if one exists; otherwise retain rows for replay and explicitly tolerate/document duplicates.

3. **High — a valid burst over 64 events permanently wedges the bridge.** `packages/sdk/src/codex-realtime-v3.ts:99-128` drains all `pendingInbound` into one request and, on rejection, prepends the unchanged batch for the next identical retry. The server caps batches at 64 (`packages/db/src/session-realtime-ledger.ts:18,1343-1346`) and text/payload sizes at `packages/db/src/session-realtime-ledger.ts:19-20,263-280,1458-1465`. While one sync is in flight, 65 queued events—or one parser-accepted oversized transcript/error—therefore poison every later persistence, replay, and ACK attempt. Send FIFO slices of at most 64, enforce the exact byte limits client-side, and deterministically quarantine/drop one invalid oversized event rather than requeueing it forever.

4. **High — negotiation can outlive the lease after heartbeats are stopped.** `packages/sdk/src/codex-realtime-controller.ts:590-594` stops heartbeat/rotation timers before microphone acquisition and provider setup (`:618-643`), then waits without a deadline for the data channel at `:658`; `waitForDataChannelOpen()` at `:1105-1137` also misses an already-aborted signal race. The mode lease is only 30 seconds (`packages/db/src/session-realtime.ts:20`) and the provider call alone may use 15 seconds. An open permission prompt or a permanently `connecting` data channel expires the lease before activation, and an abort just before listener registration can leave `stop()` waiting forever. Keep renewing ownership during negotiation or decouple proof from version increments, and add a bounded deadline below the lease with race-safe pre/post abort checks.

5. **High — finalized transcripts are injected twice into the first post-voice turn and remain in later history.** `packages/db/src/session-realtime-ledger.ts:1467-1483` inserts finalized transcripts into permanent `session_history_items`; the same ledger rows enter the one-time projection at `packages/db/src/session-realtime-context.ts:189-244`. `apps/worker/src/activities/run-input.ts:311-329` injects that projection while `:443-469` also loads permanent history. The first ordinary turn sees every transcript twice, and later turns retain it, contradicting the exactly-once crossover documented in `docs/run-lifecycle.md:82-97`. Choose one canonical path; consistent with the docs, keep transcripts in the realtime ledger/one-time projection and remove the permanent history insert. Add an integrated worker-input test for first-turn uniqueness and no later replay.

6. **High — mirrored cross-session updates are also delivered twice to the first ordinary turn.** `packages/db/src/session-realtime-context.ts:189-244` includes `session_update` rows in the realtime projection, while `apps/worker/src/activities/run-input.ts:311-329,424-440` separately injects the same still-pending update through normal internal-update delivery. The existing DB test at `packages/db/test/session-realtime-context.test.ts:829-857` demonstrates both representations. A child result/agent instruction can therefore be acted on twice. Exclude `session_update` rows from the post-realtime projection or atomically reconcile them with normal update delivery so exactly one model representation exists; assert this in integrated `turnInput` coverage.

7. **High — migration 0135 is not rolling-compatible with the sole-parent API binary.** `packages/db/drizzle/0135_session_realtime_connection_promotion.sql:8-36` declares rolling deployment while introducing simultaneous `active` plus `ready`/`negotiating` rows and new partial unique constraints used by `packages/db/src/session-realtime-ledger.ts:515-586`. The API at sole parent `d512f8c95e050204105fb257164854dafb687d4a` ignores `ready`, selects/closes from only `negotiating|active`, increments the mode epoch before negotiation, and completes directly to `active`. During coexistence it can close the wrong generation or hit the new `one_preparing`/`one_active` constraints, breaking reconnect/rotation. Use expand/contract or a feature gate: land schema compatible with old semantics, deploy support, drain old replicas, then enable two-phase states/indexes. Add old-binary/new-schema compatibility coverage.

8. **Medium — pending updates after the oldest represented 100 starve.** `packages/db/src/session-realtime-ledger.ts:1237-1277` limits to the oldest 100 pending/deferred updates before filtering rows already represented in the realtime ledger. After those 100 are mirrored, every later sync reselects and discards the same rows, so update 101+ never reaches the provider for the lifetime of realtime mode. Anti-join by `source_update_id` before `LIMIT`, or paginate deterministically beyond represented rows.

9. **Medium — operation-id replay accepts changed immutable input for every non-delegation kind.** `packages/db/src/session-realtime-ledger.ts:1427-1456` returns an existing row on operation-id/provider-id lookup without comparing direction, kind, role, provider identity, text, or canonical payload; strict comparison exists only for delegations at `:1026-1038`. A retry/collision with changed input can report an unrelated row—even an outbound row—as a successful inbound replay. Compare all immutable fields for every replay and return a changed-input conflict.

10. **Medium — activation buffering is unbounded and can reorder provider events.** `packages/sdk/src/codex-realtime-controller.ts:651-655` buffers every string message, including large audio deltas, until activation. At `:713-763` the live bridge listener is enabled before the buffered list is awaited, so newer live events can persist before older buffered ones; a slow activation also grows memory without a count/byte cap. Use one bounded FIFO across activation/drain, exclude irrelevant audio payloads from durable buffering, and do not enable direct ingestion until earlier buffered events are consumed.

Validation at this head: `git diff --check` passed; focused non-DB suites passed 101 tests; four DB suite setup hooks could not run because PostgreSQL was unavailable; all six changed-package TS7 typechecks emitted no diagnostics. The exact activation repro returned `null`; bridge repros showed provider ACK without a provider event and unchanged retries of a 65-entry batch. GitHub separately reports the single `Current-base source admission` check as failed and the PR as `DIRTY`/conflicting against current `main`; those CI/base issues are distinct from the source findings above.

This review is intentionally non-writing and exact-head scoped. The authenticated GitHub identity is the PR author, so GitHub may reject `CHANGES_REQUESTED`; if so, this review is submitted as `COMMENTED` while still recording that exact head is not accepted.

jorgensandhaug (commented):
REQUEST_CHANGES — submitted as a COMMENTED review because the available GitHub identity is the PR author and cannot formally request changes.

Reviewed exact immutable head `23736ebd771cb68cd88504cb8200daecf83057fb`, tree `9cd05c642532718120952b556b89a6957d4df802`, ordered parents `48348ee7a9781a8c20023ef93522c73b12bb50f1` and `dfc3235835964692b363daba63e625ca146f8472`.

Blocking findings:

1. **Startup replay can be acknowledged without being sent to GPT-Live.** `packages/db/src/session-realtime-ledger.ts:605-617` captures up to 100 unacknowledged rows and fences through the last sequence. `apps/api/src/session-realtime-context.ts:40-70` then combines them with history under the 128-item/8,192-token bootstrap budget and keeps only the newest tail, but `packages/db/src/session-realtime-ledger.ts:1634-1647` marks every row through the original fence provider-ACKed when `session.started` arrives. A probe with 100 captured 1,000-byte updates projected only sequences 70–100; the current ACK would permanently drop 1–69 from replay. Fence/ACK only exact rows proven included, or use a separate lossless/paginated startup replay path, with an overflow regression test joining projection to ACK behavior.

2. **The bridge persists transcript deltas and discards finalized transcripts/interruption truth.** The pinned upstream `openai/codex@fa1d4c40` Frameless parser maps `input_transcript.added`/`output_transcript.added` to `RealtimeTranscriptDelta` and maps `turn.done` to finalized `InputTranscriptDone`/`OutputTranscriptDone`. Here `packages/sdk/src/codex-realtime-v3.ts:275-278` enqueues the `*.added` events as durable transcript rows, while lines 291-292 only clear `speaking` and discard `turnId`, role, and finalized transcript. No producer writes ledger kind `interruption`. This violates both “persist finalized transcripts/interruption events” and “never streaming deltas.” Persist stable/deduplicated final `turn.done` truth, do not ledger streaming deltas, and add finalization/interruption tests.

3. **Backend admission does not require a connected Codex capability.** `apps/api/src/routes/sessions.ts:555-580` commits and publishes an active realtime mode using only lifecycle admission. Subscription enablement and connected-credential selection happen later in `/realtime/webrtc` through the broker. An authorized API caller—or a credential disconnect race after the UI check—can therefore receive 201, lock ordinary work, then fail negotiation with `subscription_disabled`/`credential_unavailable`; the controller leaves the active mode in terminal error until explicit stop or lease expiry (`packages/sdk/src/codex-realtime-controller.ts:943-973`). Codex-connected admission must be enforced server-side before active-mode ownership is committed, with disabled/missing/disconnected credential route tests proving no mode/lock is created.

4. **Configuration lockout is UI-only.** React disables the tool pickers, but the authorized `PUT /sessions/:id/tool-policy` path (`apps/api/src/routes/sessions.ts:1093-1100` → `packages/core/src/domain/sessions.ts:1964-2015`) has no realtime-mode fence inside its locked mutation. A direct SDK/API client can mutate durable session tool configuration while realtime/recovery owns the session. Apply a common server-side normal-mode guard to configuration mutations under the canonical lock and test the direct API race; UI disablement is not an authority boundary.

5. **This exact head is incompatible with current `main`.** Read-only fetch found current `main` at `f6b88b9a9fa83bfdda4042c276fb31324e71619b`. Three-way `git merge-tree` from merge base `dfc3235835964692b363daba63e625ca146f8472` produced conflicts in `apps/worker/src/activities/run-input.ts`, `packages/db/src/index.ts`, and `scripts/release-schema-contract.test.ts`. The exact-head “Current-base source admission” check is also completed/failure: https://github.com/Cloudgeni-ai/opengeni/actions/runs/30480605500/job/90673450982. Update from current main and resolve the machine-input/realtime ordering and migration-contract conflicts explicitly.

6. **The exact-head root build fails its enforced web budget.** `bun run build` built the packages/API/worker/web, then failed `scripts/check-web-bundle-budget.ts:134`: direct-session gzip was 560,645 bytes versus the 560,128-byte limit (517 bytes over). The corrected head must pass the checked budget rather than publishing with a red gate.

Validation evidence: frozen install; installed compiler `TypeScript 7.0.2`; 162 focused non-PostgreSQL assertions passed across Codex/provider/parser, SDK bridge/controller/WebRTC, API broker/auth/context, UI lockouts, worker continuity, and release/closure parsing; all 23 workspace typechecks passed; focused lint/format passed on 79 changed JS/TS files; 18-package closure guard passed; release-shaped publish consumer passed frozen reinstall, strict TS 7.0.2, bundles, SSR, CSS, and runtime imports. Real PostgreSQL suites could not start because this gVisor sandbox forbids Docker image-layer `unshare`; that setup limitation is not counted as a product test failure.

PR #764, opened 2026-07-29
perf(api): cache immutable workspace capture manifests

jorgensandhaug:
## Summary

- cache successfully validated inline workspace-capture manifests per API process
- bind cache hits to the full durable capture-row identity
- bound retained data to 64 entries / 16 MiB with LRU eviction
- keep large signed manifests and invalid data out of the cache

## Evidence

- `bun test apps/api/test/workspace-capture-routes.test.ts`
- `bun run --filter @opengeni/api-router typecheck`
- release acceptance reproduced capture API p95 at 203.37 ms and 206.34 ms against a 200 ms bound because every sample fetched and reparsed the same immutable object

PR #765, opened 2026-07-29
fix(db): support PostgreSQL creator management grants

jorgensandhaug:
## Summary

- classify PostgreSQL role relationships with one PostgreSQL 15-compatible catalog query shared by provisioning and runtime posture
- allow only PostgreSQL 16+'s exact non-inheritable, non-settable creator-management grant while keeping every privilege-bearing or uncertain edge fail-closed
- let managed non-superuser `CREATEROLE` administrators converge permitted app-role attributes after proving protected attributes are already disabled
- document the managed-role contract and add a patch changeset for `@opengeni/db`

## Testing

- [x] `bun run typecheck`
- [ ] `bun test`
- [x] `bun test packages/db/test/role-relationships.test.ts packages/db/test/runtime-posture.test.ts`
- [x] `bun run --cwd packages/db typecheck`
- [x] `bun run lint`
- [x] PostgreSQL 17 migration-backed regression covering automatic creator grant, repeat provisioning, runtime posture, protected-attribute rejection, and unsafe reverse membership rejection
- [x] PostgreSQL 15 catalog-query compatibility and fail-closed classification

## Notes

- PostgreSQL 15 has no per-membership `inherit_option` or `set_option`; the shared query leaves those values unknown and the classifier rejects the edge.
- The integration test uses `pgvector/pgvector:pg17` by default and can instead target a caller-supplied disposable root URL through `OPENGENI_TEST_THROWAWAY_POSTGRES_ROOT_URL`.


jorgensandhaug:
@davletd @AlmazErmilov @IuliiaPT — independent exact-head review requested for `2aca9647868637e55a73d286f2b00a7160b61b25`.

Please focus on the fail-closed PostgreSQL 15/16+ role-relationship classifier, the managed non-superuser protected-attribute preflight, and the migration-backed PG17 regression. The implementer will not self-approve or self-merge. Once an independent review is accepted and exact-head checks are green, please route the merge through the normal serialized main authority.

jorgensandhaug:
@davletd @AlmazErmilov @IuliiaPT — exact-head review update for the Gecko blocker fix:

- Current base advanced to `37bb6f78ce3d42c77df810743e15ba2f24127ac4` via #764.
- The fix branch was merged forward conflict-free; new exact head is `60514ee9a97b97df8aa6bea9e42634f2bc15f7e3`.
- Combined-tree local validation is green: DB role/posture tests 11/11, workspace-capture tests 28/28, all 23 typecheck projects, lint (0 warnings/errors), docs references, and `git diff --check`.
- New protected runs started automatically: source admission `30498747887` and CI `30498748788`.

Please review/approve only this exact head after its protected checks are green. I will not self-approve or self-merge; accepted work should go through the normal serialized main authority.

jorgensandhaug (commented):
BLOCK

Exact head: `60514ee9a97b97df8aa6bea9e42634f2bc15f7e3`
Tree: `8c532cc3e6e78c2b5733a84b9b8f66ee093f303b`
Recorded base: `37bb6f78ce3d42c77df810743e15ba2f24127ac4`

### P2 — canonical security documentation overstates the accepted ADMIN edge

`docs/architecture.md:619` says the accepted PostgreSQL 16+ ADMIN-only reverse edge “cannot convey or activate runtime privileges.” The exact automatic row is indeed `ADMIN=true`, `INHERIT=false`, `SET=false`, so it does not itself inherit the app role or permit `SET ROLE`. However, ADMIN OPTION is creator-management authority: the holder can issue a new `GRANT` to itself or another role with `INHERIT`/`SET`, after which the app role can be activated. PostgreSQL 17 documents the default false options as an accident safeguard rather than a security boundary.

Fresh PG17.10 proof on this exact code:

- automatic row: `ADMIN=true`, `INHERIT=false`, `SET=false`, non-superuser `CREATEROLE` member, superuser grantor;
- classifier result before mutation: management-only / no unsafe edge;
- the ADMIN holder successfully executed `GRANT opengeni_app ... WITH INHERIT TRUE, SET TRUE` to itself and then `SET ROLE opengeni_app`;
- classifier result after mutation: the same relationship became unsafe, as intended.

The implementation’s fail-closed behavior is correct; the canonical architecture claim is not. Reword it to state that the accepted edge does not **itself** confer inherited/SET access, while the ADMIN holder remains a trusted role-management authority capable of creating a privilege-bearing edge that subsequent posture checks reject. Because `docs/architecture.md` is the repository’s canonical security map, this is a substantive documentation-truth defect rather than optional polish.

### Verification

- Natural exact-head CI inspected once: current-base source admission, typecheck/unit, deployment artifacts, and workload image builds all succeeded; inapplicable reporting/version jobs were skipped.
- Focused classifier/runtime posture: 11 pass / 0 fail.
- PostgreSQL 15.18: shared query parsed; zero automatic reverse edges; non-superuser provisioning converged twice; a PG15 reverse edge produced unknown SET/INHERIT values and failed closed.
- PostgreSQL 17.10 migration-backed creator regression: 3 pass / 0 fail.
- PostgreSQL 17.10 dedicated-schema FORCE-RLS isolation: 7 pass / 0 fail / 35 assertions, including exact posture, same-tenant behavior, cross-tenant zero visibility, and transaction-local GUC cleanup.
- Pinned DB typecheck, stable TypeScript 7.0.2 DB typecheck, affected lint/format, docs-reference check, DB build, and diff check all passed.
- Latest captured `main` `da8cacc92b5fe60009a5a7d47bae8fa4504a991b`: base remains its ancestor; overlap is only `docs/architecture.md` and `docs/deployment.md`; disposable merge had zero unmerged entries, passed diff check, and produced synthetic tree `78b435baccf72552d630b05d61a4f92a2d47bdca`, preserving all seven non-overlap blobs exactly and both sides of the documentation overlap.

No other in-scope P0–P2 correctness, security, privacy, data-loss, migration, or protocol defect was found.

PR #773, opened 2026-07-30
perf(api): cache validated capture responses

jorgensandhaug:
## Summary
- cache the fully validated inline capture response under the existing immutable revision identity
- avoid re-running full manifest schema validation on every capture poll
- retain existing byte/entry bounds and identity invalidation

## Verification
- `bun test apps/api/test/workspace-capture-routes.test.ts` (28 passed)
- `bun run --filter @opengeni/api-router typecheck`
- `git diff --check`

## Production evidence
The exact-main live gate measured capture API p95 at 200.28 ms against the 200 ms SLO after the DB round-trip fix. This removes the remaining repeated CPU work without relaxing that SLO.

### DIFF
--- a/.changeset/calm-captures-cache.md
+++ b/.changeset/calm-captures-cache.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/api-router": patch
+---
+
+Reuse the already-validated inline workspace-capture response for an immutable capture revision instead of repeating full manifest schema validation on every poll.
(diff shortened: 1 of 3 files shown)


jorgensandhaug (commented):
```json
{
  "version": 3,
  "reviewedBaseSha": "c83ad0fd1bb8f31b3ad24225da7be2c26685cb10",
  "reviewedHeadSha": "83db425b81f3ff6ce2686d07edaee9a6481aaf67",
  "verdict": "PASS"
}
```

PR #778, opened 2026-07-30
Seed capability catalog by default

jorgensandhaug:
## Summary
- import the committed reviewed integrations catalog by default after Helm installs and upgrades
- fingerprint snapshots so unchanged deployments skip database, network, and logo-storage writes
- seed local `bun run dev` stacks by default and provide explicit Helm/local opt-outs
- keep global-catalog writes scoped to the one-shot job's migration credential

## Validation
- `bun test scripts/import-integrations-catalog.test.ts` (33 passed)
- `bun test scripts/helm-upgrade-contract.test.ts` (4 passed)
- `bun run typecheck` (23 projects clean)
- `bunx oxlint --deny-warnings ...` (targeted changed TypeScript files)
- `bunx oxfmt --check ...` (changed TypeScript files)
- `helm lint deploy/helm/opengeni`
- `helm template` with default, opt-out, and single-node values
- `bun run check:docs-refs`
- `bun test packages/db/test/catalog-imports.test.ts` (suite passed but real DB cases skipped because Docker is unavailable on the host)


jorgensandhaug (commented):
BLOCK

Exact-head substantive review of `848287fc5b4431457e8809ece3be83596cdbf89e` / tree `954ffc443fa4ef25cac6841ea56e6b5b74b7e106` (publication base `b2fb36a25aea929f56d68a5b94c51e3ca4de130b`).

## Blocking findings

1. **[P2] The new default local-dev import loses the migration-owner database URL and can abort `bun run dev`.** `scripts/dev-stack.sh:110-112` correctly sets `OPENGENI_DATABASE_URL="$OPENGENI_MIGRATIONS_DATABASE_URL"`, but then invokes `bun run catalog:import`. That package script (`package.json:34`) immediately sources `.env`, overwriting the command-scoped owner URL with `.env`'s restricted `opengeni_app` URL (and potentially reverting an auto-selected alternate port). `.env` is always created/sourced by this dev path, while the global catalog/import tables are intentionally unavailable to the runtime role. An exact Bun-script precedence probe returned the `.env` runtime URL instead of the supplied migration URL. Invoke the importer directly from `dev-stack.sh`, or otherwise preserve the already-selected owner URL across the package-script wrapper.

2. **[P2] `--if-changed` fingerprints only snapshot bytes, so documented import-mode changes are permanently skipped for the same snapshot.** `catalogSnapshotRef` (`scripts/import-integrations-catalog.ts:1156-1160`) excludes `skipLogos` and any importer/normalization contract version, while the Helm Job always uses `--if-changed`. After the default `catalogImport.skipLogos=true` import completes, changing the documented option to `false` still finds the same completed batch and exits before fetching/storing logos; later importer-semantic corrections can likewise be skipped until snapshot bytes change. Include all output-affecting options plus a deterministic importer contract/version in the completion key, or provide an explicit forced reimport path for the Helm option.

## Evidence

- Full 15-file diff and adjacent importer/DB/schema/image/Helm/network-policy/dev-stack contracts reviewed; catalog admission still fails closed for failed probes, unknown auth, unsafe API-key metadata, duplicates, and stale entries.
- Focused tests: 37 passed / 0 failed / 144 assertions.
- Stable non-preview TypeScript `7.0.2`: DB project and changed importer/Helm script surface typechecked cleanly. (The broader top-level importer test file has two pre-existing negative-fixture `"sse"` literal typing errors and is not part of the repo project typecheck.)
- Helm v3.19.0: lint plus default, opt-out, and single-node templates passed; rendered command/credential/network-policy shape is correct.
- Targeted oxlint: 0 warnings/errors; oxfmt check, Bash syntax, docs-reference guard, DB build, catalog dry-run (620/620 importable), and `git diff --check` passed.
- Natural exact-head CI was inspected once and is terminal: Current-base source admission, Typecheck and unit tests, Deployment artifacts, and Workload image builds succeeded; two automation-only jobs skipped.
- Real PostgreSQL rerun was attempted, but this isolated Modal forbids the nested-container layer `unshare`; no DB result is claimed beyond exact-head CI and source/test inspection.
- Current `main` independently fetched at `1973d2a5c14a9450e59d25226364129d2963d455`; the five overlapping paths merge with zero conflict markers, and current-main deltas contain no catalog-import/snapshot/logo/import-batch semantic overlap.

No source edits, CI dispatch, merge, release, deployment, provider/credential mutation, Azure inference, Connected Machine use, or descendants.

PR #792, opened 2026-07-30
fix(db): support managed Postgres role administrators

davletd:
## Summary

- accept PostgreSQL 16+ creator administration membership only when it is the exact `ADMIN TRUE, INHERIT FALSE, SET FALSE` edge from the app role to the current session role
- keep every other role relationship fail-closed
- let a non-superuser role administrator rotate and normalize an already-restricted app login without mentioning superuser-only attributes
- reject pre-existing privileged app-role attributes that the current administrator cannot safely remove

## Why

CloudGeni PR Cloudgeni-ai/cloudgeni#1577 embeds OpenGeni into an Azure managed PostgreSQL deployment. Its exact Gecko deployment is blocked because PostgreSQL automatically grants a non-superuser `CREATEROLE` actor this administration-only relationship when it creates `opengeni_app`. The existing guard treats that server-generated edge as unsafe, and the normalization statement also mentions attributes that only a superuser may alter.

The administration edge does not confer inherited privileges and cannot be used with `SET ROLE`; revoking it would prevent the managed database administrator from maintaining or rotating the application role.

## Verification

- real `pgvector/pgvector:pg17` integration test reproducing the managed-admin role graph
- provisioning and password rotation are idempotent
- unexpected members remain rejected
- privileged pre-existing role attributes remain rejected for non-superuser administrators
- existing dedicated-schema isolation suite
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`

## Consumer

Unblocks the next exact OpenGeni embedded release and Gecko deployment proof required by Cloudgeni-ai/cloudgeni#1577.

### DIFF
--- a/.changeset/managed-postgres-role-administration.md
+++ b/.changeset/managed-postgres-role-administration.md
@@ -1,0 +1,7 @@
+---
+"@opengeni/db": patch
+---
+
+Allow PostgreSQL 16+ managed-database role creators to normalize an application
+role when its only relationship is the server-generated, non-inheriting,
+non-settable creator administration edge.
--- a/packages/db/src/provision-roles.ts
+++ b/packages/db/src/provision-roles.ts
@@ -262,6 +262,9 @@ async function assertAppRoleSafeToNormalize(sql: postgres.Sql, role: string): Pr
     return;
   }
 
+  // PostgreSQL 16 added per-grant INHERIT and SET options. Reading through
+  // to_jsonb keeps this inspection compatible with PostgreSQL 14/15; missing
+  // options default to true so every legacy membership remains unsafe.
   const memberships = await sql<
     {
       relationship: string;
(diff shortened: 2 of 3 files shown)


davletd:
@cloudgeni-ai Please review this PR in the context of Cloudgeni-ai/cloudgeni#1577, specifically the PostgreSQL 16+ managed-role administration semantics, preservation of the bootstrap-granted ADMIN-only edge, and fail-closed handling of all other role relationships and privileged states.

davletd:
Follow-up `68a9823f` preserves PostgreSQL 14/15 inspection compatibility: the PostgreSQL 16 per-grant options are read through `to_jsonb`, and missing options default unsafe. Verified the managed-administrator provisioning/rotation test against both `pgvector/pgvector:pg15` and `pg17`; the PostgreSQL 17 run plus the dedicated-schema RLS suite passes 8/8 tests (39 expectations). Package typecheck and zero-warning targeted lint remain green.

davletd:
Closing as superseded by #765 (`2aca9647`), which implements the same managed-PostgreSQL creator-administration support on the newer role-posture architecture. The landed version is stricter: one shared relationship classifier is used by both provisioning and runtime posture, checks the grantor and actor capabilities, documents the invariant, and includes live PostgreSQL 17 plus classifier coverage. Revalidated on current `main` (`3035b59f`): `packages/db/test/role-relationships.test.ts` and `packages/db/test/postgres-role-creator-membership.test.ts` pass 6/6. Merging this older patch would conflict with and weaken the landed classifier.

PR #869, opened 2026-08-01
fix(sandbox): isolate turn-end workspace capture

jorgensandhaug:
## Problem

Production workbench acceptance on both 0.22.36 and 0.22.37 completed the real Modal turn and durable filesystem snapshot, but produced no `workspace_captures` row. The worker timed out after 60 seconds while the inner provider operation remained resident. Exact snapshot replay proves the workspace, restore, repository discovery, Git status/diff, file reads, tree, and Azure storage are healthy; the full capture finishes in about 29 seconds through a clean provider handle.

The turn-end capture reused the mutable Modal SDK session object that had just driven model tools. That handle retains exec-process bookkeeping and is the only material difference at the failure boundary. The 60-second budget also provided only a 2x margin over the production-real acceptance fixture.

## Fix

- Reopen the exact live sandbox ID as a fresh read-only/non-owning provider handle before capture.
- Fail closed if the resumed handle does not prove the same provider instance.
- Keep the turn lease heartbeat alive while blocking periodic snapshots during capture, so the 90-second holder TTL cannot expire under the new 120-second capture budget.
- Extend the bounded capture budget to 120 seconds and report the exact failing stage.
- Never close or terminate the fresh handle; the durable lease remains the sole owner.

## Evidence

- Production snapshot replay: 4 repositories, full acceptance diff/content/tree capture in ~29.5s.
- Focused worker tests: 152/152 pass.
- Workspace-capture unit tests: 26/26 pass.
- Repository typecheck: 23/23 projects clean.
- Worker typecheck, formatting, and diff hygiene clean.
- The local docker/API integration suite was attempted but its required local API was not running (`127.0.0.1:8000` connection refused); this is preserved as unavailable rather than reported green.
- UBS pre-commit scan was triaged against changed hunks; reported findings were pre-existing whole-file heuristics, with no changed-hunk defect found.

### DIFF
--- a/.changeset/clean-modal-capture-handles.md
+++ b/.changeset/clean-modal-capture-handles.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/worker-bundle": patch
+---
+
+Reopen turn-end workspace capture through an exact-instance, non-owning sandbox read handle and allow a production-realistic capture deadline.
--- a/apps/worker/src/activities/agent-turn.ts
+++ b/apps/worker/src/activities/agent-turn.ts
@@ -301,7 +301,7 @@ import {
   prepareRecordingForSettlement,
   type ActiveRecording,
 } from "./recording";
-import { captureWorkspaceRevision } from "./workspace-capture";
+import { captureWorkspaceRevision, openFreshWorkspaceCaptureSession } from "./workspace-capture";
 import type { ChannelASession } from "@opengeni/runtime/sandbox";
 import { createObjectStorage, type ObjectStorage } from "@opengeni/storage";
 import {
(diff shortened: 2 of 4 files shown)


jorgensandhaug (commented):
```json
{"version":3,"kind":"opengeni-exact-head-release-review","repository":"Cloudgeni-ai/opengeni","reviewedBaseSha":"8d54aabd4d4789260f25eab737880cb3af7a4bcf","reviewedHeadSha":"ccfe296db2e01bb751f194fdfe00b6487ca13c37","reviewerLogin":"jorgensandhaug","reviewProfile":"exact-head-maintainer-v1","verdict":"PASS"}
```

PR #920, opened 2026-08-02
fix(runtime): preserve routed exec results

jorgensandhaug:
## Summary

- preserve the structured `exec()` contract when the routing proxy targets an `execCommand()`-only backend such as Modal
- retain stdout, terminal exit status, and yielded provider session IDs instead of returning a raw SDK banner string
- restore Channel-A Files/Git framing and durable PTY process authority through the routed Modal session
- add a runtime patch changeset and a regression covering both completed commands and a live PTY

## Root cause

`RoutingSandboxSession` always exposes `exec()`. On Modal, the underlying session exposes only `execCommand()`, so the proxy fallback returned its formatted banner string directly. `SandboxChannelAService` selected the proxy's structural `exec()` method and treated that string as a `SandboxExecResult`, silently losing output, exit status, and `sessionId`. The provider process still ran and could be durably retained, but Files/Git received no trusted frame and PTY open could not adopt its process authority.

## Validation

- `bun test --isolate --parallel 1 packages/runtime/test/routing-proxy.test.ts --test-name-pattern 'exec preserves structured output and PTY authority'`
- routing proxy suite produced `51 pass / 0 fail / 206 expect()`
- `bun run --cwd packages/runtime typecheck`
- targeted Oxfmt and Oxlint — zero findings
- `git diff --check`

A pre-existing local performance test (`fsListPruned indexes a production-sized tree`) exceeded its 10-second yield guard on this Mac at ~23 seconds. The same isolated failure reproduced on untouched `0c4c96f`, so it is not caused by this change; protected CI remains authoritative for that host-sensitive test.

### DIFF
--- a/.changeset/curly-mails-rest.md
+++ b/.changeset/curly-mails-rest.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/runtime": patch
+---
+
+Preserve structured exec results when routing to execCommand-only sandbox backends so file, Git, and PTY operations retain provider output and process authority.
--- a/packages/runtime/src/sandbox/routing/routing-session.ts
+++ b/packages/runtime/src/sandbox/routing/routing-session.ts
@@ -30,7 +30,7 @@
 import type { ExposedPortEndpoint } from "../stream-port";
 import { SelfhostedControlError } from "../selfhosted/control-rpc";
 import { renderSelfhostedFault } from "../selfhosted/fault-rendering";
-import { isExecSessionLostBanner } from "../channel-a";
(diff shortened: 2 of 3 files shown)


jorgensandhaug:
Protected CI is terminal green at exact head `58460ebc31f52d877174195df350cd79d09df71c` (12 success, 3 expected skips; no failures or pending checks). @davletd, the independent exact-head review is the remaining merge gate. Root cause, focused/full validation, and the unrelated baseline Mac timing failure are documented in the PR body.

PR #928, opened 2026-08-02
fix(react): keep fleet timeline stable on unpin

jorgensandhaug:
## Summary
- remove the `React.lazy` component boundary around fleet-decision timeline rows
- keep the fleet renderer available during unrelated session metadata rerenders such as personal pin/unpin updates
- add a renderer rerender regression and a production bundle-manifest guard so the row cannot silently become a dynamic React component again

## Root cause
A pin update rerenders the open session tree. Fleet-decision rows were the timeline's only runtime-lazy JSX component and passed the JSX runtime plus shared React components across that split boundary. Under the affected production transition that boundary could resolve as an undefined element type, producing minified React error #130. The renderer is small enough to remain in the already-budgeted direct session graph.

## Verification
- `bun test packages/react/test/timeline-renderers.test.tsx --timeout 30000` (54 pass)
- `bun test apps/web/src/lib/session-context.test.tsx apps/web/src/lib/session-pins.test.ts apps/web/src/App.test.ts --timeout 30000` (121 pass)
- `bun run typecheck`
- `bun run lint`
- `bun run build` in `apps/web` (production build, precompression, bundle budget pass)
- production-bundle Playwright mock: overflow-menu unpin with a fleet decision rendered, no console or page errors

## Environment limitation
The real-PostgreSQL browser E2E could not be run in this sandbox because no Docker daemon or local PostgreSQL service is available. Existing pin browser coverage remains unchanged; the deterministic production-bundle flow and all database-free gates passed.


davletd (approved):
Reviewed exact head b3ab6426bfb5fe0881750dbfaaa0fcb42a2510ba in the CloudGeni embedded-timeline context. The static FleetDecisionRow import removes the production-only lazy-boundary failure mode without introducing an import cycle, and the production build confirms the row is no longer a dynamic manifest entry. I added the missing isDynamicEntry manifest typing before approval. Independent validation passed: 54/54 timeline renderer tests, all 23 TypeScript projects, formatting, and the production web build plus bundle budgets. No unresolved review findings. Merge remains gated on the still-running protected CI jobs.

PR #931, opened 2026-08-02
fix(react): isolate timeline renderer failures

jorgensandhaug:
## Summary
- isolate every top-level and nested timeline group behind a React error boundary
- replace an invalid or unavailable renderer with a visible, bounded fallback while preserving neighboring conversation groups
- add a regression that deliberately resolves a tool renderer to `undefined`, reproducing the React #130 failure class
- remove the fleet-specific static-import workaround and bundle guard from #928 because production-session replay did not establish that lazy row as the cause

## Diagnosis
React error #130 means React received an undefined element type. The affected session's retained event window, goal, lineage, queue, SSE failure/reconnect transition, dock state, and fleet row were replayed against the deployed bundle without reproducing a deterministic fleet-specific failure. The reliable failure surface is a timeline renderer resolving to an invalid component type. The new boundary contains that exact failure at the smallest stable timeline group and leaves the rest of the chat usable.

## Regression
The new test passes a custom `ToolRegistry` whose `resolve()` returns `undefined`. Without the boundary, React #130 tears down the timeline render. With this change, the broken group becomes “Timeline item unavailable” and the messages immediately before and after remain mounted.

## Verification
- `bun test packages/react/test/timeline-renderers.test.tsx` — 54 pass
- `bun test packages/react/test --max-concurrency=8 --timeout=30000` — 950 pass
- focused `oxlint --deny-warnings` — clean
- `bun run build` in `apps/web` — production build, precompression, and bundle budgets pass

## Follow-up context
This is a forward correction to #928, which was merged concurrently before the evidence-based replacement head could be pushed.


davletd (approved):
Reviewed exact head 1f6a8151b83f2aef68c355b8a10407561e7a49a9 in the context of CloudGeni #1577's MessageTimeline embedding. The per-group boundary contains invalid renderer types without remounting neighboring stable groups; the nested placement covers turn children, and restoring the lazy fleet row preserves the bundle split. Independently verified 54 focused renderer tests, all 23 TypeScript projects, and formatting. No blocking defects found.

davletd (approved):
Re-reviewed exact head cda2e4cba17eda0957ae4ee960d0dcff9b476aa3. The reset-key follow-up correctly lets a contained row recover when its group/registry/callback inputs change without remounting stable neighboring groups. Independently reran all 54 focused timeline tests and all 23 project typechecks; clean.

davletd (approved):
Re-reviewed exact final head 78dc87b082d428e6026c8fd149943190c2909660 after formatting repair and current-main reconciliation. The diff remains focused: per-group renderer failures are contained, reset keys allow recovery, stable neighboring groups remain mounted, and the lazy fleet bundle split is preserved. All protected checks are green; independent focused tests, full typechecks, and formatting are clean.

PR #978, opened 2026-08-03
feat: add optional X XMCP workload

jorgensandhaug:
## Summary
- add an opt-in private X XMCP workload built from a pinned upstream revision
- register it in the MCP catalog with a read-only tool allowlist
- use static OAuth1 credentials without starting an interactive browser flow
- build the image in CI and include its digest in the immutable image receipt

## Validation
- Helm lint and default/opt-in renders pass
- duplicate reserved `x` server ids fail closed
- adapter credential contract and source archive checksum pass
- documentation reference checks pass

The workload remains disabled by default and no credentials are included.


jorgensandhaug:
Closing because this implemented the wrong shape. The intended change is a normal built-in capability-catalog entry, not a private Kubernetes XMCP workload.

PR #980, opened 2026-08-03
OPE-165: make composer transcription recoverable

jorgensandhaug:
## Summary

- persist composer `MediaRecorder` output as ordered 5-second IndexedDB chunks with stable manifests and SHA-256 identities
- recover stopped/failed/aborted recordings after reload and retry the same saved audio without reopening the microphone
- fail closed when local persistence is unavailable, add explicit Retry/Discard states, and fence final-chunk, workspace-switch, unmount, and post-handoff races
- raise the default recording duration to 30 minutes and the configurable ceiling to 2 hours
- export the durable recording-store primitives, document the operational boundary, and add a patch changeset

## Current boundary

This is the authorized client-only checkpoint. Capture is durable in the browser, but final transcription still uses the existing one-shot multipart request and its current server/provider byte ceiling. This PR does not claim resumable server upload and makes no database, schema, RLS, deployment, release, installation, or cloud changes.

## Validation

- `bun run lint` — 0 warnings/errors across 1,345 files
- `bun run format:check` — 1,407 files clean
- `bun run --cwd packages/react build` — pass
- `bun run --cwd packages/react demo:build` — pass
- `bun run test:react:clean` — 1,011 pass, 0 fail
- focused transcription suite — 47 pass, 0 fail
- Chromium composer acceptance — 5 pass, 0 fail (responsive/a11y, send, reload recovery, permission/upload retention, reduced motion)
- Chromium IndexedDB acceptance — 1 pass, 0 fail (reload, deduplication, conflict detection, intentional discard)
- Changesets status — direct patch releases for config/contracts/react/sdk

A repository-wide local unit invocation also ran 5,743 tests: 5,688 passed, 6 opt-in tests skipped, and 49 unrelated environment-backed fixtures failed because this sandbox has no PostgreSQL/Docker service and injects broker credential variables into lifecycle-wrapper tests. The changed voice tests passed in that run and in the clean focused runs above.


jorgensandhaug (commented):
OPE165_RECOVERABLE_CAPTURE_REVIEW_BLOCK_V1

**BLOCK — required correctness/privacy/product-truth fixes on this exact head:**

1. **A storage/setup failure can leave the microphone live.** `getUserMedia()` succeeds before the local `mediaStream` is assigned to `streamRef`; `MediaRecorder` construction and `await store.createManifest(...)` can both fail in that window, while the catch path calls `clearCaptureRuntime()`, which only stops `streamRef.current` (`packages/react/src/hooks/use-voice-input.ts:313-338`, `:102-108`). A quota/IndexedDB failure therefore reaches the error UI while the granted track can continue capturing. Stop the local stream on every pre-registration failure path and add a regression test.

2. **Cancel/unmount/workspace replacement can still upload stale audio.** After the generation check, finalization awaits `store.listChunks(...)`; the `AbortController` is created only afterward and there is no generation/workspace check before `transcribeAudio(...)` (`packages/react/src/hooks/use-voice-input.ts:181-214`). If cancellation or unmount happens while chunk enumeration is pending, it cannot abort anything yet and the stale generation still starts the multipart upload. Fence immediately after every awaited storage read and before provider handoff (or establish an abortable controller earlier), with a deferred-`listChunks` regression.

3. **A second tab can retry or discard a recording that is still live in the first tab, and older recordings become inaccessible.** Recovery includes every non-discarded manifest, including `captureState: "capturing"` (`packages/react/src/voice-recording-store.ts:225-239`); mount exposes the newest one as recovered (`packages/react/src/hooks/use-voice-input.ts:518-526`) with Retry and destructive Discard controls. There is no owner lease/heartbeat/tab lock, so another tab can transcribe partial audio or delete the live manifest/chunks and force data loss in the recording tab. Selecting only `manifests.at(-1)` also strands any older recoverable recordings after the newest is handled. Add explicit live-owner/staleness fencing plus a way to enumerate/clean multiple recoverables.

4. **The 30-minute default / 2-hour maximum is not bounded by the 25 MiB one-shot upload ceiling.** The hook hard-stops only on `maxDurationSeconds` and never consults `capability.maxSizeBytes`; it continues persisting chunks until the duration timer and only discovers `too_large` after assembling and attempting the final multipart request (`packages/react/src/hooks/use-voice-input.ts:341-400`). This can waste a long capture that is guaranteed not to finalize. Enforce a proactive byte threshold (with final-chunk/multipart margin and truthful UI), or temporarily reduce the duration ceiling to a safely bounded value.

5. **Draft handoff is not exact-once when local state mutation and cleanup both fail.** `updateManifestBestEffort` suppresses all errors; after `setValue(next)`, the handed-off update is also best-effort and `discard()` errors are ignored before the UI is cleared (`packages/react/src/hooks/use-voice-input.ts:156-167`, `:231-255`). If both writes fail—or the page dies after append but before either durable step—the manifest remains recoverable and a reload/retry can transcribe and append the same recording again. Make handoff durable/replay-safe and add the combined update+discard failure regression.

Positive findings: chunk + manifest commits are atomic in one IndexedDB transaction; ordered hashing/conflict checks and final `dataavailable` settlement are covered; the composer has live-region labels, reduced-motion classes, and 44px coarse-pointer targets; public exports, changeset, docs, and the remaining one-shot server/provider boundary are present.

Checks: focused storage unit tests **5/5 passed**; browser-target builds for the store and hook passed; docs-reference guard and `git diff --check` passed. Local React/contracts/browser suites could not start because this read-only checkout has no installed `react`, `zod`, `playwright`, or `@axe-core/playwright`; no packages were installed. GitHub reports **14 successful checks and 3 skipped** at this exact head.

jorgensandhaug (commented):
OPE165_RECOVERABLE_CAPTURE_REVIEW_BLOCK_V2

**BLOCK — the V1 repairs are substantial, but three correctness/privacy gaps remain on this exact head:**

1. **The production “per-tab” owner ID is duplicated into opener-created tabs, defeating the ownership fence.** `createDefaultVoiceRecordingOwnerId()` stores and reuses one value in `sessionStorage` (`packages/react/src/hooks/use-voice-input.ts:69-80`). A normal `window.open()` child receives a copy of the opener's session storage, so both tabs use the same `ownerId`; `manifestAvailableToOwner` and every owner-checked update/discard then treat the second tab as the live owner. I reproduced this in the sandbox with the installed Chromium and actual `window.open()`: `{"openSucceeded":true,"openerOwner":"owner-from-opener","childOwner":"owner-from-opener","duplicated":true}`. The new tests inject distinct hard-coded IDs (`tab-a`/`tab-b`), so they miss the production path. Use a genuinely per-document owner plus reload-safe live-owner coordination (for example a lock/broadcast handshake) and add an opener/duplicate-tab browser regression.

2. **`MediaRecorder.onstop` can still repopulate a stale workspace recording after the generation was replaced.** The handler checks `generation` before `await updateManifestBestEffort(...)`, but does not check again before `rememberManifest(stopped)` (`packages/react/src/hooks/use-voice-input.ts:568-590`). If the workspace changes while that update is pending, the workspace effect increments the generation and clears visible state; the late `rememberManifest` then restores the old workspace manifest while status remains `idle`. `start()` rejects because `manifestRef.current !== null` (`:461-475`), but the recovery controls are hidden in `idle`, leaving the microphone inert until another workspace transition/reload. Add a post-await generation/workspace/recording/owner fence and a delayed-update + workspace-rerender regression.

3. **A transient cleanup failure leaves handed-off audio and transcript permanently hidden in IndexedDB.** After durable handoff, both automatic and explicit insertion paths ignore `discard()` errors and clear the visible recording (`packages/react/src/hooks/use-voice-input.ts:393-419`, `:700-725`). Recovery permanently filters out `finalizationState: "handed-off"` (`packages/react/src/voice-recording-store.ts:269-278`), and the new cleanup-failure test explicitly leaves that manifest behind. There is no later cleanup retry or tombstone GC, so sensitive audio/result data from one transient IndexedDB failure becomes an inaccessible orphan until site data is manually cleared. Add owner-safe retry/GC for handed-off records.

Verified repairs: setup failures now stop acquired tracks; cancellation during delayed chunk enumeration does not upload; finalization is generation/abort fenced around chunk reads and provider handoff; the client/server duration defaults are restored to 60 seconds with a 600-second maximum; over-limit audio is not uploaded; recovery is oldest-first; and provider text is persisted as `transcript-ready` before draft mutation with explicit uncertain insertion instead of automatic retranscription.

Checks: storage unit tests **5/5 passed**; browser-target builds for the store and hook passed; docs-reference guard and `git diff --check` passed. A dependency-free Chromium CDP check reproduced the duplicate production owner ID above. This checkout has no installed React/Zod/Playwright dependencies, so those local suites could not start and no packages were installed. Exact-head GitHub checks currently show **12 success, 3 skipped, 2 failure**; the browser failure is in unrelated `session-pins.browser.e2e.ts`, and the other failure is the aggregate check reporting that browser result, so CI is not the basis for this BLOCK.

jorgensandhaug (commented):
OPE165_RECOVERABLE_CAPTURE_REVIEW_BLOCK_V4

**BLOCK — a stale recorder callback can destroy the successor workspace capture.**

At exact head `fade1978d940b7aaf4ba5ee97c20d25696b89086` (`packages/react/src/hooks/use-voice-input.ts`):

- The workspace effect stops the old recorder and immediately clears the visible/runtime state (`852-866`), so the user can start a new recording while the old recorder's `onstop` event is still deferred.
- The old `onstop` callback then calls the shared `clearCaptureRuntime()` and consumes the shared persistence/settled refs (`601-610`) **before** its generation/workspace/owner fence (`611-618`).
- `clearCaptureRuntime()` operates on the hook's current global timer, stream, and recorder refs (`154-160`), not on the recorder/stream captured by that callback.

This lets an old-workspace callback stop the new microphone track, null the new stream/recorder refs, and resolve the new attempt's settle promise. The new UI remains in `recording`, but `stop()` is then a no-op because `recorderRef.current` was cleared.

I confirmed this with a focused exact-head hook reproducer using deferred stop delivery: start in `ws-old`, switch to `ws-new`, start the successor recording, then deliver the old `onstop`. The reproducer observed the successor track stopped once, `stream === null`, `status === "recording"`, the successor recorder still marked recording, and a subsequent `stop()` unable to change status (`1 pass`, `0 fail`, `12 assertions`; temporary harness removed).

**Required repair**

1. Fence callback ownership before any shared mutation. Capture attempt-local recorder, stream, persistence queue/errors, and settle resolver; only clear shared refs when they still identify that same generation/workspace/owner/recording attempt.
2. Add a committed regression using the existing deferred-stop fake: switch workspaces, start a successor capture, fire the old stop event, and prove the successor stream/recorder/timer/status remain active and stoppable.

The V4 prerequisite-generation/workspace checks do repair the previously reported permission/setup race, but they do not protect late recorder callbacks after a successor starts.

**Verification/provenance**

- Candidate tree: `dd3da689cef6c2e198e2f85dadc82998375dd951`; immutable PR base: `31fb86b4c67acb3eb99389d90fa702079aa0b68a`.
- Exact-head CI: 17/17 completed — 14 success, 3 intentional skips, no other conclusions.
- Focused store test: 5 pass; docs-reference check and `git diff --check`: pass.
- Live `main`: `4e89f269b0efb09b3f5913bee6455bfbc5a5c63d`. Read-only synthetic integration was conflict-free, `git diff --check` clean, tree `25e3833f05036cde3a8c226c5e7f4ec1a31466b8`; the post-handoff main delta has zero path overlap with the two V4 repair files.
- GitHub's published merge ref remains the earlier conflict-free synthetic `465df3a36dd8a6725923f38b6dc90ce2c16a1853` (tree `d61cf643ca045951f9e403c004e01fb72513bd40`, parents `b121e7c...` + candidate).
- Existing review threads: 0.

jorgensandhaug (commented):
OPE165_RECOVERABLE_CAPTURE_REVIEW_PASS_V5

**PASS — the V4 stale-callback blocker is repaired at exact head `682fb5d57eaf1dc26a5b24543d51440cd708c7f8`.**

I audited the complete `fade1978d940b7aaf4ba5ee97c20d25696b89086..682fb5d57eaf1dc26a5b24543d51440cd708c7f8` repair and the surrounding capture lifecycle.

The repair now makes recorder-attempt state local to the `start()` invocation: stream, recorder, manifest identity, persistence queue/error, capture-limit error, settlement resolver, chunk sequence, and timing. `attemptOwnsSharedCapture()` requires the exact generation, workspace, owner, stream, recorder, and manifest before a callback may replace or clear shared runtime state.

In particular:

- A delayed stale `onstop` no longer invokes shared cleanup or consumes the successor's queue/resolver. It stops only its captured old stream, drains only its attempt-local persistence queue, resolves only its attempt-local settlement, and then fails the generation/workspace/owner fence before any visible-state or finalization mutation.
- A stale permission/setup rejection no longer clears successor runtime or reports a stale error. Cleanup is limited to the stream acquired by that attempt, shared resolver cleanup is identity-guarded, and the generation fence precedes visible status/error mutation.
- Active capture behavior remains intact: the current attempt still exposes its queue/settlement to discard, workspace-change, cancel, and unmount paths; final chunks settle before finalization; storage and size failures remain recoverable.

The committed regressions cover both required adversarial sequences:

1. Deferred old-workspace stop after switching workspaces and starting a successor: the successor track is not stopped, stream identity and `recording` status remain unchanged, the successor recorder remains active, and it subsequently stops normally with ordered chunks `[0, 1]`.
2. Delayed old permission rejection after the successor starts: the old start returns false without stopping or clearing the successor, and the successor remains active and stoppable.

**Verification/provenance**

- Candidate tree: `115188fbc625cc063911af2ba155e1a4ced36a29`; parent: `fade1978d940b7aaf4ba5ee97c20d25696b89086`; immutable PR base and merge base: `31fb86b4c67acb3eb99389d90fa702079aa0b68a`.
- Exact-head CI: 17/17 completed — 14 success, 3 intentional skips, 0 pending/failure/cancelled.
- Live `main`: `d46922247e7d8e734d3a628ba03de58ec6ab3b7f`. The post-V4 main delta has zero path overlap with the two V5 repair files.
- GitHub current synthetic merge: `1d10662cc734ee9398200950824939db574af09e`, tree `aa2c9f10ec42f304557dd2570b8e4159f4693d00`, parents live main + exact candidate; PR reports mergeable.
- Existing review threads: 0.

No blocking correctness, privacy, ownership, workspace/generation, cleanup, data-loss, compatibility, accessibility, documentation, public-contract, or provider-boundary issue remains in this V5 repair.

PR #1012, opened 2026-08-04
fix: include API contract header on managed mutations

jorgensandhaug:
## Summary
- include the API contract header on managed-session state mutations
- preserve cookie-only authentication for managed sessions
- add regression coverage for redemption requests

## Verification
- bun test apps/web/src/api.test.ts
- bun run --cwd apps/web typecheck

### DIFF
--- a/apps/web/src/api.test.ts
+++ b/apps/web/src/api.test.ts
@@ -3,6 +3,7 @@ import {
   authHeadersForAccessKey,
   configureClientAuth,
   createOpenGeniClient,
+  redeemCodexResetCredit,
   resolveApiBaseUrl,
   sendVerificationEmail,
   setStoredAccessKey,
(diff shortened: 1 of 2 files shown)


jorgensandhaug (commented):
Current-main integration is not clean for the requested base. A three-way merge of exact head 6ea0bb627431070c80bf303481bab742e9eb5d5c onto current main a532e2ba55d1e3e92fed936ad6eb72e33205ce46 (merge base 664c1d86b7cd9bcafa9a00a4de75b8e997c6cc6e) produces a modify/delete conflict at .changeset/bright-otters-timeout.md: base and head retain blob e51e9236ff2fc45554a814b5fbe13ecd3923dba5, while current main deletes the path. Resolve/rebase this conflict before merge; I am not marking the head safe against the requested current main.

PR #1048, opened 2026-08-04
fix(git): preserve provider-defined repository paths

davletd:
## Summary

- preserve the exact provider-defined HTTPS clone path instead of manufacturing or removing a trailing .git suffix
- centralize exhaustive provider path semantics in contracts: GitHub and GitLab opt into .git aliases; Azure DevOps and provider-neutral remotes remain exact
- use the shared identity and alias helpers across core normalization, runtime credential routing, worker credential grouping, and SDK manifest materialization
- make stock GitHub clients submit an explicit GitHub provider while manual/custom repository URLs retain exact-path behavior
- document and changeset the cross-package contract

## Root cause

OpenGeni treated a trailing .git suffix as universally optional. That assumption corrupted the CloudGeni Gecko Azure DevOps clone URL because Azure treats the suffix as part of the repository name. The same assumption also existed in runtime and worker credential identity, where repositories named repo and repo.git could collide.

## Design

The clone URI is now transport data: OpenGeni validates and sanitizes HTTPS authority but preserves the path. Provider-specific equivalence is an explicit exhaustive capability policy. Adding a future provider requires selecting exact or dot-git-alias semantics, and unqualified providers default to exact. Mount paths, display names, resource identity, and credential aliases are derived separately.

## Verification

- bun run typecheck: 23 projects clean
- focused repository/core/runtime/worker/API/web suite: 501 pass, 0 fail
- bun run format:check
- bun run lint
- bun run check:docs-refs


davletd:
Independent review update for the current exact head:

- PR head: c6c603bff6f3a28e461511eee78d5cd038637e21
- Repository-path fix: 38d3927b431549ca03a1e61b29f85836b5aa1af2
- Latest head is a clean merge of current main with no manual conflict resolution.
- Source Admission: green
- Full CI run 30946723845: green across source/type, all four unit shards, real-service/recovery, browser, package/bundle, deployment artifacts, and all workload images.
- Local focused verification: 502 pass / 0 fail across the affected OpenGeni suites; all 23 typecheck projects clean; format/lint clean.

The implementation preserves exact provider-defined HTTPS clone paths, makes dot-git aliasing an exhaustive provider capability for GitHub and GitLab only, and keeps Azure DevOps and provider-neutral paths exact. CloudGeni's corresponding credential authorization guard is now fixed and under exact-head CI in PR 1577.

@jorgensandhaug the repository's required independent approval is the only remaining blocker before this can enter the immutable release chain.

PR #1051, opened 2026-08-04
fix(worker): recover required MCP connectivity

jorgensandhaug:
## Summary
- preserve a secret-safe, allowlisted connectivity marker when required MCP setup is wrapped by the SDK
- recover the same accepted turn with bounded connectivity backoff instead of terminally failing the session
- keep HTTP client failures and unknown transport codes terminal, without retaining raw URLs, messages, or response bodies
- document the lifecycle invariant and cover runtime, worker, and activity integration behavior

## Validation
- `bun test packages/runtime/test/runtime.test.ts` (211 passed)
- `bun test apps/worker/test/agent-turn.test.ts apps/worker/test/session-state-mcp-timeout-recovery.test.ts apps/worker/test/session-activity-cancellation.test.ts` (145 passed)
- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun run check:docs-refs`
- `bun run build`

The real-service activity integration case is included for protected CI.


jorgensandhaug:
Current main gained migration `0174_session_wake_live_interruption.sql` after this PR's first green run. The rebased CI correctly exposed that main's immutable release-schema test had not been advanced with the migration. Commit `85287424` adds only the missing count/hash/latest-migration/digest/mode assertions; `bun test scripts/release-schema-contract.test.ts` passes (5/5). A fresh exact-head CI run is now required.

PR #1083, opened 2026-08-05
fix(config): enable Codex tool search by default

jorgensandhaug:
## Summary
- default progressive Codex MCP tool disclosure to enabled
- retain an explicit  emergency opt-out
- lock the contract with a configuration test and changeset

## Verification
- bun test v1.3.13 (bf2e2cec) (92 passed)


jorgensandhaug:
Superseded by #1085, which includes this change together with the provider-accounted compaction hotfix and is now merged to main.

PR #1089, opened 2026-08-05
fix(runtime): honor current Modal snapshot timeout

jorgensandhaug:
## Root cause

`OPENGENI_SANDBOX_SNAPSHOT_TIMEOUT_MS` bounded OpenGeni capture orchestration but was never passed into `ModalSandboxClient`. Modal therefore retained its 60s default. Worse, a warm resume replayed the stale timeout frozen in an older provider envelope, so raising production configuration could not repair an existing large box.

## Fix

- pass the configured timeout into every newly constructed Modal client;
- on warm resume, replace only the operational snapshot-timeout field with current policy while preserving every provider identity and filesystem field;
- retain the existing Modal 0.9 indefinite-image-retention adapter;
- document the current-policy/legacy-envelope contract.

## Production evidence

Lease `21fd77b3-8e9c-4bd3-b2ea-b45833a680b4` owns a healthy 1.4 GiB `/workspace` with 152,314 files. Reaper captures repeatedly timed out at ~60s. After production config was raised to 600000ms, the next legacy-envelope resume still timed out at ~60s, deterministically proving the stale-envelope/config-wiring defect.

## Validation

- `bun test packages/runtime/test/modal-snapshot-retention.test.ts` — 12/12
- `bun run typecheck` — 23/23 projects
- `bun run lint`
- `bun run format:check`
- `git diff --check`
- UBS pre-commit scan triaged: reported only existing syntax/test-fixture false positives; no actionable finding


jorgensandhaug (commented):
PASS — exact head e39264c380d892ab53efccdbbad7bd1d60685b6c (tree 349453d6fd31a00ce9bb5522ef1378aa9acc24fb). Verified production root cause, current-config wiring for new Modal sessions, stale-timeout replacement on legacy warm resume without mutating the input provider state, Modal 0.9 retention policy preservation, focused 12/12 tests, 23-project typecheck, lint, format, diff hygiene, and UBS triage. No replay or provider-identity semantics changed.

jorgensandhaug (commented):
PASS — exact-head review of db141c426786ec78690f640555467443a8f711a8. Verified current Modal snapshot policy is applied both when building new clients and when resuming legacy envelopes without changing provider identity; durable Agents SDK backend IDs are normalized through one shared descriptor-registry mapping before reaper/routing provider lookup; unknown/prototype keys fail closed. Focused sandbox/provider/reaper suite 64/64, 23-project typecheck, lint, formatting, and diff hygiene are clean. UBS diff scan was inspected; reported findings are pre-existing/textual false positives outside this patch.

PR #1124, opened 2026-08-05
fix(test): follow current secret contracts

jorgensandhaug:
## Summary

- validate social OAuth PKCE state by decrypting through the supported variable-set crypto contract instead of pinning `v1`
- validate migration 0172's applied default against its immutable SQL literal instead of the evolving first-party tool registry
- keep both fixes test-only and future-compatible

## Verification

- `bun test apps/api/test/social-oauth.test.ts packages/db/test/migration-0172-retire-model-visible-github-token.test.ts` — 15 pass
- `bun run typecheck`
- `bun run lint`
- `bun run check:public-repo-hygiene`

Linear: OPE-174

### DIFF
--- a/apps/api/test/social-oauth.test.ts
+++ b/apps/api/test/social-oauth.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, test } from "bun:test";
 import { randomBytes } from "node:crypto";
-import type { Settings } from "@opengeni/config";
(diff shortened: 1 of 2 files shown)


jorgensandhaug:
Superseded by #1111, which landed fixes for both failing tests on main before this PR could merge. Closing to avoid duplicate test-only changes and CI churn.

PR #1151, opened 2026-08-06
fix(react): retain SessionChrome token defaults

jorgensandhaug:
## Summary
- make the Tailwind source bridge import the same derived fallback-token layer used by compiled CSS
- compile from that canonical bridge instead of maintaining a second token import path
- cover SessionChrome in the release-shaped external-consumer proof

## Why
SessionChrome consumes derived fallback tokens. The compiled stylesheet included those tokens, while the Tailwind source bridge did not, so source-integrated consumers could lose the component defaults even though the compiled path remained correct.

## Validation
- React typecheck
- focused compiled-CSS and SessionChrome tests
- generated CSS freshness check
- repository formatting check
- release-shaped packed-consumer proof, including strict types, SSR, session-only and realtime-only bundles, compiler-free scoped CSS, and packed artifacts
- visual verification at narrow and wide widths for both source-bridge and compiled-CSS paths

The web production bundle remains at the exact current-main size and hits the existing bundle-budget baseline failure; this patch does not change JavaScript output or weaken that gate.

### DIFF
--- a/.changeset/calm-session-chrome-tokens.md
+++ b/.changeset/calm-session-chrome-tokens.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/react": patch
+---
+
+Keep derived design tokens available through both the compiled stylesheet and the Tailwind source bridge so SessionChrome retains its complete default presentation in every supported integration path.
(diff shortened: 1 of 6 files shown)


jorgensandhaug (commented):
```json
{
  "version": 3,
  "reviewedBaseSha": "a382cbebd741e3c1615bd17ffaa09b31d78ef6c5",
  "reviewedHeadSha": "26f94880f637c55e787b279e500a4d690c4569fa",
  "verdict": "PASS"
}
```

PR #1158, opened 2026-08-07
fix(agent): rotate split NATS generations

jorgensandhaug:
## Summary

- Treat either control or bulk NATS transport loss as loss of the entire Connected Machine connection generation.
- Latch async-nats disconnect events so an internal reconnect cannot leave one lane stale while the other continues serving heartbeats.
- Escalate the first bulk publish failure into the same generation-rotation path instead of repeatedly publishing into a closed channel.
- Preserve the existing op-stream contract: streamed host work detaches and re-attaches across the reconnect rather than being replayed or killed.

## Problem

The control and bulk clients form one logical generation, but only control subscription/heartbeat failure previously ended that generation. If the bulk client became unusable while the control client internally reconnected, the machine could continue reporting healthy liveness while op-frame publication failed indefinitely. That also produced a closed-channel warning storm and prevented retained output from healing through a fresh attachment.

## Validation

- `cargo fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test -p opengeni-agent` (165 passed)
- `cargo test --workspace --all-features`
- `cargo build --release --workspace`

Regression coverage includes a real local NATS server exit, level-triggered disconnect signaling, and bulk-lane loss ending an otherwise healthy control generation.

### DIFF
--- a/agent/crates/opengeni-agent/src/engine.rs
+++ b/agent/crates/opengeni-agent/src/engine.rs
@@ -39,7 +39,7 @@ use opengeni_agent_engine::registry::{
 use opengeni_agent_engine::retention::RetentionConfig;
 use opengeni_agent_engine::{Frame, HostCapacity, OpId};
 use opengeni_agent_platform::ContainedExec;
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{mpsc, oneshot, Notify};
 use tracing::{info, warn};
 
 use crate::job::{run_job, JobCommand, JobConfig, JobEnd, JobExit, JobHooks, JobParams};
@@ -714,7 +714,7 @@ impl Engine {
 
     /// Detaches only jobs owned by one deployment/workspace link. Other links
     /// continue streaming uninterrupted when this transport reconnects.
-    pub fn detach_scope(&self, scope: &str) {
+    pub async fn detach_scope(&self, scope: &str) {
         let prefix = format!("{scope}:");
         let routes: Vec<mpsc::Sender<JobCommand>> = self
             .routes
@@ -724,8 +724,27 @@ impl Engine {
             .filter(|(op_id, _)| op_id.as_str().starts_with(&prefix))
             .map(|(_, sender)| sender.clone())
             .collect();
+        let mut completions = Vec::with_capacity(routes.len());
         for sender in routes {
-            let _ = sender.try_send(JobCommand::Detach);
+            let completed = Arc::new(Notify::new());
+            if sender
+                .send(JobCommand::Detach {
+                    completed: Some(completed.clone()),
+                })
+                .await
+                .is_ok()
+            {
+                completions.push((sender, completed));
+            }
+        }
+        // FIFO mailbox ordering makes each receipt a hard generation fence:
+        // every earlier Ack/Attach has been handled and Detach has disabled
+        // emission before a successor bulk lane may become visible.
+        for (sender, completed) in completions {
+            tokio::select! {
+                () = completed.notified() => {}
+                () = sender.closed() => {}
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


jorgensandhaug:
Hosted CI note: `Package and bundle contracts` fails only at the web bundle budget (`initial raw graph: 1298735 bytes exceeds 1294336`). Exact current `main` SHA `8df543e475922562eb567c306dc663521c96a79f` fails with the identical byte counts in CI run `31145867974` / job `92765046962`. This PR changes only `agent/crates/opengeni-agent/src/supervisor.rs`; all Agent CI platforms and local Rust workspace gates are green. The base failure is being kept out of this focused transport fix.

jorgensandhaug:
Additional exact-base CI comparison before merge:

- `Real-service and recovery tests` fails on both PR head and exact base `8df543e475922562eb567c306dc663521c96a79f` at the same sandbox E2E timeout, with the same terminal event sequence ending in `turn.failed`.
- `Browser and visual acceptance` fails on both because the production web build trips the same pre-existing bundle budget; the missing screenshot artifacts are downstream.
- `Build API, worker, and web workload images` fails on both in the web-image build at `initial raw graph: 1298777 bytes exceeds 1294336`; the workload-image aggregate is downstream.
- `Typecheck and unit tests` is an aggregate gate reporting those same failed lanes.

Exact-base run: `31145867974`. PR run: `31147556010`. The Agent CI matrix, source/type contracts, all four unit shards, deployment artifacts, relay image, and headless-sandbox image are green. The patch remains confined to the Rust supervisor.

jorgensandhaug:
Independent-review remediation is now pushed at exact head `3b68e22c91a0fda448314f31c14df2bd715518e6`:

- Connection-generation teardown sends a receipt-bearing FIFO `Detach` to every scoped live pump and waits until each pump has applied it; receiver closure is accepted as terminal proof so a crashed pump cannot strand reconnect.
- The shared bulk sender is not replaced until that hard detach fence completes, preventing an old pump from emitting through the successor generation before explicit `OpAttach`.
- The old bulk publisher is now aborted **and awaited** before the generation returns.
- Added regression coverage for a full/delayed pump mailbox, closed-pump proof, an active streaming op remaining silent on a successor lane until higher-generation reattach, and real isolated control-only and bulk-only async-nats disconnects.

Fresh local validation on the exact head:

- `cargo fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test -p opengeni-agent` — 169 passed
- `cargo test --workspace --all-features` — passed
- `cargo build --release --workspace` — passed

A fresh independent review of this new SHA is in progress; the prior blocked verdict is not being reused.

jorgensandhaug:
Final exact-head verification for `3b68e22c91a0fda448314f31c14df2bd715518e6` on base `8df543e475922562eb567c306dc663521c96a79f`:

- Independent read-only re-review: **PASS, no blocking findings**. The reviewer verified the FIFO detach completion fence, receiver-closure terminal path, stale-frame exclusion until explicit higher-generation reattach, and publisher abort + awaited teardown. An archived copy of the exact commit passed all 169 `opengeni-agent` tests.
- Agent CI run `31149249523`: all Linux, macOS, and Windows test/build and install-smoke jobs passed; fmt + clippy passed.
- General CI run `31149249458`: source/type contracts, deployment artifacts, all four unit shards, relay image, and headless sandbox image passed.
- Remaining red jobs are proven exact-base failures from run `31145867974`, not introduced here:
  - package budget: both base and head report `initial raw graph: 1298735 bytes exceeds 1294336`;
  - web image: both base and head report `initial raw graph: 1298777 bytes exceeds 1294336`;
  - real-service recovery: both base and head time out waiting for `workbench-capture-e2e-complete` with the same event sequence ending in `turn.failed` then `workspace.revision.captured`;
  - browser and aggregate jobs are downstream of those unchanged failures.
- Local exact-head validation passed: `cargo fmt --all --check`, clippy with warnings denied, `cargo test -p opengeni-agent` (169 tests), full workspace tests, and release workspace build.

No deployment or activation is part of this merge.

PR #1171, opened 2026-08-07
fix(worker): guard single-node memory headroom

jorgensandhaug:
## Summary

- continuously enforce the resource-based turn worker's memory target after activity admission using host `MemAvailable` and the most pressured finite process cgroup or ancestor
- request the existing graceful worker drain/checkpoint lifecycle after a sustained breach; do not hard-kill turns or replay outcome-unknown provider operations
- keep a synchronous shutdown-delivery failure retryable and expose it through the guard failure metric instead of permanently latching the worker in `draining`
- add bounded guard metrics plus running-workload-scoped node PSI, swap, kubelet runtime, and readiness recording/alert rules
- map kubelet `job`/`instance` series to Kubernetes nodes through `kubelet_node_name` before node aggregation
- update the canonical worker dashboard, observability verifier, single-node configuration example, lifecycle/architecture/deployment docs, and package changesets

## Why

Temporal's resource-based slot tuner prevents new activity admission when the configured memory target is already exhausted. It cannot by itself react when a long-running activity grows substantially after admission. On a shared single-node deployment, that growth can consume host headroom and cascade into swap, I/O pressure, container-runtime failures, and node-wide service unavailability before a container-specific limit or OOM signal exists.

The guard uses the same memory target as admission, requires a sustained breach, and enters the ordinary graceful worker shutdown path so durable conversation/checkpoint and outcome-unknown settlement semantics remain authoritative. It resolves the process's cgroup membership from `/proc/self/cgroup` and evaluates finite current and ancestor limits, covering ordinary containers, systemd services, and embedded workers with a host-root cgroup mount.

## Validation

- 128 focused changed/adjacent tests pass across worker memory/lifecycle/observability, configuration, and Helm upgrade contracts
- 34 final regression tests pass across worker memory/lifecycle/observability and the observability deployment plan
- `bun run typecheck` (23 projects)
- `bun run lint`
- `bun run format:check`
- `bun run check:docs-refs`
- `cd apps/worker && bun run build`
- application Helm lint and render with `monitoring.coreos.com/v1` enabled
- pinned self-hosted and production observability-wrapper Helm lint/render plus dashboard/RBAC verifier
- rendered rule file loaded successfully by Prometheus 2.55.0 through server startup

The broad unit command on the first revision also ran 6,173 tests in this sandbox: 6,123 passed, 6 skipped, and 44 infrastructure-dependent tests could not acquire Docker/PostgreSQL because neither is available in this environment. The focused changed/adjacent suites above pass on the current revision.


jorgensandhaug:
Independent read-only reliability review: **PASS — no blocking findings** at exact final head `3a997e9800924f7da51f7f8079656cf1928b735d`.

The review specifically re-verified:

- `/proc/self/cgroup`-based v1/v2 current and ancestor finite-limit discovery;
- retryable synchronous worker shutdown failure with guard failure telemetry;
- kubelet `(job, instance)` to Kubernetes `node` mapping and running-pod scoping;
- tests, documentation, dashboard/verifier contracts, and public API compatibility;
- exact patch equivalence after the one-time rebase onto current `main`.

Residual non-blocking portability risks remain documented around unusual cgroup-v1 mount layouts, environment-specific node-name label equality, empty live label joins, and the intentionally unchanged public `drain(): void` compatibility surface.

Final-head hosted CI and Source Admission are green.

PR #1183, opened 2026-08-07
Fix sandbox reaper snapshot timeout contract

jorgensandhaug:
## Summary
- make the sandbox reaper activity outlive the durable provider-capture fence and cleanup margin
- fail worker startup when configured snapshot timing violates that contract
- add positive, near-boundary negative, and workflow-wiring tests

## Validation
- `bun test apps/worker/test/sandbox-reaper-timeout-contract.test.ts` (3 pass)
- `bun test apps/worker/test/embedded-lifecycle.test.ts` (13 pass)
- `bun run typecheck` (23 projects clean)
- `bun run lint` (0 warnings/errors)
- `git diff --check`

No deployment, merge, workflow dispatch, or CI rerun was performed.


davletd (approved):
Independent deep review complete for exact head 323c3336. Reconciled current main; added the missing worker changeset; fixed the uncovered multi-lease timeout gap by bounding each activity to one capture-bearing drain; updated architecture; 50 sandbox/worker regressions, all 23 typechecks, lint/format/diff checks, and the Temporal worker bundle pass locally. Merge remains gated on exact-head hosted CI.

davletd (approved):
Fresh exact-head approval after reconciling versioned main 9f814151. The PR-only diff remains the timeout/batch contract, worker changeset, tests, and architecture update. Revalidated 50 sandbox/worker tests, all 23 typechecks, lint/format/diff, changeset plan, and worker bundle on exact head d067a831.

jorgensandhaug:
Independent release-boundary review found two blocking issues:\n\n1. [P1] Timeout budget ignores work before the drain capture. apps/worker/src/sandbox-reaper-timeout.ts proves only archiveCaptureTimeout times maxDrainable plus cleanupMargin, but reapSandboxLeases does gauge refreshes, rotation requests, warm metering, force-drain checks, and retained-process reconciliation (up to 20 provider probes) before the bounded drain batch. A permitted near-boundary snapshot timeout therefore leaves effectively no pre-capture budget, so the fixed activity deadline can still expire on the very path this patch intends to protect. Reserve a bounded pre-capture budget or put that prelude behind its own proven deadline.\n\n2. [P2] Fail-closed validation is skipped by non-schedule-owning control workers. The assertion currently runs only from registerSandboxReaperSchedule, gated by workerOwnsInternalSchedules. Control workers configured not to own internal schedules can still poll and execute reapSandboxLeases on the global control queue without validating their environment. Enforce the assertion in every control-worker startup or activity-construction path that can execute the activity.\n\nPlease add regressions for both paths and rerun exact-head CI and independent review before merge.

jorgensandhaug:
[P1 additional CI evidence] Unit shard 2 fails at apps/worker/test/sandbox-lease.test.ts:581 because the new sandboxReaperDrainableBatch limit admits only one of the two simultaneously drainable leases. The existing test expects both the post-create warming-death row and the already-draining row to be terminated in one reaper pass; under this patch one remains undrained. This is a direct behavior change, not an unrelated flake. Please resolve the backlog and throughput contract explicitly rather than only weakening the assertion. A one-row-per-schedule limit can accumulate drainable boxes under ordinary concurrent usage, especially with the added pre-capture work identified above.

davletd (approved):
Fresh exact-head approval after fixing the relevant hosted integration expectation: immediate force-drain/admission fencing is asserted, teardown is proven across bounded subsequent sweeps, and turn-held boxes remain protected. Lint/format and focused timeout/warm-meter tests pass locally; exact hosted DB-backed CI remains required.

davletd (approved):
Deep review on exact head d6e536b42: bounded reaper admission now matches the 65-minute timeout budget; regression tests cover sequential draining across scheduled sweeps. Local contract tests, lint, formatting, and diff validation pass. Approval remains conditional on all hosted integration/browser/image/package checks passing for this exact commit.

jorgensandhaug:
Recommended correction shape: replace the fixed one-row batch with a configuration-derived batch capacity. Compute each drain slot from sandboxArchiveCaptureTimeoutMs plus an explicit per-drain overhead allowance; subtract explicit prelude and cleanup reserves from the activity deadline; admit floor(available divided by slotCost), failing startup if that is below one. Record activity start time and, after the prelude, refuse to begin provider capture if the prelude has already consumed its reserve, so the proof is about actual remaining time rather than an optimistic constant. This keeps default configurations capable of draining many boxes per sweep while expensive snapshot settings naturally admit fewer. Enforce the same validator for every control worker that can poll the activity, independent of schedule ownership. Tests should cover default multi-row throughput, near-boundary one-row capacity, prelude exhaustion, non-schedule-owning control startup, and eventual backlog without cross-test global sweep races.

jorgensandhaug:
The latest head d6e536b changes only sandbox-lease.test.ts to loop across repeated sweeps. It does not address either implementation finding or remove the fixed one-row throughput bottleneck. Please do not treat a passing retry loop as resolution. The implementation must become configuration-derived and time-budget-aware, enforce admission after the actual prelude elapsed time, and validate every control worker that can execute the activity. Until then this remains release-blocking.

davletd (approved):
Exact-head deep review for 94759efc5: resolves the two release blockers. Drain capacity is now configuration-derived from a 65m activity minus explicit 5m prelude and 5m cleanup reserves, with capture-fence + 2m overhead slots (default capacity 13; near-boundary capacity 1; zero slots fail startup). Real prelude elapsed time is measured and capture admission fails closed once its reserve is consumed. Every control worker validates before polling regardless of schedule ownership, and the activity validates again for embedded/direct execution. Default multi-row and near-boundary backlog regressions added. Local focused tests, all 23 typechecks, lint/format/diff checks, changeset plan, and worker/workflow bundle pass. Merge remains gated on exact-head hosted DB, recovery, browser, package, deployment, and image checks.

jorgensandhaug:
Superseded by #1186. The replacement retains this PR's 65-minute activity deadline work but fixes the fixed-one throughput regression, accounts for actual elapsed prelude time, derives batch capacity from the configured durable capture fence, and validates every poll-capable control-worker construction path. #1186 has local DB-backed coverage and independent review PASS.

jorgensandhaug:
Closing in favor of #1186 so only the corrected timeout contract can merge.

PR #1186, opened 2026-08-07
Fix sandbox reaper activity budget proof

jorgensandhaug:
Supersedes #1183 with the complete timeout contract.\n\n- reserves explicit prelude, per-drain settlement, and cleanup budgets inside the 65-minute activity deadline\n- derives multi-drain throughput from the configured durable capture fence instead of fixing the batch to one\n- uses a monotonic elapsed-prelude gate and starts zero provider captures after that budget is spent\n- validates every poll-capable control-worker construction path before opening Temporal\n- preserves the existing rule that Connected Machines are detached but never provider-stopped\n\nVerification:\n- focused timeout and embedded lifecycle tests: 19 pass\n- real database-backed sandbox lease and warm meter tests: 37 pass\n- worker typecheck, oxlint, oxfmt, changeset status, and diff check: green\n- independent adversarial autoreview: PASS after fixing its lower-level worker-factory finding


davletd (approved):
Independent exact-head review on a77ae63a9: the replacement correctly resolves #1183's blockers. Capacity is derived from the configured durable capture fence plus explicit prelude/per-drain/cleanup budgets (default 18, expensive 2, near-boundary 1, unsafe 0 rejected); actual prelude admission uses an injected monotonic clock; every poll-capable control worker validates before Temporal connection; the workflow-reachable contract remains dependency-free. Locally revalidated 56 focused worker tests, all 23 typechecks, lint/format/diff, and the published worker + deterministic workflow bundle. Hosted PostgreSQL shards, recovery, packages, deployment and workload images are green to date; merge remains gated on the two exact-head jobs still running.

PR #1017, opened 2026-08-04
feat(machines): safely remove connected enrollments

jorgensandhaug:
## Summary

Adds a first-class, workspace-authorized removal lifecycle for self-hosted Connected Machine enrollments.

- Removes only self-hosted enrollments; Modal/session sandboxes cannot be targeted.
- Works while the agent is offline.
- Revokes the enrollment server-side so stale heartbeat/link/reconnect credentials are rejected.
- Preserves sessions, routes, leases, archives, audit history, and workspace history.
- Keeps revoked enrollments out of the default active machine picker while retaining an authorized revoked audit view.
- Adds accessible desktop/mobile confirmation UX with exact dependency blockers and immediate list reconciliation.
- Adds SDK and first-party MCP operation with idempotency and stale-revision protection.

## Safety and concurrency

- FORCE-RLS migration/table and tenant-scoped transaction.
- Workspace admin/operator authorization remains required by the existing enrollment-management surface.
- Removal serializes with attach/route/lease mutations and blocks active routes, commands, live leases, and unreconciled recovery dependencies with actionable responses.
- Duplicate enrollments are isolated by enrollment identity.
- Repeated requests are idempotent only for the same request fingerprint.
- Durable audit event records the removal without hard-deleting evidence.

## Verification

- `bun run typecheck` — all 23 projects clean.
- `bun run format:check` — all files formatted.
- `git diff --check` — clean.
- React machine tests — 30 pass / 97 assertions.
- Real PostgreSQL 16 + pgvector + non-superuser FORCE-RLS DB suite — 3 pass / 24 assertions.
- API enrollment route suite — 15 pass / 95 assertions.
- SDK client contract suite — 37 pass / 186 assertions.
- Chromium browser suite — 2 pass / 8 assertions, including desktop keyboard flow, narrow/mobile layout, accessible labels, confirmation, success reconciliation, and screenshots.

## Production acceptance gate

The two specified offline duplicate enrollments are intentionally **not** mutated by this PR. After merge, staging validation, and identical-digest production deployment, acceptance will re-fetch each enrollment, prove it is detached/unreferenced, remove them one at a time through the supported operation, verify active-list absence plus revoked audit visibility and stale reconnect rejection where safely testable, and confirm unrelated Modal/self-hosted machines and sessions remain intact.

Related canonical lifecycle work: OPE-14 / OPE-60. This PR does not modify OPE-165, OPE-27, Capabilities UX, portfolio audit, or unrelated releases.

PR #1313, opened 2026-08-10
fix(machines): safely detach sessions before removal

jorgensandhaug:
## Summary

- return every session that still routes through a connected machine instead of reporting only the first dependency
- add an explicit `moveSessionsToDefaultSandbox` confirmation that atomically clears those route pointers, advances their active epochs, and revokes the machine only after lease/recovery blockers are clear
- render linked dependent sessions and the follow-up action inline in the removal dialog
- treat resolved `swapped: false` sandbox-switch responses as visible mutation errors rather than silent success

This is a focused follow-up to the removal lifecycle landed in #1017. It does not revive or broaden #466.

## Safety

- keeps session history, goals, events, queue state, working directories, sandbox groups, and recovery/archive state intact
- uses the existing epoch fence when moving routes
- retains existing active-command, active-lease, and recovery-pending blockers
- includes the explicit move choice in the idempotency fingerprint
- no deployment changes

## Validation

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun test packages/react/test/use-machines.test.tsx apps/web/src/routes/machines-removal.test.tsx packages/sdk/test/client.test.ts -t 'useMachines|connected machine removal conflict|removeEnrollment posts'`
- `bun run build` in `packages/sdk`
- `bun run build` in `packages/react`
- `bun run build` in `apps/web`

The focused DB/API integration cases are included, but this local sandbox has no Docker/PostgreSQL fixture. The repository-wide unit sweep also reaches unrelated suites that require Docker/PostgreSQL and `@temporalio/workflow`; those setup-only failures are left to service-backed CI.

### DIFF
--- a/.changeset/safe-machines-detach.md
+++ b/.changeset/safe-machines-detach.md
@@ -6,4 +6,4 @@
 "@opengeni/sdk": patch
 ---
 
-Make connected-machine removal show every dependent session and support an explicit atomic move-to-default-sandbox confirmation before revocation. Surface typed sandbox swap rejections as visible React mutation errors instead of treating resolved failures as success.
+Make connected-machine removal show every dependent session and support an explicit canonical move-to-default-sandbox confirmation before revocation. Default moves prove managed sandbox readiness through the existing fleet route, active turns remain fail-closed, and typed swap rejections surface as visible errors instead of false success.
--- a/.changeset/safe-machines-detach.md
+++ b/.changeset/safe-machines-detach.md
@@ -1,0 +1,9 @@
+---
+"@opengeni/api-router": patch
+"@opengeni/contracts": patch
+"@opengeni/db": patch
+"@opengeni/react": patch
+"@opengeni/sdk": patch
+---
+
+Make connected-machine removal show every dependent session and support an explicit atomic move-to-default-sandbox confirmation before revocation. Surface typed sandbox swap rejections as visible React mutation errors instead of treating resolved failures as success.
--- a/apps/api/src/mcp/server.ts
+++ b/apps/api/src/mcp/server.ts
@@ -2407,22 +2407,20 @@ function registerConnectedMachineTools(
     "connected_machine_remove",
     {
       description:
-        "Remove one enrolled self-hosted machine while it is offline. Access is revoked, future heartbeat/reconnect credentials are rejected, session/route/lease/archive history is retained, and a fresh human-approved device-flow enrollment is required to reconnect. Pass the enrollmentId from the Machines surface, never a Modal sandbox id. Blocked outcomes include every dependent session and the action needed before retrying. Set moveSessionsToDefaultSandbox only after explicitly reviewing those sessions; the removal then atomically repoints them before revocation.",
+        "Remove one enrolled self-hosted machine while it is offline. Access is revoked, future heartbeat/reconnect credentials are rejected, session/route/lease/archive history is retained, and a fresh human-approved device-flow enrollment is required to reconnect. Pass the enrollmentId from the Machines surface, never a Modal sandbox id. Blocked outcomes include every dependent session and the action needed before retrying. Move each dependent session through the canonical sandbox_swap target=default path before retrying removal; the removal authority never rewrites routes directly.",
       inputSchema: {
         enrollmentId: z4.string().uuid(),
         expectedUpdatedAt: z4.string().datetime({ offset: true }).optional(),
         idempotencyKey: z4.string().trim().min(1).max(200).optional(),
-        moveSessionsToDefaultSandbox: z4.boolean().optional(),
       },
     },
-    async ({ enrollmentId, expectedUpdatedAt, idempotencyKey, moveSessionsToDefaultSandbox }) => {
+    async ({ enrollmentId, expectedUpdatedAt, idempotencyKey }) => {
       const result = await removeEnrollment(deps.db, {
         accountId: grant.accountId,
         workspaceId: grant.workspaceId,
         enrollmentId,
         operationKey: idempotencyKey?.trim() || randomUUID(),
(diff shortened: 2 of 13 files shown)


jorgensandhaug (commented):
BLOCKED — two concrete correctness fixes are required at exact head `79b170599c821c5d3ee09ae52adc17207e257e3c`.

1. **Block running turns that are actively routed to the machine** (`packages/db/src/index.ts:37373`). The `active_commands` query checks `sandbox_group_id = machine.id`, but ordinary managed-home sessions use their own group and reach the Connected Machine through `active_sandbox_id`. Therefore a session with `active_sandbox_id = machine.id` and non-null `active_turn_id` is included in `dependentSessions`, misses both this query and the self-hosted lease query, and `moveSessionsToDefaultSandbox:true` revokes the enrollment while its accepted machine operation may still be running. The repository’s canonical active-op definition at `packages/db/src/index.ts:38520-38547` is exactly active pointer + non-null active turn. Use that condition (under the same locks) as the blocker before detachment/revocation.

2. **Do not treat the raw `active_sandbox_id = NULL` update as a successful default-sandbox move without the canonical destination checks** (`packages/db/src/index.ts:37527`). `swapActiveSandbox(..., "default")` deliberately proves the session group is ready before publishing the route (`packages/core/src/sandbox/fleet.ts:463-490`), but this path skips that contract for every dependent session. It also strands machine-home sessions: they retain `sandbox_backend = selfhosted`, and Channel-A rejects a null pointer with `409 "machine-home session has no active Connected Machine"` (`apps/api/src/sandbox/channel-a.ts:541-552`) even though the removal response/UI says the session now uses a managed default. Preflight/hold each real default using the canonical readiness semantics and either perform a supported machine-home authority migration or return a blocker for those sessions; do not revoke and report success while REST/SDK sandbox operations are unusable.

Evidence: full diff and routing/removal contracts inspected in an isolated detached worktree. `packages/sdk/test/client.test.ts` passed 41 tests. The focused DB/React tests could not start in the detached worktree because the owner checkout’s ignored Bun dependency symlinks are not portable; no candidate files were modified.

jorgensandhaug (commented):
BLOCKED — the V2 active-turn repair is correct, but one substantive machine-home/default-route blocker remains at exact head `361cee6199ceffdf576cff46a5517ced7f701dc8` (tree `28ed7a63f3a1da92b3c76e6d61daf17de4d4a792`).

**Do not report a usable managed-default move for machine-home sessions by only setting `active_sandbox_id = NULL`** (`packages/db/src/index.ts:37566-37606`). A top-level session created on a Connected Machine persists `sandbox_backend = "selfhosted"`. After this force path clears the pointer and revokes the enrollment, worker turns happen to substitute the deployment cloud backend when `machinePrimary` is false (`apps/worker/src/activities/agent-turn.ts:5436-5450`), but the public API surfaces do not make that substitution: Channel-A branches on the persisted selfhosted backend and returns 409 when the pointer is null (`apps/api/src/sandbox/channel-a.ts:541-552`), while viewer readiness acquires/establishes the group with `session.sandboxBackend` (`apps/api/src/sandbox/viewer.ts:220,287,363`), which reaches the selfhosted provider without a bound agent id. Consequently SDK terminal/files/git/viewer operations remain unusable even though the removal response and UI say the session now uses a managed default.

This path also bypasses the canonical default swap's readiness hold, which is intentionally completed before publishing the null pointer (`packages/core/src/sandbox/fleet.ts:463-504`). Minimal acceptable fix: either return a typed blocker for machine-home dependents until a supported home-authority migration exists, or migrate them to a real managed backend/default authority and prove that destination under the canonical readiness semantics before revocation. Preserve the current V2 `active_turn_id` fail-closed ordering.

Focused evidence: V2 now locks all dependent active-pointer rows and blocks on non-null `active_turn_id` before route/epoch/enrollment mutation; DB package typecheck passed; 57 targeted tests reported pass, with PostgreSQL assertions self-skipped because Docker was unavailable. The remaining blocker follows directly from the divergent worker versus API backend-selection code above.

jorgensandhaug (commented):
PASS — exact head `ff7a2743731ffb53732c5c5dc4ec40f7736c3a3f` (tree `efea365444d50d1bfc5ae7f0a37ece73473e9318`) resolves both prior blockers without weakening the removal boundary.

The transactional path now locks every active pointer to the selected machine and checks `active_turn_id` first, so accepted/running work cannot be detached. It then returns the typed `machine_home` blocker for a self-hosted-home session before lease/route/revocation mutation. Ordinary managed-home dependents are no longer rewritten by the removal transaction: the web flow moves each session through canonical `swapActiveSandbox(..., { target: "default" })`, which proves managed-home readiness before the epoch-fenced CAS, stops visibly on `swapped:false`, and only then invokes a fresh removal transaction that re-locks and rechecks dependencies. The enrollment-vs-swap lock ordering also makes concurrent reattach/remove fail closed.

The removal itself changes only the selected enrollment lifecycle state and preserves session rows, history, workspace/group identity, route/archive evidence, and machine records. Receipt fingerprinting remains scoped to enrollment plus optional revision; stored pre-field results remain readable with `dependentSessions: []`. REST requires `enrollments:manage`, canonical swaps require `sessions:control`, MCP remains explicitly high-trust and does not gain a direct route-rewrite flag, and contracts/SDK expose the additive `machine_home`/dependency response consistently. Inline UI lists every affected session and preserves actionable API or `swapped:false` errors.

Verification in a clean detached worktree: `bun run typecheck` passed all 25 projects; 75 focused tests across API, web, React, SDK, and DB files reported pass; `git diff --check` passed. The six DB lifecycle cases self-skipped their PostgreSQL assertions because Docker is unavailable in this reviewer sandbox, while the exact-head GitHub real-service/recovery check was successful when reviewed.

PR #1321, opened 2026-08-11
Finalize session activity at transaction commit

jorgensandhaug:
## Summary

- finalize semantic session activity once per transaction, assigning one workspace revision to every affected session only after constraints settle
- keep raw protocol/event appends on the no-invalidation fast path through an explicit, fail-closed raw-versus-semantic contract
- make session discovery use coherent MVCC fences without taking a workspace mutation lock
- centralize lock ordering, savepoint escalation, bounded transient retries, and sanitized persistence failures
- install workspace-counter initialization before migration backfill so concurrent workspace creation cannot escape the rolling migration
- harden runtime setup ownership, failed-owner recovery, immutable completion proof, and bounded stale-lock handling

## Why

Session mutations previously advanced monitoring activity from individual write paths. That made multi-write transactions harder to reason about, created unnecessary contention, and left correctness dependent on every caller remembering the same revision protocol. Discovery could also participate in mutation-oriented coordination despite being read-only.

This change makes commit the single authority for semantic activity. Raw append-only protocol traffic remains cheap, while any caller that declares a raw-only operation but attempts a semantic change fails atomically.

## Verification

- full `bun prep`
- real PostgreSQL rolling/current/historical migration replay
- concurrency, deadlock, serialization, savepoint, and fail-closed fault coverage
- real NATS and Temporal workflow/worker-death recovery coverage
- complete browser, accessibility, reflow, Docker sandbox, rig-setup, and local desktop-image gates
- structured independent review of the final commit: no actionable findings


jorgensandhaug (commented):
```json
{
  "version": 3,
  "reviewedBaseSha": "a57f1eecc3ce1333a52f77d56d730a0c9465616f",
  "reviewedHeadSha": "45347acdaa48ad7e6b72894e125c7e16da88d9d2",
  "verdict": "PASS"
}
```

PR #1357, opened 2026-08-12
feat: add canonical human identity authority

jorgensandhaug:
## Summary

- add organization-independent canonical human identities and multiple verified login bindings
- add revision-CAS, immutable operation receipts, collision containment, lost-factor recovery, and same-transaction Better Auth session invalidation
- stamp and validate managed sessions against canonical identity/auth revisions
- expose metadata-minimal `/v1/identity` lifecycle routes with no organization, workspace, membership, or resource metadata
- classify the four identity authority tables as FORCE-RLS/no-direct-DML with exact SECURITY DEFINER runtime capabilities
- synchronize already-verified Better Auth accounts idempotently before session creation
- export the canonical identity contracts/database surface and document the identity-vs-tenancy authority boundary

## Final integration

- rebased onto protected `main` at `942f8eb8e48a2d01e8a25c3f955e860391658717`
- preserves authoritative migration `0222_session_visibility_authority_epochs.sql`
- retains OPE-194's reserved migration `0223_canonical_human_login_bindings.sql`
- updates the governed release-schema contract to 226 migrations with 0223 as the latest rolling migration
- integrates the previously fenced shared barrel exports and architecture documentation
- intentionally leaves Better Auth raw-SQL-owned tables out of the Drizzle schema barrel

## Validation

- clean full SQL migration ledger through `0223_canonical_human_login_bindings.sql` on PostgreSQL 17.10 — 226 migrations
- PostgreSQL-focused 0223 and runtime-posture suites together — 25 pass, 0 fail, 136 expectations
- release-schema contract plus runtime-posture suites — 25 pass, 0 fail, 227 expectations
- dedicated PostgreSQL-backed 0223 suite after formatting — 6 pass, 0 fail, 54 expectations
- package typechecks: `@opengeni/contracts`, `@opengeni/db`, `@opengeni/core`, `@opengeni/api-router`
- all branch-changed TypeScript files: `oxlint --deny-warnings` — 0 warnings/errors
- all branch-changed source/docs files formatted with `oxfmt`; `git diff --check` clean

No deployment, release, cloud/provider, production, ruleset, merge, or unrelated PR action is part of this PR.

Linear: OPE-194

## V2 managed-session fixture repair

- preserves fail-closed production validation for missing, unstamped, and stale Better Auth session IDs
- upgrades the preference-registry and API integration managed-auth adapters to seed real verified auth accounts, synchronize canonical identity authority, insert revision-stamped sessions, and return the exact session ID
[...]

PR #1359, opened 2026-08-12
fix(db): stage session-channel FK validation

jorgensandhaug:
## Summary

- keep migration 0220's `sessions.channel_id` addition metadata-only
- add the channel foreign key as `NOT VALID`, then validate it in a separate rolling migration
- preserve databases that already applied the original auto-named validated foreign key
- align the turn-density release-gate expectations with the active-history serving envelope already on `main`

## Testing

- [x] `bun run typecheck`
- [ ] `bun test` (full suite pending CI)
- [x] `bun test packages/db/test/migration-0220-session-channels.test.ts scripts/release-schema-contract.test.ts`
- [x] `bun test scripts/operator/turn-density-profile.test.ts`
- [x] `bunx changeset status --output /tmp/changeset-status.json`
- [x] `bun run check:public-hygiene`
- [x] `bun run format:check`
- [x] `git diff --check`

## Delivery integrity

- [x] Candidate/version labels represent substantive source changes, not base-only refreshes.
- [x] If `main` advanced, I kept the candidate head frozen and verified current-base mergeability/material compatibility, or documented the actual conflict/incompatibility that required a source change.

## Notes

- The real-PostgreSQL migration test covers both a fresh database and a database carrying the original `sessions_channel_id_fkey` plus migration receipts for 0220/0221.

### DIFF
--- a/.changeset/quiet-channel-fk-validation.md
+++ b/.changeset/quiet-channel-fk-validation.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/db": patch
+---
+
+Stage and validate the session-channel foreign key without retaining the column-addition lock across a populated sessions-table scan.
--- a/apps/worker/test/retained-process-reconciliation.test.ts
+++ b/apps/worker/test/retained-process-reconciliation.test.ts
@@ -939,7 +939,7 @@ describe("retained-process terminal-owner reconciliation", () => {
     expect(candidatePlan).toContain('"Node Type":"Limit"');
     expect(candidatePlan).toContain('"Actual Rows":7');
     expect(candidatePlan).toMatch(
-      /session_turn_attempts_(?:workspace_id_uq|pkey|latest_session_idx)/,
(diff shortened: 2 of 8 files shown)


jorgensandhaug:
TENANCY_MIGRATION_COLLISION_FENCE_V1

Merge blocker as of August 12, 2026: protected `main` already owns `0222_session_visibility_authority_epochs.sql`, while OPE-194 PR #1357 owns reserved `0223_canonical_human_login_bindings.sql`. This PR still introduces `0222_sessions_channel_fk.sql` and `0223_sessions_channel_fk_validate.sql`. Please renumber both migrations and all tests/release-schema references after the current live reservations: 0224 PR #1330, 0225 session activation, 0226 personal Codex authority, and 0227 user-scoped variable sets/rigs. `0220_session_channels.sql` is already on protected main and must remain history-identical rather than being reintroduced or rewritten.

PR #1360, opened 2026-08-12
fix(runtime): avoid model polling for short commands

benyhh:
## Summary

- preserve the 250 ms provider/cancellation boundary while honoring the model-requested command wait internally
- return short command completion from the original tool call instead of requiring model-only `write_stdin` polls
- validate function-transport `view_image` data URLs before constructing structured image content, returning a safe conversion instruction for BMP or mismatched bytes
- document the updated shell and image lifecycle contracts

Tracks OPE-212.

## Verification

- `bun run typecheck`
- `bun run lint`
- 453 focused runtime and worker tests passed, 3 platform-specific tests skipped
- real local-process cancellation tests passed

## Linux Terminal-Bench 2 rerun

Exact source: `77e1da3adef9199c8639a60187f94319d9f08e72`
Model: GPT-5.6 Sol, high reasoning, Codex subscription authentication only

| Task | Result | Billed tokens | Model time | Model-visible `write_stdin` |
| --- | ---: | ---: | ---: | ---: |
| Scheme evaluator | pass | 978,589 | 9m 01s | 20 |
| MIPS interpreter | pass | 1,389,879 | 9m 34s | 1 |

Compared with the previous OpenGeni run, total billed tokens fell from 2,898,571 to 2,368,468 (-18.3%), total model time fell from 23m03s to 18m35s (-19.4%), and hidden-verifier solves improved from 1/2 to 2/2. The Scheme run itself regressed because the model created and monitored a genuinely long-running test process, so this two-task sample does not establish a uniform per-task latency win.

The MIPS BMP inspection returned a safe conversion instruction, the agent converted it to PNG, and the turn completed without the previous provider HTTP 400. The hidden verifier scored 1.0.

Direct Codex remains more efficient on this sample: 1,796,432 billed tokens and 12m26s for 2/2 solves versus OpenGeni's 2,368,468 and 18m35s.


benyhh:
## Benchmark overview

| Run | Solved | Billed tokens | Model time |
|---|---:|---:|---:|
| Previous OpenGeni | 1/2 | 2,898,571 | 23m03s |
| New OpenGeni | 2/2 | 2,368,468 | 18m35s |
| Direct Codex | 2/2 | **1,796,432** | **12m26s** |

New OpenGeni versus the previous OpenGeni run: **18.3% fewer billed tokens** and **19.4% less model time**, while improving from 1/2 to 2/2 solved. Direct Codex remains the efficiency leader.

Changes aimed at billed-token and model-time efficiency:

- Short shell commands can now settle inside the original tool call, avoiding model turns spent polling every 250 ms.
- Longer shell waits are handled internally in cancellation-responsive slices, reducing repeated `write_stdin` calls without weakening cancellation safety.
- Unsupported or mismatched image bytes are rejected before the provider request, avoiding wasted failed inference calls and recovery turns.
- Image failures return concise conversion guidance, giving the model a direct corrective action instead of exploratory retries.

These are single runs per task, so repeated trials are still needed before treating the percentage improvements as stable.

PR #1356, opened 2026-08-12
test: wait on Slack acceptance UI readiness

davletd:
## Summary

- replace `networkidle` navigation waits in Slack installation-binding acceptance with `domcontentloaded` while retaining UI-visible readiness and security assertions
- keep changed E2E files in the E2E tier instead of also handing their custom suffixes to Bun's unit runner
- derive synthetic timeline tip geometry from its owned layout state so unrelated process-global DOM stubs cannot create a false 24 px gap
- wait for the requested control-center text state itself after asynchronous feature mutations
- unblock the next Changesets Version PR without retry-to-green

## Evidence

- reproduced the Version PR's same-URL `networkidle` timeout
- Slack browser acceptance passed 5/5 before the PR and again after the complete fix (13 assertions per run)
- impact planner regression suite: 34 passed, including changed-E2E tier ownership
- message timeline suite: 10/10 consecutive complete-file passes with the exact bottom-follow assertion retained
- complete custom-API browser suite passed, then the formerly failing Gmail feature pass passed 10/10 additional runs
- formatting and `git diff --check` pass

These are test synchronization, selection, and synthetic-layout corrections; no product assertion is skipped or weakened.

PR #1363, opened 2026-08-12
test: align density gate with history envelope

davletd:
## Summary

- align the turn-density release-gate expectations with the active-history envelope raised by #1358
- keep the negative artifact-verification case noncanonical by mutating the row limit back to the old value

## Validation

- `bun test scripts/operator/turn-density-profile.test.ts` (18 pass)
- `git diff --check`

This is intentionally separate from release-unblocker #1356 so that candidate remains immutable.


davletd:
Superseded by merged #1359 (`3f816083f`), which contains this exact density-envelope assertion update and the related authority-epoch planner expectation. Closing to avoid duplicating already-merged main fixes.

PR #1414, opened 2026-08-13
fix(web): make agent children visible by default

benyhh:
## Summary

- Automatically expand and lazy-load matching Active agent branches so child work is visible by default.
- Remove the depth and children selectors plus the technical browser-memory warning.
- Make child and root pagination single-flight, progress-truthful, retryable, and stable across refreshes.

## Verification

- `bun test apps/web/src/lib/agent-topology.test.ts` (10 passed)
- `bun run --cwd apps/web typecheck`
- `bun run lint`
- `bun run format:check -- apps/web/src/routes/agents.tsx apps/web/src/lib/agent-topology.ts apps/web/src/lib/agent-topology.test.ts`
- `bun run --cwd apps/web build`
- Browser harness: 200 tree items rendered, collapse/expand restored 1/200 items, Depth 1 and Depth 2 children visible, Outline/Tree switch verified, no obsolete selectors or technical warning, and no warning/error logs.

Linear: OPE-189

### DIFF
--- a/apps/web/src/lib/agent-topology.test.ts
+++ b/apps/web/src/lib/agent-topology.test.ts
@@ -4,10 +4,12 @@ import type { AgentTopologySession } from "@opengeni/sdk";
 import {
   AGENT_DIAGRAM_NODE_HEIGHT,
   AGENT_DIAGRAM_NODE_WIDTH,
+  agentHasMatchingDescendants,
   buildAgentTopology,
   filterAgentTopology,
   layoutAgentTopologyDiagram,
   limitAgentTopology,
+  mergeAgentTopologySessions,
   summarizeAgentTopology,
 } from "./agent-topology";
 
@@ -62,7 +64,10 @@ describe("agent topology", () => {
 
   test("active filtering retains the ancestor path", () => {
     const root = session("root");
-    const child = session("child", { parentSessionId: "root", status: "running" });
+    const child = session("child", {
+      parentSessionId: "root",
+      status: "running",
+    });
     const filtered = filterAgentTopology(buildAgentTopology([root, child]), "active", "");
     expect(filtered[0]?.session.id).toBe("root");
     expect(filtered[0]?.children[0]?.session.id).toBe("child");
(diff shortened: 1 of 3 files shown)


jorgensandhaug (commented):
OPE189_PR1414_EXACT_HEAD_REVIEW_V1_BLOCK

Exact reviewed head: `20fd9263f0cc89c01877e6c81213eb4aeab6fbc5`
Exact tree: `afdc41286e5b0210bea27c613f3062271553e98a`

COMMENTED BLOCK — three valid source findings remain:

1. **Auto-expand does not enforce its declared concurrency ceiling.** The effect slices four candidates per render, but each `loadChildren` immediately updates `branchPages`, causing another render/effect pass while the previous requests remain in `branchRequests`. Successive waves can therefore launch until every candidate is in flight, rather than holding at `AUTO_EXPAND_CONCURRENCY = 4`.
2. **Periodic first-page refresh can race root pagination and regress progress.** `refresh()` rejects a second cursor request but permits a no-cursor interval refresh while `rootPageRequest.current` is active. If that refresh settles after the page request, it overwrites `nextCursor`/`hasMore` with first-page metadata, so the next click can refetch an already loaded page.
3. **Root pagination remains actionable at the 200-agent client ceiling.** The root Load-more control is rendered from server `hasMore`/`nextCursor` alone, and cursor reads do not short-circuit at `MAX_LOADED_AGENTS`. Once child/root loading fills the client cap, repeated clicks can issue requests that merge no new rows, contrary to the PR's progress-truthful/no-op-control claim.

Required repair: one focused source revision that globally accounts for in-flight branch requests, fences no-cursor refresh against root-page work, and hides/short-circuits root pagination at capacity, with deterministic tests. Preserve moving-main compatibility without rebasing unrelated protected-main commits.

jorgensandhaug (commented):
OPE189_PR1414_EXACT_HEAD_REVIEW_V1_AMENDMENT

Amendment to COMMENTED BLOCK review `4927733152` on exact head `20fd9263f0cc89c01877e6c81213eb4aeab6fbc5`:

**Retract finding 3.** Exact source reinspection confirmed the existing render condition already suppresses root pagination when `data.sessions.length >= MAX_LOADED_AGENTS`. That item was an incorrect reading of the earlier patch excerpt and is not a valid defect.

The two remaining valid findings are unchanged:
1. auto-expand successive effect waves can exceed the declared four-request concurrency ceiling;
2. a no-cursor periodic refresh can race an in-flight root page and regress pagination metadata.

The focused repair is narrowed to those two items only. No source change will be made for the retracted finding.

jorgensandhaug (approved):
OPE189_PR1414_EXACT_HEAD_REVIEW_V2_PASS

APPROVE — exact reviewed head `8da59478ea3cb25daeff7194db96804da23a1fad`, tree `87339f745605c7b0bae99d21cc39159289b8c614`.

The two valid V1 findings are resolved by the focused repair:
- root first-page refresh and root pagination share one request fence, preventing cursor regression;
- automatic child expansion accounts for all globally in-flight branch requests before filling the four-request ceiling.

Verification:
- deterministic topology suite: 12/12 pass;
- web typecheck, repository lint (0 warnings/errors), focused format, production web build and bundle budgets pass;
- Source Admission `31707176297` passed;
- CI `31707180301` passed with 16 successful jobs, 12 intentional skips, and no bad leaf conclusion;
- protected main remains `f158ed565f96b4bb4b8b391ef25c5c1299329b5a`;
- GitHub merge commit `0c533d86551cc2385be4b8f576b8f49122d2c85b` has integrated tree `18e07fcab867bb90cb937955832024e2723c4224`;
- source and integrated patches share stable patch ID `b84ac570c49bd10dd82cb28420e3b11ba9fd7efc`, with zero changed-path overlap against protected-main drift;
- exact integrated-commit topology tests 12/12, web typecheck, production build, and bundle budgets pass.

No unresolved valid review findings remain. Current-main compatibility was verified without source mutation.

PR #1455, opened 2026-08-14
Add compact session browse filters and grouping

benyhh:
## Summary
- add one compact filter control beside session search
- group loaded sessions by last activity, created date, or creator
- filter by frozen creator plus either created date or last activity over today, 7-day, or 30-day windows
- preserve the current hierarchy and channel layout until a browse choice is active, then use an explicit flat browse projection
- retain pagination with an honest loaded-session note and a Load older path when the current page has no matches

## Verification
- 116 session rail, grouping, browse, channel, and app tests
- web typecheck
- oxfmt and oxlint on touched files
- browser verification against the live local API, including active creator grouping

Linear: OPE-238

### DIFF
--- a/apps/web/src/components/rail/session-list.tsx
+++ b/apps/web/src/components/rail/session-list.tsx
@@ -98,8 +98,7 @@ import {
   groupSessionsForRail,
   mergeSessionForRail,
   relativeTimeLabel,
-  sessionCreatorKey,
-  sessionCreatorLabel,
+  sessionCreatorOptions,
   visibleForestRows,
   visibleTreeRows,
   type SessionTreeNode,
@@ -335,13 +334,7 @@ export function SessionList() {
     return [...source.values()];
   }, [pinOverrides, serverSessions]);
   const creatorOptions = useMemo(() => {
-    const byKey = new Map<string, string>();
-    for (const session of allSessions) {
-      byKey.set(sessionCreatorKey(session), sessionCreatorLabel(session));
-    }
-    return [...byKey.entries()]
-      .map(([value, label]) => ({ value, label }))
-      .sort((left, right) => left.label.localeCompare(right.label));
+    return sessionCreatorOptions(allSessions);
   }, [allSessions]);
   const browseSessions = useMemo(
     () =>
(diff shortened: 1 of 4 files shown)


jorgensandhaug (changes requested):
Exact-head review of `563518a4ce3a9e3bce5ada898236fdb0fd6a6869` — **changes requested**.

Blocking finding: creator disambiguation is inconsistent across the same browse view. `sessionCreatorOptions(allSessions)` builds picker labels from every loaded creator, but `groupSessionsForBrowse(...)` recomputes collisions from the filtered, non-pinned subset. With two loaded creators whose frozen label is `Alex`, the picker shows identity-qualified labels; after selecting one creator (or filtering/pinning the other out), the group heading falls back to plain `Alex`. This is visually and accessibly inconsistent and defeats the collision repair.

Required repair: derive one canonical creator-label map from the complete loaded creator set and pass/reuse it for both picker options and creator group headings; add a regression where the complete set contains two duplicate labels but the grouped subset contains only one.

Verified on this head: the opaque subject-ID repair remains correct; the creator submenu is height-bounded and vertically scrollable; focused browse/pagination/channel/control tests and focus/pin/tree tests pass; web typecheck and `git diff --check` pass. The explicit flat browse projection is intentional per PR scope and is not part of this change request.

`PR1455_EXACT_HEAD_REVIEW_V2`

jorgensandhaug (approved):
Exact-head approval for b8bfc49ede6182e5f5c4f76ab9668c06e0312337 (tree 10122ce52f615e3a6c9f18cf4eff6c3bfaa27f01).

Independent fresh review: PASS with no substantive findings. The final repair derives one canonical creator-label map from the complete loaded session set and reuses it for both the picker and filtered/unpinned creator groups; opaque creator IDs, flat browse hierarchy, pagination, accessibility, and the bounded scrollable menu are preserved. The regression covers duplicate frozen labels in the complete set where the filtered non-pinned subset contains only one identity.

Validation: focused and integrated web tests/typecheck/format/lint/diff checks and production build/bundle budgets passed; current-main integration against 38438255fc054f98734eacd70615f742e452fbef is conflict-free and patch-identical. Exact-head Source Admission and all required CI are green. The one initial browser-knowledge timeout was unrelated to this delta and passed on rerun at the unchanged SHA.

PR1455_EXACT_HEAD_REVIEW_V3

PR #1524, opened 2026-08-17
feat(catalog): move connector curation into a committed data overlay (OPE-252)

benyhh:
## Summary

Moves connector curation out of the import script and into a committed data file, `data/catalog/curated.json`, so that adding or correcting a connector is a reviewable data diff rather than a TypeScript change. First slice of the integration foundation.

- **Migrates the existing in-code overlay** — `officialCatalogContractsByMcpUrl` (Gmail, Slack, Mobbin contracts) and `brandedNamesByMcpUrl` (Linear) — into the data file, keyed by exact MCP URL and validated at module load. The migrated entries produce **byte-identical** import output; all 36 existing import tests pass unchanged.
- **Adds three curated fields**: `category` (persisted on the row, replacing the hardcoded `"integrations"`), `featured`, and `official` (persisted under `metadata.curation`).
- **Curates 25 real snapshot rows** to fix aggregator names: `stripe`, `wrangler` (the CLI, not the server), `sentry-cli`, `clickup20`, `zapier.com – nla`, `Lucidtech API` (a mislabelled Webflow), `Box Platform API`, `Dropbox Dash`, and the bare domain `front.com`.
- **UI**: featured connectors lead the Browse grid via a stable sort; `official` renders a small marker on the tile meaning *the provider publishes this server on its own domain*. There is deliberately **no "Reviewed" or "Verified" label anywhere** — we have no review process, and both flags are checkable claims, not trust signals.

## What this does not do

- Does not add snapshot rows for providers absent from `integrations-snapshot.json` (Notion, HubSpot, Intercom, Atlassian, GitHub). Those need a real probe via `bun run catalog:refresh`, not fabricated curated data. Tracked as a follow-up on OPE-252.
- Does not build a distinct Featured strip or the connector detail sheet — that is OPE-256's redesign. This PR only makes the data exist and lead the grid.
- No schema migration. `category` was already a column; curation flags ride in the existing `metadata` jsonb.

## Deployment impact

None to the running workspace. Re-running `catalog:import` is idempotent. `CATALOG_IMPORT_SEMANTIC_VERSION` is bumped to 2 so the `--if-changed` fingerprint refreshes curated rows on next deploy.

## Verification

- `bun test scripts/catalog-curation.test.ts scripts/import-integrations-catalog.test.ts` — 52 pass
- `bun test apps/web/src/lib/capabilities.test.ts` — 76 pass
- `bun test packages/db/test/catalog-imports.test.ts apps/web/src/routes/capabilities.test.tsx` — 16 pass
- `bun run typecheck` — all projects clean
- `bun run lint`, `bun run format:check`, `bun run check:docs-refs` — clean

Linear: OPE-252, parent OPE-16.


benyhh:
Independent review round addressed in `acb99e146`.

**Critical fix:** the `--if-changed` import fingerprint did not include the curated overlay, so an overlay-only PR would have been a silent no-op on deploy. The overlay's canonical parsed form is now hashed in; whitespace-only reformatting does not churn it, any semantic change does. Covered by a fingerprint test.

**Also fixed:** parser now rejects unknown keys and non-canonical MCP URLs; the importer reports and the CLI fails on curated entries that match no snapshot row; `canonicalMcpUrl` has a single source; category slugs render through a display-label map; `metadata.curation` is excluded from catalog search; the vacuous `official` test is replaced with a real host-under-domain check.

That last check immediately caught two entries whose official server lives off the snapshot domain (Sentry on `sentry.dev`, Front on `frontapp.com`). Both are no longer marked `official`. The badge means "checkable", not "we believe it".

The earlier red `Browser and visual acceptance (interaction)` check was a Codex settings-page e2e timing out on a "Hide details" toggle. It failed identically on `main` at `a3ed1332f` before this branch existed and touches nothing near the catalog. New head re-runs it.

PR #1528, opened 2026-08-17
feat(slack): two authorities, bot-token search scopes, no workspace proxy identity (OPE-257)

benyhh:
## Summary

Slack goes from three authorities to two. The **workspace-owned hosted Slack MCP connection** — shared OpenGeni authority over one designated human's Slack grant, i.e. a service account made from a real employee — is removed. What remains:

- **Personal hosted Slack MCP** (subject-owned) — unchanged, byte for byte.
- **OpenGeni workspace bot** (`app_install`) — unchanged, plus three new bot scopes.

The workspace path existed because Slack's `search.messages` accepted only user tokens. Slack's Real-time Search API (`assistant.search.context`, February 2026) accepts **bot** tokens with `search:read.public`, `search:read.files`, `search:read.users` as bot scopes, so workspace-wide public search no longer needs a human proxy.

The whole workspace path was introduced by exactly one commit, `dfcf6980f` (#1240). This restores the fence that existed before it.

## Changes

- `oauth-client.ts`: OAuth start, post-lookup reconnect, and the callback all reject non-personal ownership for `https://mcp.slack.com/mcp` (`assertHostedSlackMcpPersonalOwnership`, `isHostedSlackMcpTarget`). Slack start with omitted `ownership` now 422s, matching the Gmail precedent; the web and e2e always send `personal`.
- `packages/core/src/domain/capabilities.ts`: `validateMcpCapabilityConnectionRef` treats the official Slack URL as personal-only, so `enableCapability` cannot store a workspace-scoped Slack `connectionRef`.
- `packages/contracts/src/slack-bot-scopes.ts`: the three search scopes added as **REQUESTED + SAFE_OPTIONAL, not REQUIRED** — the same pattern as `reactions:read`. An existing installation stays eligible and gains bot search after a reinstall from the new manifest. `channels:join` and `chat:write.public` stay off; no auto-join.
- Web: Slack MCP catalog tile is personal-only ("Connect only for me"); the personal-only notice is parametrised on the item name instead of hardcoding Gmail.
- Docs: `docs/slack-bot.md` intro rewritten to two authorities with the bot-token search rationale and an explicit "do not reinstate a workspace proxy identity" rule; `docs/mcp-surfaces.md`, `docs/architecture.md`, `AGENTS.md`, `.env.example` updated.

## What this deliberately does not do

- **No bot search tool yet.** Scopes are requested but nothing calls `assistant.search.context`. That is a separate slice; until it lands, "agent searches public channels with no personal connection" is not met.
- **No DB CHECK constraint.** Every writer path is now fenced in code (POST/PATCH already rejected `slack.com/oauth2`; OAuth and enable now reject non-personal). A `NOT VALID` constraint is ready if we want the durable guarantee, but adding a migration risks ordinal collision with in-flight PRs.
- Existing legacy rows / capability installations are not auto-cleaned (see below).

## Deployment impact — please check before merge

We are our own only production user. Durable state that could reference the removed path:

1. `connections` rows: `provider_domain='slack.com' AND kind='oauth2' AND subject_id IS NULL`
2. `capability_installations.config.connectionRef` with `providerDomain='slack.com'` and `subjectScope='workspace'`

If both are empty, nothing breaks. If (1)/(2) return rows: disable that capability installation and delete the connection via the normal DELETE route; a personal reconnect mints a fresh subject-owned row.

**Bot reinstall needed** to pick up the search scopes. Existing installation keeps working meanwhile.

## Verification

- 243 focused tests green across api/core/contracts/db/web/e2e-adjacent suites (real Postgres via Docker; DB tests actually ran)
- `bun run typecheck`, `lint`, `format:check`, `check:docs-refs`, `check:public-hygiene` — clean
- `bun run slack:manifest` bot scopes now end with `reactions:read, search:read.public, search:read.files, search:read.users`
- Rebased onto current `main` after #1524 merged; no conflicts

Linear: OPE-257, parent OPE-16.


benyhh:
Independent review round addressed in `d1c6f6b32`.

**Real gap closed:** every fence in the first head was enable-time, so a workspace-scoped Slack MCP installation enabled by an earlier release would have kept executing a shared human token at runtime indefinitely. `listEnabledMcpCapabilityServers` now omits it. Covered by a test that writes the legacy row directly and fails without the fence.

**Also:** omitted `ownership` on a personal-only resource defaults to personal (restores the pre-#1240 fence; the SDK field is optional, so no caller breaks); the Slack curated entry carries `connectionOwnership: personal_only` like Gmail; docs now say what actually ships — scopes requested, bot search tool not yet built, shared Slack search unavailable rather than proxied until it is; the personal-only rule is scoped to the hosted-MCP resource so the separate REST-API integration is not misread as impossible.

**Manifest ordering:** the install URL requests every scope in `OPENGENI_SLACK_BOT_REQUESTED_SCOPES`, so a bot scope the Slack app is not configured for fails the whole install with `invalid_scope`. That is by design — the manifest and the request set are the same list so they cannot drift — and it is the same step reactions:read already required. Documented in `docs/slack-bot.md`: **apply the manifest's bot scopes to the Slack app before deploying this version.**

**Deliberately not in this PR:** the bot search tool that calls `assistant.search.context`. Separate slice; keeps this a clean removal.

Original head was CI 0/44 green; new head re-runs.

PR #1530, opened 2026-08-17
feat(catalog): repair the snapshot refresh and add the missing headline connectors (OPE-252)

benyhh:
## Summary

Two things, and the first is the one that matters.

**`catalog:refresh` on `main` was silently broken.** integrations.sh changed its feed to a flat index with no MCP endpoint URLs. The refresh normalized upstream to zero rows and quietly took the `source_normalized_to_zero_rows` fallback: it re-probed the committed snapshot instead, dropped 11 rows (5 on transient timeouts), and discarded the 264-row skip evidence. Nothing failed; the file just got a little worse.

**Fix, in the pipeline rather than the data:**
- Fetch one integrations.sh discovery document per MCP domain (`/api/{domain}/discovery`, documented in their `llms.txt`) and attach it as the surface docs the importer already parses. Zero upstream candidates now **throws** instead of overwriting the committed snapshot.
- Retain committed rows the index no longer lists and re-probe them like everything else. This is what keeps Gmail and Mobbin — never indexed by integrations.sh — from vanishing and breaking the curated tests. `--no-retain` disables it.
- Bounded transient probe retry (2 retries, 2s/4s) for connection errors, timeouts and 5xx only. 404 and non-MCP bodies are never retried. One refresh without this evicted six live endpoints on one-off DNS failures under concurrency 24; all six answer as MCP when re-probed.

**Then, the actual refresh.** 621 → 644 rows. New headline connectors, all real probed rows, all curated:

| Provider | Row | Featured | Official |
|---|---|---|---|
| Notion | `mcp.notion.com/mcp` | yes | yes |
| HubSpot | `mcp.hubspot.com/anthropic` (new; was skipped `auth_unknown` before) | yes | yes |
| Atlassian | `mcp.atlassian.com/v1/mcp/authv2` (rekeyed `atlassian.net` → `jira.com`) | yes | no — host isn't under `jira.com`, same rule as Sentry |
| Intercom | `mcp.intercom.com/mcp` (new) | no | no |

**Intercom deserves a note.** The aggregator reports its auth kind as `unknown`, which the importer would drop. Intercom's own MCP guide says OAuth (recommended) and the server publishes RFC 8414 metadata with DCR at `mcp.intercom.com/.well-known/oauth-authorization-server`. So the overlay sets `authKind: oauth2` — the same override mechanism Gmail already uses. The row, URL and probe are upstream evidence; only the auth kind is curated, and the curated entry's `notes` say so.

**Not added:** GitHub (in the feed only as openapi/cli/graphql; its MCP surface has no URL), Square (only an SSE endpoint upstream), Google Drive/Calendar (not in the feed at all). Those need either upstream to catch up or first-party bridges — the bridge-kit seam on OPE-16, not curation.

**Churn:** 33 rows added, 9 removed (all dead upstream or rekeyed), ~59 changed (mostly the #1524 overlay now baked into rows). **No previously curated row lost.**

## Deployment impact

None to the running workspace. Re-running the import is idempotent; the fingerprint refreshes on the new snapshot.

## Verification

- `bun test scripts/catalog-curation.test.ts scripts/import-integrations-catalog.test.ts scripts/refresh-integrations-catalog.test.ts apps/web/src/lib/capabilities.test.ts` — 144 pass
- `bun scripts/import-integrations-catalog.ts --snapshot … --dry-run` — importable 644, no `curated_entries_unmatched`
- typecheck, lint, format, docs-refs, public-hygiene — clean
- Rebased onto current `main`

Linear: OPE-252, parent OPE-16.

### DIFF
--- a/data/catalog/curated.json
+++ b/data/catalog/curated.json
@@ -345,6 +345,58 @@
       "description": "Meeting transcripts and summaries.",
       "category": "productivity",
       "official": true
+    },
+    {
+      "mcpUrl": "https://mcp.notion.com/mcp",
+      "name": "Notion",
+      "description": "Search, read, and update pages and databases in the Notion workspaces you connect.",
+      "category": "productivity",
+      "featured": true,
+      "official": true,
+      "documentationUrl": "https://developers.notion.com/guides/mcp/overview",
+      "notes": [
+        "Snapshot name came from the aggregator index ('Notion'; an earlier snapshot said 'Notion API'). Tier, provenance, auth kind, and scopes stay snapshot-derived: integrations.sh detected the server and its OAuth metadata directly.",
+        "official is checkable: mcp.notion.com is under the snapshot domain notion.com. Notion documents this hosted server at developers.notion.com/guides/mcp/overview."
+      ]
+    },
+    {
+      "mcpUrl": "https://mcp.hubspot.com/anthropic",
+      "name": "HubSpot",
+      "description": "CRM contacts, companies, deals, and tickets from the HubSpot account you connect.",
+      "category": "sales",
+      "featured": true,
+      "official": true,
+      "documentationUrl": "https://developers.hubspot.com/docs/apps/developer-platform/build-apps/integrate-with-the-remote-hubspot-mcp-server",
+      "notes": [
+        "Name only plus docs. Tier, provenance, auth kind, and scopes stay snapshot-derived: integrations.sh detected the server (mcp:initialize) and its OAuth metadata.",
+        "official is checkable: mcp.hubspot.com is under the snapshot domain hubspot.com. The /anthropic path is the endpoint integrations.sh detected; HubSpot's remote MCP server documentation is linked as documentationUrl.",
+        "Earlier snapshots skipped hubspot.com as auth_unknown because the aggregator had no credential entry for the surface. The refreshed upstream discovery document now carries hubspot_mcp_oauth_app (oauth2)."
+      ]
+    },
+    {
+      "mcpUrl": "https://mcp.atlassian.com/v1/mcp/authv2",
+      "name": "Atlassian",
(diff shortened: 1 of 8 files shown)


benyhh:
Independent review round addressed in `1fdbde8e8`.

- **Retention now visible in the diff:** snapshot header gains `retention: { candidateRows, retainedRows[], droppedRows[] }`, so an upstream delisting a committed row survives is reviewable. Header regenerated without re-probing; `importRows` are byte-identical.
- **Shrink floor:** the refresh throws if kept rows fall below 50% of the committed file (`--allow-shrink` overrides). Symmetric with the zero-candidate guard, so a mass probe failure can no longer write an empty snapshot that would stale every registry row on import.
- **Testability:** the main flow is now an exported `buildRefreshedSnapshot(...)`; zero-candidate throw, retention wiring, `--no-retain`, `generatedAt` fallback and the floor are all covered hermetically. The 503 test runs the retry loop again with an injected `sleep`, and asserts 404/405/429 are attempted exactly once.
- **Upstream fetch hardening:** `redirect: "manual"`, 30 s `AbortSignal.timeout` on index and discovery fetches, 8 MiB bounded body. (`@opengeni/network` isn't a dependency of the root scripts workspace, so a local cap.)
- Probe re-checks the deadline after a retry sleep and short-circuits to `overall_budget_exhausted`; single-surface discovery docs accepted like the importer does; malformed committed rows and unparseable committed files now fail loudly.
- Doc note: an upstream domain rekey (`atlassian.net` → `jira.com`) yields a new capability id and requires re-enabling.

155 tests green; typecheck, lint, format, docs-refs clean; dry-run import 644 rows, no unmatched curated entries. Earlier red check on the previous head was the sandbox image build's artifact-runtime verification, which failed identically on `main` at `e9aabaa5f` and touches nothing in this PR.

PR #1547, opened 2026-08-17
fix(agent): restore macOS installer release gate

davletd:
## Summary
- stage the branded macOS app icon in the installer smoke mock release
- isolate installer smoke HOME directories so local/runner state cannot preserve an unrelated existing app bundle
- cover the direct GitHub Release fallback with the same icon asset

## Why
The merged 0.1.16 candidate correctly made the icon a required release asset, but the macOS smoke fixture did not stage it. Main Agent CI therefore failed before the agent could be tagged.

## Verification
- `bash -n agent/install/test/ci-install-smoke.sh`
- `cargo build --locked --release -p opengeni-agent`
- `bash agent/install/test/ci-install-smoke.sh` (macOS; full verification/downgrade/tamper flow)
- `sh agent/install/test/macos-bundle-sim.sh` (22 passed)
- `bun test scripts/release-image-workflow-contract.test.ts` (18 passed)

### DIFF
--- a/agent/install/test/ci-install-smoke.sh
+++ b/agent/install/test/ci-install-smoke.sh
@@ -37,6 +37,7 @@ trap 'rm -rf "$work"' EXIT
 mock="$work/mock/agent/latest"
 mkdir -p "$mock"
 cp "$built" "$mock/$asset"
+cp install/OpenGeni-Agent.icns "$mock/OpenGeni-Agent.icns"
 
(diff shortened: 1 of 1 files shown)


davletd:
@jorgensandhaug The 0.1.16 release is waiting on this macOS installer-smoke fixture fix. All three installer smokes are green now; could you review when the remaining builds finish?

PR #1575, opened 2026-08-18
fix(runtime): keep snapshots off first-request startup

jorgensandhaug:
Fixes the measured first-turn startup regression where the 10-second lease heartbeat could begin a Modal workspace snapshot before the first provider request, forcing lazy sandbox setup to wait behind checkpoint maintenance.

- gate periodic snapshots until the first provider transport boundary
- preserve the existing single-flight and turn-end capture gates
- make the enclosing runtime/model-preparation span explicit in the timeline
- promote the complete signed agent 0.1.16 bundle as the default stable pointer

Validated: worker tests, config tests, React timeline tests, affected package typechecks, oxlint, oxfmt, and diff check.

### DIFF
--- a/.changeset/quiet-startup-snapshots.md
+++ b/.changeset/quiet-startup-snapshots.md
@@ -1,0 +1,7 @@
+---
+"@opengeni/config": patch
+"@opengeni/react": patch
+"@opengeni/worker-bundle": patch
+---
+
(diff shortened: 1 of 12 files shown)


jorgensandhaug (commented):
```json
{"version":3,"kind":"opengeni-exact-head-release-review","repository":"Cloudgeni-ai/opengeni","reviewedBaseSha":"418b53195206a9a37d5242a9c7e178533c697eea","reviewedHeadSha":"68b5b20c1bb45f586a561e5173a192e24cca718d","reviewerLogin":"jorgensandhaug","reviewProfile":"exact-head-maintainer-v1","verdict":"PASS"}
```

PR #1260, opened 2026-08-09
test(api): prove Google Drive OAuth isolation

jorgensandhaug:
## Summary
- add an independently runnable, real-PostgreSQL adversarial proof for Google Drive OAuth isolation and secret boundaries
- reject signed Google Drive OAuth states whose return path is not the exact initiating workspace capabilities path
- reject reconnect states unless connection ID and integer version are present together

## Adversarial proof
The deterministic provider double and repository test database cover:
- account/workspace/subject connection and callback authority isolation
- signed-state expiry, tampering, single-use replay, and exact redirects
- PKCE S256 challenge/verifier binding and server-only secret handling
- callback scope and post-provider grant revalidation
- OPE-160 fail-closed partial/metadata-only/unknown grants before identity or persistence
- refresh/access token rotation, stale reconnect version fencing, local idempotent disconnect
- revoked-grant normalization without provider-body disclosure
- raw-secret absence from browser persistence, public metadata, logs, OAuth nonce rows, model/history, sandbox input, source metadata, audit, webhook, and host-export sinks

## Defect reproduced
Before the minimal parser fix, a validly signed state with `returnPath=https://attacker.invalid/oauth-capture` redirected the callback to that origin. The new proof was red with that exact redirect and is green after enforcing the canonical workspace path.

## Validation
- `PATH=/tmp/opengeni-docker-shim:$PATH OPENGENI_REQUIRE_REAL_DB=1 bun test --timeout 180000 apps/api/test/google-drive-oauth-isolation.test.ts apps/api/test/google-drive.test.ts packages/db/test/connections.test.ts packages/contracts/test/google-drive.test.ts apps/web/src/lib/google-drive-connection.test.ts apps/web/src/components/capabilities/google-drive-connector-card.test.tsx` — 67 pass, 0 fail, 441 assertions
- `bun run typecheck` — all 24 projects clean
- `bunx oxlint --deny-warnings apps/api/src/integrations/google-drive.ts apps/api/test/google-drive-oauth-isolation.test.ts`
- `bunx oxfmt --check apps/api/src/integrations/google-drive.ts apps/api/test/google-drive-oauth-isolation.test.ts`
- `bun scripts/check-workspace-billing-static.ts`
- `git diff --check origin/main..HEAD`

No live Google/cloud APIs, deployment, production credentials, merge, or release were used.

Linear: OPE-162 (supports OPE-124; asserts merged OPE-160 behavior)

PR #1250, opened 2026-08-09
test(release): verify terminal output across renderers

jorgensandhaug:
## Summary

- replace the private `.xterm-rows` terminal assertion with a renderer-neutral PTY websocket receipt
- encode the command marker so echoed input cannot satisfy the execution proof
- retain bounded split-frame matching across WebGL, canvas, and DOM renderers
- cover binary/split frames and cleanup after a failed keyboard action

## Production evidence

Production acceptance run 31313300626 reached a live terminal but timed out reading `.xterm-rows`. A direct production probe proved `data-og-term-renderer="webgl"`, no DOM row text, and exact decoded PTY output received by the browser. PR #1248 separately fixes the pre-connect input-readiness race; this follow-up removes the renderer-specific false negative.

## Validation

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun run check:public-hygiene`
- `bun test packages/react/test/sandbox-terminal.test.ts scripts/workbench-live-acceptance.test.ts`

PR #1585, opened 2026-08-18
test: cross-organization isolation and revocation evidence (OPE-204 slice 9)

benyhh:
Closes the phase E "validate" step of `docs/organization-tenancy.md` for OPE-204 with a durable, **executed** evidence suite. Test-only: no migration, no shipped package behaviour change, no changeset (matching how other test-only PRs in this repo land, e.g. #1260, #1250).

## What is added

One file: `packages/db/test/organization-isolation-evidence.test.ts` (12 tests), plus a paragraph in `docs/organization-tenancy.md` § E pointing at it.

It runs against a real PostgreSQL database from `acquireSharedTestDatabase()` (migrated + role-provisioned template — deliberately **not** `acquireBlankTestDatabase()`, which has neither the migration chain nor the `opengeni_app` grants), and issues every probe over a raw `shared.appUrl` connection as the genuine non-superuser, `NOBYPASSRLS` `opengeni_app` login with transaction-local GUCs set explicitly — so a test can also express the *absence* of context, which `withWorkspaceRls` / `setRlsContext` cannot.

## What it proves

**Meta-guard first.** `opengeni_app` is asserted `rolsuper=false, rolbypassrls=false, rolinherit=false`, `pg_has_role(... 'USAGE')=false` against each authority table's owner, and `row_security_active('sessions')=true` on the probing connection. Without this, the whole file could pass while measuring owner-privileged behaviour.

**Cross-organization isolation.** Two organizations, each a real `managed_accounts` row, plus a sibling workspace inside org A. One seeded row in each of 13 families: session, session event, Variable Set, Rig, Connected Machine, enrollment, connection, file, Document base, Document, knowledge memory, scheduled task, API key. Denied for: org B's honest context; org B forging org A's workspace id; org A's account paired with org B's workspace; org A's sibling workspace. Insert carrying org A's identity under org B → `42501`; update/delete of org A rows under org B → 0 rows, then re-read as superuser confirms the rows are intact and still visible to their owner.

**Positive controls everywhere.** The exact reachability map under the owning context is pinned (`visible` / `hidden` / `denied` per family), so no negative assertion can pass vacuously. `workspace_variable_sets` is pinned as `denied` — the runtime role holds *zero* privileges on it (capability-only, migration 0254) — and that is separately proved the hard way with executed `SELECT/INSERT/UPDATE/DELETE` → `42501` under the owning organization's own context.

**Fail-closed.** No GUCs at all, blank GUCs, and workspace-without-account all return nothing. Account-without-workspace correctly widens to the one deliberately account-scoped family (`api_keys`, via `optional_workspace_rls_visible`) and to nothing else — and org B's account-only context still sees zero, so that widening never crosses the organization.

**Forged lifecycle GUC.** The four FORCE-RLS authority tables' policy is a bare `current_setting('opengeni.organization_tenancy_lifecycle')` comparison, and any caller can set a GUC. The suite sets each accepted lifecycle value from the runtime role and still gets `42501` on all four tables × read + delete — because privilege is checked before RLS. Generalised across the whole catalog: every table carrying a permissive policy gated on that GUC *and nothing the caller cannot control* (no `CURRENT_USER` owner check) must hold zero runtime privileges. 11 such tables are found; all zero.

**Catalog completeness.** Every `public` table with an `account_id` column on which the runtime role holds any DML must be `ENABLE` + `FORCE` RLS with at least one policy. 254 qualify. The exception set is pinned to exactly `{auth_identities, workspace_memberships, workspaces}` — the tenancy directory the access resolver reads before an RLS context exists, whose boundary is enforced in `@opengeni/core`. A separate test states that honestly rather than over-claiming: org A's `workspaces` row *is* readable under org B's context at the database layer. Both directions of drift (a new unprotected table, or one of these gaining RLS) fail the file.

**Immediate revocation.** The same human holds an active membership in *both* organizations and owns one `user_private` session in each. Org A's private session is invisible from org B, invisible through a forged workspace id, and invisible to another subject inside org A. Setting the org A membership to `revoked` makes it invisible on the very next transaction on the same pooled connection — no reconnect, no cache flush, no epoch bump — while org B's session stays visible and the org A session row is still physically present (retention is not authority). Reactivating restores access, which proves the denial was caused by the membership status and nothing incidental.

**FORCE RLS binds a non-superuser SECURITY DEFINER owner.** OpenGeni's capability seams are `SECURITY DEFINER` routines owned by the migration owner, which in production is not a superuser. The suite constructs that exact situation — a `NOSUPERUSER NOBYPASSRLS` role owning a FORCE-RLS table and a definer function over it — and shows the function returns 1 under the owning organization, 0 under the other, and 0 with no context, while the runtime role still gets `42501` on the table directly. The role and its objects are dropped in `afterAll`.

## Non-vacuity check

I sabotaged each mechanism against a live database and confirmed the assertions flip:

| sabotage | observed |
|---|---|
| baseline | own org sees the rig `1`, other org `0` |
| `ALTER TABLE rigs DISABLE ROW LEVEL SECURITY` | other org sees `1` |
| re-`ENABLE`/`FORCE` | back to `0` |
| add an `account_id` table with app grants and no RLS | unprotected set grows to include it → catalog test would fail |
| `GRANT SELECT ON organization_memberships TO opengeni_app` | GUC-gated table reports `select=true` → forged-GUC test would fail |

## Pre-existing coverage (deliberately not duplicated)

- `migration-0218-…` and `migration-0219-…` already execute the 4-tables × 4-verbs `42501` probe on the authority tables, and 0218 covers cross-organization *referential* rejection (`23503`).
- `migration-0263-…` (~2900 lines) covers the membership lifecycle: CAS transitions, "revokes access immediately without deleting retained data", per-organization retention operator capability fencing, invitation isolation.
- `migration-0278-…` / `workspace-membership-removal.test.ts` cover fenced workspace-membership removal.
- `migration-0225-…` already covers "a suspended owner cannot see their own private session" *within* one organization — so this suite deliberately covers the cross-organization and revoke/reactivate-causality angles instead.
- `rls-dedicated-schema-isolation.test.ts` covers the generic cross-tenant probe, the declared FORCE-RLS posture, and transaction-local GUCs.
- `documents`, `connections`, `knowledge`, Rigs/Machines and Variable Sets each have scope-isolation tests in their own migration files.

## Gaps deliberately left

- **HTTP/API surface.** This suite is database-layer. Route-level cross-tenant denial already has coverage in `apps/api/test/*` and is not re-proved here.
- **`workspaces` / `workspace_memberships` / `auth_identities` have no RLS.** Documented and pinned rather than changed — adding RLS to the tenancy directory is a behaviour change well outside a test-only slice.
- **`api_keys` has a `key_hash = current_api_key_hash()` branch with no account check** (the authentication bootstrap path). Presenting the correct hash is presenting the secret itself, so it is not a cross-organization escalation; the suite asserts the account/workspace path only and does not attempt to change it.
- **Production owner is not exercised as the migration owner.** In the shared test harness every table is owned by superuser `postgres`, so the `CURRENT_USER = 'postgres'` capability policies cannot be exercised under a realistic non-superuser owner. The suite proves the underlying FORCE-RLS/definer property with a constructed non-superuser role instead. Worth noting separately: some capability policies hardcode the literal `'postgres'` while others resolve `pg_get_userbyid(relowner)` — that inconsistency is out of scope here but may matter for deployments whose migration owner is not `postgres`.
- **`inventory_organization_tenancy` (migration 0285)** is not on this base and is not depended on.


benyhh:
Moving-main compatibility check (no source mutation): verified this frozen head against current `main` (`a03b86ff`, which added migration 0285 after this branch was cut).

- `git merge-tree --write-tree origin/main HEAD` → clean, no conflicts.
- Applied this branch's test file onto a temporary branch at `origin/main` and ran it against a real PostgreSQL template built from main's full migration chain including 0285: **12 pass, 0 fail, 73 expect() calls**.

The dynamic catalog assertions (FORCE-RLS completeness, caller-settable-GUC privilege sweep) therefore hold on current main's schema. No rebase or merge performed.

PR #1601, opened 2026-08-18
style: apply oxfmt to seven files that landed unformatted on main

benyhh:
`bun run format:check` fails on current `main` for seven files, which blocks the `Source and type contracts` gate on every branch cut from it - including main's own CI, which is currently red for this reason.

Verified independently: checking each file out from `origin/main` and running `oxfmt --check` on it in isolation reproduces the failure. They landed via the open-suffix-resume work (#1596); an eighth file from the same batch was fixed separately in the meantime, so this branch was rebuilt from current main.

Pure `bun run format` output. No logic, no behavior, no test changes.

## Note on main's other failure

`main` is also failing `structured human-input HTTP surface (real PostgreSQL) > reads pending requests, rejects invalid responses, and signals one accepted settlement` with `ApprovalRunStateLimitExceededError: Approval run state is not a valid serialized JSON object`. That is a separate, genuine regression from the same PR that introduced these unformatted files, and it is **not** addressed here - this PR is formatting only. Flagging it so it is not mistaken for fallout from this change or from the capabilities work.

https://claude.ai/code/session_01BCpgr6298bGDuRDH3SEhcA

### DIFF
--- a/apps/worker/src/activities/agent-turn.ts
+++ b/apps/worker/src/activities/agent-turn.ts
@@ -483,10 +483,7 @@ import { executeGatewayImageGeneration } from "./gateway-image-generation";
 import { executeCodexImageGeneration } from "./codex-image-generation";
 import { imageProviderBindingHash } from "./image-generation-operation";
 import { resolveImageGenerationReferences } from "./image-generation-references";
-import {
-  interruptionCallIdsFromPause,
-  settleOpenSuffixResumeIfNeeded,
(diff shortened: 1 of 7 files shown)


benyhh:
Superseded: main already contains equivalent formatting (the merge tree of this branch against current main is byte-identical to main, so this PR is a no-op). The unformatted files from #1596 were fixed independently. Closing rather than merging an empty change.

PR #1553, opened 2026-08-17
feat(lens): add provider-neutral pull request reviews

davletd:
## Summary

- Add the built-in OpenGeni Lens Capability Pack with a reviewed inline `pr-review` Skill and a setup surface for dedicated provider credentials.
- Define the Skill as an applicability-driven review of every changed file and affected cross-layer path, with explicit security, application-correctness, and infrastructure/deployment rubrics.
- Treat PR text, code, filenames, comments, generated files, and repository instructions as untrusted input; allow static inspection only while provider write credentials are available.
- Support authenticated GitHub App, GitLab token, and Azure DevOps PAT webhooks through one normalized pull-request event contract.
- Dispatch idempotent ordinary OpenGeni sessions at the exact immutable PR head, on managed compute, with host-brokered provider credentials and provider CLI support.
- Persist encrypted registrations, repository bindings, and delivery journals behind FORCE RLS; recheck exact head and live Lens session/turn/attempt authority before every credential mint or provider write.
- Keep Lens CRUD behind the opt-in `@opengeni/sdk/lens` client and lazy setup panel so review-only schemas and methods do not enter ordinary session bundles.
- Pin custom GitLab DNS/transport against SSRF, bound webhook/provider responses, and document setup, permissions, limits, and architecture.

## Review coverage

- **Security:** authentication, authorization, tenant isolation, confused-deputy paths, injections, XSS/CSRF/SSRF/traversal, secrets and privacy, webhook/crypto/replay/TOCTOU, supply chain, and denial-of-service/resource bounds.
- **Application:** business and state invariants, edge/error/cleanup paths, transactions/concurrency/idempotency/retries, compatibility and migrations, cross-layer behavior, performance, observability, and missing regression tests.
- **Infrastructure:** IAM/network/storage boundaries, encryption and secret lifecycle, rollout/rollback/version skew, container and Kubernetes posture, CI/CD trust boundaries, scaling/quota/cost, backups, and monitoring.
- Findings must identify a concrete changed-code path, trigger, impact, severity, path/line, and actionable fix. Style-only, generic, speculative, unrelated, and duplicate findings are excluded.

## Testing

- [x] `bun run typecheck` (31 projects)
- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun scripts/check-docs-refs.ts`
- [x] `bun run check:migration-ordinals`
- [x] 292 affected API/core/worker/runtime/GitHub/database/schema tests
[...]

PR #1435, opened 2026-08-14
fix: reconcile orphaned provider-loss admissions safely

jorgensandhaug:
## Summary

- Rebuild the orphaned provider-loss repair from protected main `1cb4daf6e1ab13c12fa0d320d74fcea71f0f787f` as a clean successor to #1413.
- Add migration 0240's exact provider-loss claim/receipt protocol and ordinary reaper reconciliation for only an independently proven Modal `not_found` tuple.
- Preserve fail-closed fencing across lease, group, epoch, provider instance, admission, attempt, turn, retained-process, PTY, holder, and archive authority; never replay or reissue the provider operation.
- Keep the four provider-loss trigger-only guard routines non-executable by the runtime role while requiring them to exist exactly once in runtime posture evaluation.
- Preserve restrictive session-visibility isolation on both provider-loss tables, keep Drizzle metadata identical to all physical migration constraints, and publish the required package changeset.

## Testing

- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run typecheck` (31 projects)
- [x] `changeset status` resolves `@opengeni/db` and `@opengeni/worker-bundle` patch releases.
- [x] Runtime posture/schema/release contracts: 29 passed, 0 failed before final parity repair.
- [x] Focused migration/worker suite: 57 passed, 0 failed before final parity repair.
- [x] Focused migration 0225/0240 policy suite after V2 repair: 11 passed, 0 failed.
- [x] Governed release-schema contract after V3 hash refresh: 6 passed, 0 failed, 165 assertions.
- [x] V4 schema/migration parity + posture/release suite: 32 passed, 0 failed, 264 assertions.
- [x] PostgreSQL 17.11 + pgvector dedicated-schema/non-owner replay with `OPENGENI_REQUIRE_REAL_DB=1`: 10 passed, 0 failed, 138 assertions.
- [x] Direct PostgreSQL catalog receipt: both provider-loss policies are restrictive and call the target-schema `session_reference_visible` routine in both `USING` and `WITH CHECK`.
- [x] PostgreSQL 17.11 worker provider-loss/reaper suite with explicit admin/app URLs and `OPENGENI_REQUIRE_REAL_DB=1`: 42 passed, 0 failed, 404 assertions.
- [x] `git diff --check`

## Delivery integrity

- [x] Candidate/version labels represent substantive source changes, not base-only refreshes.
- [x] Protected main is an ancestor of the frozen head; no base-only refresh was required.
[...]

PR #1434, opened 2026-08-14
Fix Google Drive read authority for OPE-124

jorgensandhaug:
## Summary

- restrict the Google Drive API Integration preset to `drive.readonly` and an explicit reviewed read-operation allowlist; Discovery's POST `drive.files.download` is excluded so every retained tool compiles as `safety=read` / `approvalMode=never`
- require literal `account:admin` for every workspace-owned Google Drive credential offer and mutation (start, reconnect, callback persistence, and disconnect), while leaving personal Drive instances and other providers unchanged
- revalidate the current workspace grant after provider identity, then inside the credential-write transaction lock and recheck the exact account-admin authority plus `workspace_memberships` row and its `connections:write` / `workspace:admin` permission
- narrow legacy personal and workspace Drive reconnects to the exact current reviewed Google Drive scope set, disable Google incremental-scope inclusion only for Drive, and fail closed unless Google confirms the narrowed set
- require a newly returned refresh token when narrowing a legacy broad Drive grant, so the callback cannot relabel and reuse a historical full-Drive refresh token; already-narrow Drive, Gmail, Microsoft, and other intentional reconnect fallback behavior remains unchanged
- preserve current main's migration 0243 Google Drive object-ACL authority and all-protectors retrieval checks alongside OPE-124 migration 0242 account-admin authority
- keep the verification/release packet truthful: Drive ACL/citation reauthorization is shipped by migration 0243; only Workspace Events subscription/Pub/Sub provisioning and live-provider acceptance remain release work

## Testing

- [x] exact source: `bun run typecheck` — all 31 projects clean
- [x] exact source: lint 0 warnings/errors; format and public hygiene green
- [x] exact source: 123 release-schema, database-posture, and worker lifecycle tests passed
- [x] exact source: 84 focused provider OAuth, connection-route, Drive, and capability tests passed
- [x] exact source: DB-backed regressions cover personal/workspace legacy full-Drive narrowing, exact persisted grants, broader/missing scope rejection, missing replacement refresh-token rejection with unchanged credential/version/grants, already-narrow Drive refresh-token preservation, Gmail incremental consent, and unchanged Microsoft scope-union behavior
- [x] PR #1405 exact combined integration: conflict-free, 394 tests passed, plus all 31 typechecks
- [x] exact-head Source Admission run `31769977321` succeeded for `6be22acaae5fba4a56b0cdda36d2e66cca50e96b`
[...]

PR #1631, opened 2026-08-19
fix(db): own sessions created in a human's own personal workspace

benyhh:
## The bug

`guard_session_authority_write` (migration `0225_session_visibility_fork_activation.sql`) derives a new session's owner from the creator's active `organization_memberships` row, but gates that resolution on:

```sql
AND EXISTS (
  SELECT 1 FROM workspace_memberships workspace_membership
  WHERE workspace_membership.account_id = NEW.account_id
    AND workspace_membership.workspace_id = NEW.workspace_id
    AND workspace_membership.subject_id = membership.subject_id
)
```

A managed human's personal workspace **deliberately has no `workspace_memberships` row** — `0219_organization_tenancy_managed_human_provisioning.sql` raises `42501 'managed-human personal workspace already has runtime access'` if one exists. The `EXISTS` can therefore never be satisfied there, the `SELECT … INTO` finds nothing, and the session lands with `owner_organization_membership_id = NULL`.

Consequence: in the one workspace where the human is by construction the only possible member, they cannot own their own sessions — so those sessions can never become `user_private` and can never use that human's personal connections or Documents.

`0258_three_scope_document_knowledge_authority.sql` already handles this correctly for Documents (it requires the workspace membership only when the target workspace is *not* the member's own `personal_workspace_id`). The session trigger was never updated to match. `0297_session_ownership_classification_and_backfill.sql` names this population explicitly ("A MANAGED HUMAN IN THEIR OWN PERSONAL WORKSPACE") and repairs it historically — but nothing stopped the mint.

### Reproduced before the fix

Driving a real `createSession` for a managed human provisioned through `ensureManagedAccessForUser`, one session into their legacy Better Auth workspace and one into their own personal workspace:

```
workspace_memberships rows: [ { workspace_id: <legacy>, subject_id: user:pw-owner-… } ]   # exactly one
legacy workspace: 8b19e7ad-…   personal workspace: 2e6b8063-…

SESSION OWNERS:
  { workspace_id: 8b19e7ad-… (legacy),   owner_organization_membership_id: "05b85d14-…" }
  { workspace_id: 2e6b8063-… (personal), owner_organization_membership_id: null }
```

After the fix, the same run:

```
  { workspace_id: bb7bd814-… (legacy),   owner_organization_membership_id: "d307f379-…" }
  { workspace_id: 816a1459-… (personal), owner_organization_membership_id: "d307f379-…" }
```

## The fix — migration 0302 (rolling)

Replaces one function body and nothing else. No column, index, constraint, policy, grant, or trigger definition changes; old images keep running unchanged.

1. **Stated authority, not inference.** `membership.personal_workspace_id = NEW.workspace_id` becomes an alternative to the `workspace_memberships` EXISTS, mirroring 0258. Phase D forbids inferring user authority from `created_by`, connection attribution, a default workspace, a resource name, or current access — none of which applies: the membership row **itself** names the workspace in its own column, written by 0219's lifecycle capability. The pointer is 1:1 (`organization_memberships_personal_workspace_idx`), and there is at most one membership per (account, subject).
2. **The ordinary shared-workspace path is untouched.** A human with an active membership but no `workspace_memberships` row in an ordinary workspace still resolves to no owner there — covered by a test.
3. **`user:%`-restricted.** 0219/0263 gate the entire managed-human lifecycle to `user:%`, and 0297's classifier calls the non-`user:` lane `external_lane_owns_row` (permanently unrepairable). The new disjunct is restricted to match, so the write path never attributes a subject the classifier refuses.
4. **The silent miss is now loud.** The bare `SELECT … INTO` gained `IF NOT FOUND` classification. Genuinely-ownerless cases stay NULL exactly as before: `created_by_kind <> 'subject'` never enters the branch; `api_key:`/`configured:` subjects can never acquire an anchor; a `user:` subject with no active membership; a `user:` subject with no stated authority over an ordinary workspace; and a suspended/revoked member (excluded by the same `status = 'active'` predicate from both the resolver and the classifier). What now raises `55000` is a `subject`-kind creator **with** an active membership inserting into a workspace that is an active membership's `personal_workspace_id` and still resolving to no owner — a workspace with exactly one possible owner by construction, so an ownerless session there is a resolver defect. 0219's access model makes that unreachable in production.

## The misleading fixture

`packages/db/test/migration-0225-session-visibility-fork-activation.test.ts` hand-inserted a `workspace_memberships` row into a fake "personal workspace" and used the result as a cross-workspace fork destination. Real provisioning forbids that row, which is exactly why the defect was invisible to this suite.

Fixed in place: the second member's personal workspace now keeps the shape real provisioning produces (membership pointer, no `workspace_memberships` row, control row). Forks target a new **ordinary** second workspace with a real membership row — which is what `fork_session_content` actually requires (it checks `workspace_memberships` for the actor in the destination, `0225:1120-1125`). All 0225 assertions are unchanged and still pass.

## Tests

New `packages/db/test/migration-0302-personal-workspace-session-ownership.test.ts` — every fixture goes through the real `ensureManagedAccessForUserWithOrganizationMemberships` seam:

- real provisioning leaves the personal workspace membershipless (pins the premise);
- a session in the human's own personal workspace is owned;
- a session in an ordinary workspace they are a member of is still owned;
- a session in an ordinary workspace they cannot access is still **not** owned;
- service-created and `api_key:`-subject sessions stay ownerless;
- a suspended member's personal-workspace session stays ownerless **without** raising;
- an intruder in another member's personal workspace raises `55000` with the exact message (0214's commit gate also uses `55000`, so the message is asserted too) and writes no row;
- a child session still inherits its parent's owner pair verbatim.

Plus a block under `acquireOwnerMigratedTestDatabase` — a `NOSUPERUSER NOBYPASSRLS` migration owner, where `FORCE ROW LEVEL SECURITY` genuinely engages for the SECURITY DEFINER trigger's `organization_memberships` read. The shared harness migrates as the container superuser and cannot observe that boundary at all.

`migration-0297-session-ownership.test.ts` was updated rather than weakened: its `measured population` test now pins the new live-trigger behaviour (personal-workspace sessions attributed; the cross-personal shape refused), and the backfill fixtures seed pre-0302 rows explicitly through a new `insertHistoricalSession` helper — because that is exactly the population the backfill exists for.

## Verification

`typecheck`, `lint`, `format:check`, `check:public-hygiene` (post-commit), `check:docs-refs`, `check:migration-ordinals`, `scripts/release-schema-contract.test.ts` (re-pinned from the test's own reported values), and the full `packages/db` suite — **1513 pass / 0 fail across 265 files** against real Postgres.

## Historical backfill — reported, not implemented

**No new backfill is needed, and none is included here.** Ownerless rows from this bug cannot be repaired by app DML (`0225:616-620` raises `42501 'legacy workspace session has no transferable owner authority'`), but migration **0297 already ships the repair** for precisely this shape — its "Personal-workspace convergence" candidate (`sessions.workspace_id` equals one active membership's `personal_workspace_id` **and** `created_by_subject_id` is that membership's subject, with `created_by` used only as a veto), plus its parent-inheritance closure. It runs behind a transaction-scoped capability policy that works under a non-superuser migration principal, is dry-run by default, bounded, resumable, and idempotent:

```
bun run db:backfill-session-ownership --organization-id <uuid> --classify
bun run db:backfill-session-ownership --organization-id <uuid> --apply --limit 100
```

So the operator step after this merges is simply to re-run that existing command; it is already correct for the now-resolvable case and needs no change. Rows where the workspace anchor and the creator disagree remain `ambiguous_candidate_authority` and are deliberately left alone.

Doing it inside this migration would be wrong regardless: it would fold a bounded, receipt-recording, operator-driven data pass into a schema cutover.

## Known follow-up (not in this PR)

`0298_organization_tenancy_parity.sql`'s `sessionsAttributableButUnattributed` lane still encodes the *old* fence predicate (workspace-membership only), so it now under-reports pre-0302 personal-workspace rows. The prose in `packages/db/src/index.ts` was updated to say so honestly rather than silently drift. Widening that lane means a `CREATE OR REPLACE` of the ~480-line `check_organization_tenancy_parity` seam and belongs in its own change.

Also noted: `fork_session_content` requires a `workspace_memberships` row in the destination, so forking a session **into your own personal workspace** is impossible today. Separate gap, separate change.

## Unrelated formatting

`apps/worker/test/agent-turn.test.ts` carries a pre-existing `oxfmt` drift on `main`; `format:check` fails without it. Two mechanical line wraps, no behaviour.


benyhh (commented):
## Independent adversarial review — APPROVE

I rebuilt the tree from scratch (`/home/bendik/rev1631`, fresh clones of `origin/main` and `refs/pull/1631/head`, `bun install`, real Postgres), reproduced the bug myself before reading the fix, then tried to break the fix five different ways. I would stake production on this.

---

### 1. The bug is real — reproduced independently on `origin/main` (81d2da0b)

I wrote my own repro rather than trusting the PR body: provision a managed human through `ensureManagedAccessForUserWithOrganizationMemberships`, then drive real `createSession` into the legacy workspace and into the personal workspace.

```
[repro] legacy workspace   = c06a5430-…
[repro] personal workspace = d19b8176-…
[repro] workspace_memberships rows = [ "c06a5430-…" ]        # legacy only, exactly one
[repro] session LEGACY    owner = 8c51f215-…                 # attributed
[repro] session PERSONAL  owner = null                       # NOT attributed
```

Same repro on this branch: `session PERSONAL owner = 8c51f215-…`. The premise holds too — `0219:169-176` raises `42501` on **any** `workspace_memberships` row in the personal workspace, and that check runs on every convergence, so the `EXISTS` gate in 0225 is unsatisfiable there by construction.

The repaired path is live, not theoretical: `managedPersonalWorkspacePermissions` (`packages/db/src/index.ts:1160`) includes `sessions:create`.

### 2. Privilege widening — I could not construct one

- **Uniqueness / null-safety.** `organization_memberships_personal_workspace_idx` is `UNIQUE (account_id, personal_workspace_id) WHERE personal_workspace_id IS NOT NULL`, and `organization_memberships_account_subject_idx` is `UNIQUE (account_id, subject_id)`. Exactly one candidate; no disambiguation needed. A `NULL` pointer makes `membership.personal_workspace_id = NEW.workspace_id` evaluate `NULL` → not true, in every branch, and the `active_personal_workspace_check` CHECK forbids `NULL` on an active row anyway.
- **Cross-organization collision.** `organization_memberships_personal_workspace_account_fk` is the **composite** `(personal_workspace_id, account_id) → workspaces(id, account_id)`, and every new predicate is `account_id = NEW.account_id`-scoped. A pointer cannot name a foreign-org workspace, and a foreign-org membership cannot match.
- **Stale / suspended / revoked.** `status = 'active'` on the resolver; `status <> 'revoked' → revoked_at IS NULL` is a CHECK, so 0297's extra `revoked_at IS NULL` is redundant and the write path and the backfill candidate are now *exactly* aligned.
- **"A human whose `personal_workspace_id` points at a workspace they should not control."** To reach the fence at all you must be able to INSERT a session into that workspace. I traced the whole grant chain: `accessGrantAuthorization` → `context.workspaceGrants` ∪ `getWorkspaceGrant`, and `getWorkspaceGrant` is a bare `workspace_memberships` join (`packages/db/src/index.ts:2079`). `ensureManagedAccessForUserWithOrganizationMemberships` pushes workspace grants **only** from persisted `workspace_memberships` rows plus *the caller's own* `personal_workspace_id`. Organization role produces `accountGrants` only. So no organization owner/admin, API key, other member, or service acquires ambient access to somebody else's personal workspace. The widening is exactly the intended one.

### 3. `user:%` — correct, and stricter than the precedent

`LIKE 'user:%'` is the established convention (0219, 0263, 0285, 0290, 0292, 0297, 0298), and 0297's classifier gates the *same* column with the *same* test, so the write path can no longer attribute anything the classifier refuses. I dumped `pg_proc.prosrc` from a migrated database to confirm `format()`'s `%%` renders as a single `%`:

```
NEW.created_by_subject_id LIKE 'user:%'
AND membership.personal_workspace_id = NEW.workspace_id
```

Crafting a passing subject requires already owning an active `organization_memberships` row for that exact string, which only the `user:%`-gated 0219/0263 lifecycle can mint. Note the disjunct is *narrower* than 0258's document equivalent, which carries no prefix gate at all.

### 4. The `55000` raise — walked every caller, found no legitimate path

- `created_by_kind <> 'subject'` (service/scheduler/`unattributed-legacy`) never enters the branch.
- `api_key:` / `configured:` / `unattributed-legacy` subjects fail `LIKE 'user:%'` in the classifier → NULL, no raise.
- **Fork:** `fork_session_content` supplies `owner_organization_membership_id` explicitly (`destination_owner_id`), so the whole derivation block is skipped; it also requires a destination `workspace_memberships` row, which a personal workspace can never have.
- **Children:** `parent_session_id IS NOT NULL` takes branch 1, which has no raise — so a child of a pre-0302 NULL-owner parent in a personal workspace still creates cleanly. Important for rolling safety, and covered by a test.
- **Agent-spawned:** `frozenInitiatorForCommandActor` inherits the *calling turn's* initiator, and children are same-workspace, so the creator is always the workspace's own human.
- **Deleted workspace:** the pointer FK is `ON DELETE RESTRICT`; the anchor cannot dangle.
- **Cross-org:** account-scoped, see §2.

Two residual notes, neither blocking:

- A first-party **delegated** token signed with `subjectId: user:Y` + `workspaceId: <X's personal workspace>` would now hard-error instead of minting an ownerless `workspace_shared` session. That combination is already an impersonation the deployment secret authorizes, and failing loudly is the better outcome — worth knowing it exists.
- The resolver `SELECT` and the classifier `SELECT`s are separate statements under READ COMMITTED. A membership **reactivation** committing between them could produce a spurious `55000`. It fails closed and is retryable; folding the classification into the resolver's query would remove it.

Nothing in the app maps `55000` in a way that would misfire — the three handlers (`scoped-knowledge.ts:377`, `governed-learning-activation.ts:245`, `organization-membership-lifecycle.ts:7`) are all on unrelated paths.

### 5. Phase D — genuinely stated authority

The predicate reads `personal_workspace_id` **off the authority row itself**, written by 0219's SECURITY DEFINER lifecycle, constrained 1:1 and same-org. That is not `created_by`, not connection attribution, not a resource name, not current access — and specifically **not a default workspace**: I confirmed empirically that `defaultWorkspaceId` remains the legacy Better Auth workspace and is a *different* id from the personal workspace. 0258 already blesses this exact shape for Documents, and 0297's already-shipped backfill uses this exact conjunction as its "reviewed, deterministic evidence for user ownership". Honest verdict: stated authority, not inference wearing a hat.

### 6. The rewritten 0225 fixture — not reshaped to pass

Diffed old vs new explicitly. **9 tests before, 9 after. 79 literal `expect(` before, 79 after.** Not one assertion dropped, weakened, or turned into a `toMatchObject` escape hatch — the only change is the *value* of `destinationWorkspaceId` / `workspaceId`.

Cross-workspace fork coverage is genuinely preserved, and is now *more* faithful: forks target a new ordinary second workspace with a real `workspace_memberships` row for the actor, which is what `fork_session_content` actually requires (`0225:1119-1126`). `otherPersonalWorkspaceId` keeps its membership pointer and control row but loses the hand-inserted `workspace_memberships` row — i.e. it now matches what 0219 actually produces, which is precisely why the defect was invisible to this suite. Runs green (9 pass, 81 expect calls).

### 7. The `NOSUPERUSER NOBYPASSRLS` block earns its keep — proved by sabotage

I removed the `set_config('opengeni.organization_tenancy_lifecycle', 'session_visibility_activation', …)` marker, so the SECURITY DEFINER's `organization_memberships` read has to survive FORCE RLS on its own:

```
(fail) migration 0302 under a NOSUPERUSER NOBYPASSRLS migration owner > resolves the personal-workspace owner …
 8 pass, 1 fail
```

Every shared-harness test stayed green; **only** the owner-migrated block went red. That is exactly the blindness the block exists for, and it is not decorative.

Full sabotage matrix (each applied to the migration, tests re-run against real Postgres):

| Sabotage | Result |
|---|---|
| new disjunct → `false` | 3 fail (incl. owner-migrated) ✅ |
| drop the RLS lifecycle marker | 1 fail — owner-migrated only ✅ |
| drop the `user:%` gates | 0297 `the write path never attributes a subject the classifier calls unrepairable` fails ✅ |
| widen to `personal_workspace_id IS NOT NULL` | 4 fail across 0302 + 0297 ✅ |
| resolver accepts `suspended` | suspended-member test fails ✅ |
| **drop `personal_owner.status = 'active'`** | **9 pass — nothing fails** ❌ |
| append a byte to the migration | release-schema hash ladder goes red on the exact pinned hash ✅ |

**The one gap:** the `status = 'active'` predicate on the raise classifier's `personal_owner` anchor is untested. The "suspended member" test never reaches it — there the *creator* is the suspended member, so `creator_has_active_membership` short-circuits first. The shipped code is correct; the guard is just unpinned. A test where an **active** member creates a session in a **suspended** member's personal workspace (expect NULL owner, no raise) would close it. Worth adding, not worth blocking.

### 8. Ordinal / ladder

`check:migration-ordinals` green: `311 migrations, base origin/main highest 0301`. I also enumerated every open PR's `packages/db/drizzle/**` — the only other migrations in flight are 0280 (#1553), 0240 (#1435), 0242 (#1434). No 0302 collision. The ladder pin is a real recomputation, not a copied string: appending one comment line to the migration turns `scripts/release-schema-contract.test.ts` red on `e704c389…`.

### 9. Claim 1 (0298 lane) — I checked, and deferring is right

The divergence is real, and I measured it. With a durable pre-0302 personal-workspace ownerless session present:

```
[parity] lane count = 0   drained = true
[parity] cutoverReady = true
[parity] undrained lanes = []   failed gates = []
```

So `cutoverReady` can read green while that population is still unrepaired. **But I ran the identical probe on plain `origin/main` and got byte-identical output** — the false green is pre-existing, not introduced here. This PR only makes the lane's own doc-string stale, which the prose change now states honestly, and the operator retains a correct signal in `db:backfill-session-ownership --classify` (`repairablePersonalWorkspace`). Widening the lane means `CREATE OR REPLACE` over the ~480-line parity seam — correctly its own change. Please do open the follow-up issue; a `kind: "drainable"` lane feeding `cutoverReady` should not stay knowingly narrow.

### 10. Verification I ran

On the PR head: `typecheck` ✅ · `lint` ✅ · `format:check` ✅ · `check:public-hygiene` ✅ · `check:docs-refs` ✅ · `check:migration-ordinals` ✅ · `release-schema-contract.test.ts` ✅ · full `packages/db` — **1513 pass / 0 fail across 265 files** (exactly the PR's claim) · `apps/api` personal-workspace + organization-membership route tests ✅.

(To be scrupulous: `check:migration-rls-backfills` is *not* in that green list. On the PR head alone it is red — see below.)

**The red CI on this PR is stale, not a defect.** All five failing jobs ("Source and type contracts" + unit shards 1–4) are the same `check:migration-rls-backfills` false positive that also flags **0301, already on `main`** — and `origin/main`'s 81d2da0b explicitly fixes it ("ignore trigger-function DDL in the FORCE-RLS backfill scan (0301 false positive)"). That commit landed at 13:32; the CI run is from 11:34. I merged this branch with current `main` locally and re-ran:

- `check:migration-rls-backfills` → *No unreviewed FORCE-RLS migration backfills* ✅
- `scripts/check-migration-rls-backfills.test.ts` → 19 pass / 0 fail ✅
- `lint`, `check:migration-ordinals`, `check:docs-refs`, `check:public-hygiene` ✅
- `format:check` → fails **only** on `packages/xai-subscription/src/fetch.ts`, the known separate defect being fixed in #1632

Merge is conflict-free (`git merge-tree` clean). Note 81d2da0b made the **same** two `agent-turn.test.ts` wraps this PR does, so that "unrelated formatting" hunk merges to a no-op — harmless, and it was genuine drift at this branch's base (I confirmed `format:check` fails on 89bc6b73 for exactly that file). **Just re-run CI; do not touch the source branch** — a protected-branch advance is not a source revision.

Per-repo policy I also checked: `-- deployment-mode: rolling` is correct (one `CREATE OR REPLACE FUNCTION` + a `COMMENT`; no column, index, constraint, policy, grant, or trigger change), and the schema-contract `fileCount` was incremented accordingly. The semantic diff of the function body against 0225 is exactly three things — one `DECLARE`, the disjunct, the `IF NOT FOUND` classifier — with everything else byte-faithful.

### 11. Things I could not verify

- Behaviour against real production tenancy data and volume; everything here is synthetic.
- My full-repo `test:unit` run has two failure clusters, **both of which I reproduced on pristine `origin/main`**, so neither is attributable here:
  - `scripts/prepare-development-artifact-runtime.test.ts` — 4 tests, missing native toolchain on this box.
  - 17 tests across `apps/api/test/sandbox-desktop-stream.test.ts`, `sandbox-consent-gate.test.ts`, and `sandbox-shared-and-viewer.test.ts` — all `RangeError: Maximum call stack size exceeded`. It is a cross-file `mock.module` bleed with `apps/api/test/selfhosted-stream-mint.test.ts`: both sides wrap `getSessionAuthorityEpoch` with a `realDb.getSessionAuthorityEpoch` fallback, so when they land in the same bun worker the second mock's "real" reference *is* the first mock and it recurses forever. `sandbox-desktop-stream.test.ts` passes alone on **both** branches (17 pass / 0 fail); running the four together on **pristine `origin/main`** gives `52 pass / 17 fail` — the exact same 17. CI's sharding evidently keeps them apart today. A latent flake worth its own issue, unrelated to #1631.

  Whole-repo `test:unit` on the merged tree: **10503 pass / 22 fail across 1177 files**, and every failure falls in those two pre-existing buckets.

### 12. Adjacent gap worth a separate issue (not a change request)

The PR's motivation says these sessions "can never use that human's personal connections or Documents". Owning the session is necessary but not sufficient: `apps/worker/src/activities/google-drive-publication.ts:59,70` revalidates via `getMembership: typeof getWorkspaceGrant`, i.e. a bare `workspace_memberships` lookup — which a personal workspace never has. So personal-connection execution inside a personal workspace still fails for the *same* root cause this migration repairs. Same class of bug, different seam, separate change — but the fix will feel incomplete to a user until that one lands too.

---

**Verdict: approve.** Minimal, precedent-consistent, correctly restricted, honestly documented, and the tests actually fail when the code is wrong — including at the FORCE-RLS boundary the shared harness cannot see. The two follow-ups I'd like recorded are the untested `personal_owner.status = 'active'` predicate and the 0298 lane; neither justifies holding this.

PR #1635, opened 2026-08-19
fix(ci): close four timeout-guard evasions and harden the budget parser

benyhh:
Follow-ups deferred from #1624, plus fixes for three parser defects an independent review then found in this PR itself.

## Guard evasions closed

The timeout guard added in #1624 only understood `--timeout-seconds N` and millisecond `--timeout N`. Four spellings slipped past it. All are now counted and mutation-tested:

| evasion | before | after |
|---|---|---|
| step-level `timeout-minutes: 90` under a cap of 10 | passed | **CAUGHT** |
| bare `timeout 3600 bun run slow` | passed | **CAUGHT** |
| `--timeout-seconds=1800` equals form | passed | **CAUGHT** |
| `--timeout-seconds ${{ env.BUDGET }}` | passed | **CAUGHT** |

Expression-valued budgets cannot be evaluated statically, so they **fail closed** rather than counting as zero. That now also covers a bare `timeout ${{ ... }}` and a step-level `timeout-minutes` that is not a number.

## Parser defects found in review, and fixed

| defect | effect | fix |
|---|---|---|
| prose read as budget — a comment or `echo` mentioning `timeout 3600` added 60 min | failed a job that was fine | strip comments; anchor to a command position, not any whitespace |
| only the `m` suffix understood — `3600s`, `2h`, `1d` counted as **zero** | silent hole, and `s` is this repo's own idiom | bare/`s`/`m`/`h`/`d` per coreutils |
| `timeout -k 15 7200` read the SIGKILL grace as the duration | counted 15s instead of 2h | consume leading options explicitly |

## Also

The `__dirlock` cleanup step was unguarded — deleting it passed every test. Now asserted, since cache entries are immutable and a lock baked in by a SIGKILLed install is permanent for that key.

Its comment claimed Playwright treats the lock as stale after **~5 minutes**. It does not: Playwright locks via `proper-lockfile` passing only `retries`/`onCompromised`/`lockfilePath`, so the default `stale: 10000` applies — **~10 seconds**. Verified against pinned `playwright-core` 1.62.1 in `node_modules`, independently by two reviews.

## Safety

The parser is **monotone** over the previous revision: it can never compute a *smaller* budget, so no fail-open regression is possible. Computed budgets for all nine currently-evaluated jobs are **identical** to before — this changes nothing about today's workflows.

## Verification

- 13/13 mutation cases, covering every original evasion, all three review-found defects, and a negative control
- 147/147 workflow-touching tests; `bun test scripts/` 623 pass / 0 fail
- `typecheck`, `check:public-hygiene`, `oxlint --deny-warnings` clean

## Note on CI

An earlier run showed `format:check` failing on `packages/xai-subscription/src/fetch.ts`. That file is **not touched by this PR** — it is drift inherited from the branch point `4fa4d1df6`, fixed on main afterwards by `6e200c946` (#1632). Confirmed independently in review, including by test-merging onto current main. No source mutation was made for it, per the moving-main policy.

### DIFF
--- a/.github/actions/playwright-browsers/action.yml
+++ b/.github/actions/playwright-browsers/action.yml
@@ -137,9 +137,13 @@ runs:
     # step. One step is sufficient because it reduces the re-run to a
     # single-browser download, which has never breached the bound - but it does
     # mean a partial set persists for the life of this Playwright version.
-    # A SIGKILLed install can leave a stale __dirlock behind. Playwright only
-    # treats it as stale after ~5 minutes, so baking it into an immutable cache
-    # would tax every later run against this key.
+    # A SIGKILLed install can leave a stale __dirlock behind. Playwright locks
+    # the registry through proper-lockfile passing only retries/onCompromised/
+    # lockfilePath, so the default `stale: 10000` applies - the lock clears
+    # after ~10 seconds, not the ~5 minutes an earlier revision of this comment
+    # claimed. Removing it is therefore cheap hygiene rather than a large
+    # saving: it keeps a lock out of an immutable cache entry and avoids the
+    # onCompromised path entirely.
     - name: Drop stale Playwright dirlock before saving
       if: ${{ always() && steps.restore.outputs.cache-hit != 'true' }}
       shell: bash
--- a/scripts/workflow-timeout-budget.ts
+++ b/scripts/workflow-timeout-budget.ts
@@ -1,0 +1,769 @@
+export type BudgetAnalysis = Readonly<{
+  minutes: number;
+  uncertainties: readonly string[];
+}>;
+
+export type TimeoutStep = Readonly<{
+  run?: string;
+  shell?: string;
+  "timeout-minutes"?: unknown;
+}>;
+
+type WordToken = Readonly<{
+  kind: "word";
+  value: string;
+  dynamic: boolean;
+}>;
+
+type OperatorToken = Readonly<{
+  kind: "operator";
+  value: string;
+}>;
+
+type ShellToken = WordToken | OperatorToken;
+
+const ZERO: BudgetAnalysis = { minutes: 0, uncertainties: [] };
+const LIST_SEPARATORS = new Set(["\n", ";", "&&", "||", "&"]);
+const PROSE_COMMANDS = new Set(["echo", "printf"]);
+const SHELL_COMMANDS = new Set(["bash", "dash", "sh", "zsh"]);
+
+function sum(analyses: readonly BudgetAnalysis[]): BudgetAnalysis {
+  return {
+    minutes: analyses.reduce((total, analysis) => total + analysis.minutes, 0),
+    uncertainties: analyses.flatMap((analysis) => analysis.uncertainties),
+  };
+}
+
+function maximum(analyses: readonly BudgetAnalysis[]): BudgetAnalysis {
+  return {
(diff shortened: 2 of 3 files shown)


benyhh:
Independent post-fix review: **REJECT** at exact head `24c0ddd34de6f58a3ac6a772b6e1d1254763ba92`.

Blocking findings:

1. **P1 — executable unquoted heredocs are erased.** `cat <<EOF\n$(timeout 10h job)\nEOF` returns zero minutes with no uncertainty even though Bash executes the substitution.
2. **P1 — substitutions and static evaluators are treated as prose.** `echo "$(timeout 10h job)"`, assignment/argument substitutions, and `eval 'timeout 10h job'` all return zero/no uncertainty despite executing the timeout.
3. **P2 — the raw-text prefilter permits pre-lex spelling bypasses.** Split/escaped tokens such as `time""out`, `ti\meout`, line-continuation forms, `bun test --time""out`, a static command variable, and `gtimeout` all evade accounting.

The shipped focused suite and broader checks pass, and the first-round cases are fixed, but these fail-open paths still permit the runner-starvation class the guard is intended to prevent. A second repair revision is being dispatched.

benyhh:
Independent final review: **REJECT** at exact head `ca21b756cd86b7b3c95c56fd6441b3cc11a9d437`.

Blocking findings after the second substantive repair revision:

1. **P1 — heredoc terminators remain executable parser tokens.** A legal delimiter named `done` closes modeled loop control early, so a real three-iteration `timeout 10m` path is certified as 10 minutes with no uncertainty instead of 30.
2. **P1 — unquoted variable field splitting is lost.** Static and dynamic cases such as `cmd='timeout --foreground'; $cmd 10h job`, `args='--timeout 36000000'; bun test $args`, and `args=$OPTIONS; bun test $args` all return zero/no uncertainty despite supplying executable timeout command/flag words.
3. **P1 — evaluator and alias indirection remains invisible.** `xargs bash -c 'timeout 10h job'`, `nice bash -c ...`, `ksh -c ...`, and an enabled alias expanding to `timeout` all return zero/no uncertainty; `xargs` can additionally amplify execution count.

Remote identity was verified before and after review. The 153-assertion focused suite and current-main synthetic merge pass, but they do not cover these zero-uncertainty policy bypasses.

Per repository delivery policy, two substantive repair revisions have now been exhausted. An incident/scope review is underway to replace the expanding partial-shell-interpreter design with a bounded, sound policy rather than dispatching a third blind patch.

benyhh:
Incident/scope review conclusion: **close and replace this PR; do not merge `ca21b756`.**

After two substantive repair revisions, the parser grew from a small regex guard into a 1,321-line partial shell evaluator while still failing open on heredoc delimiters, unquoted word splitting, wrapper indirection, and aliases. General shell execution is dynamic; extending the evaluator again would create another unsound patch cycle.

Repository inventory shows the actual scope is much smaller: 13 budget-bearing `run:` steps, all direct/simple, plus three effective Playwright action callers. The replacement will start fresh from current main and:

- add 16 native numeric step caps;
- use native caps as the wall-clock authority;
- replace semantic shell evaluation with a small YAML-level lint that requires caps when budget spellings or the local Playwright action are present;
- count each step cap once against its job cap;
- retain the Playwright retry/process-group tests and the corrected Playwright dirlock assertion.

A replacement PR is being dispatched. This branch is preserved as incident evidence but is not an acceptable merge candidate.

PR #1637, opened 2026-08-19
fix(integrations): stop silently creating personal Connections

benyhh:
> **Revised three times after adversarial review** (`e83f076` → `adf288b3` → `c899c65f` → `5ff5768`, then merged with `origin/main`). Four blocking findings are addressed; two claims in the original description were wrong and are corrected below. Diff of the fixes: `e83f0762..c899c65f`.
>
> **No deprecation window** for the 422 — shipped on the standing position that legacy compatibility is not carried forward as long as our own usage keeps working. The attached condition (verify our own callers empirically) is discharged in [First-party caller audit](#first-party-caller-audit-the-condition-on-shipping-the-422).

## Problem

`startApiIntegrationProviderOAuth` resolved an omitted `ownership` to `personal`:

```ts
const ownership = existingOwnership ?? input.payload.ownership ?? "personal";
```

`AGENTS.md` and `docs/integrations-design.md` both document connection setup as
defaulting to **workspace-owned**, and personal ownership as the explicit
"Connect only for me" choice. This site inverted that for every caller that did
not spell ownership out.

Separately, a *personal* connection is only executable through the immutable
delegation snapshot frozen on a **human's** causal turn, and
`bind_connection_authority` (migration 0256) can mint the `user` authority scope
only for a subject holding an active organization membership. So a non-human
principal on this path produced a row on the `legacy_user` compatibility lane
that can never become a real organization-scoped authority.

## 1. An omitted ownership is now refused, not guessed

The first revision resolved the omission to `workspace`. **That was a real
widening**, and the review was right to block it. Verified by executing the flow
against the merge base and the head, rather than reasoning about it:

| `POST /integrations/oauth/start`, `microsoft-outlook-mail`, `ownership` omitted | resulting `connections.subject_id` |
| --- | --- |
| merge-base `6e200c9` | `user:probe-c2d26794-…` — subject-scoped |
| first revision `e83f076` | `null` — **workspace-shared mailbox** |
| now `adf288b3` | *no row*; `422 ownership is required` |

Both fallbacks are defects, so the endpoint no longer picks either. When the
resolved provider profile allows both ownerships, an omitted value is a **422**
and the caller chooses. A profile allowing exactly one ownership (Gmail, hosted
Slack MCP) is unambiguous and still resolves; a reconnect keeps the existing
row's ownership; the web app always sends an explicit value and is unaffected.
The MCP OAuth start keeps its existing, already-correct `defaultOwnershipFor`
behaviour — it was not the buggy path and changing it would be a gratuitous
break.

### First-party caller audit (the condition on shipping the 422)

Verified by execution, not inspection. Every first-party caller of the affected
endpoint:

| caller | sends `ownership`? | evidence |
| --- | --- | --- |
| `apps/web` account hook (`use-api-integration-accounts.tsx`) | **always** | `ownership = "workspace"` default parameter, applied client-side before the request is built, so the field is always a defined `ConnectionOwnership` |
| `apps/web` Google Drive hook (`use-google-drive-integration.tsx`) | **always** | omits the *prop*, so it takes the hook default above — the wire request still carries `ownership` |
| `packages/sdk` `startApiIntegrationOAuth` | pass-through | library surface; the only in-repo caller is the web hook |
| docs samples, `examples/`, `scripts/`, e2e | none construct a request | `grep` over `startApiIntegrationOAuth` and the raw endpoint path returns only the sites above plus route/label/test definitions |

The wire behaviour is pinned by the capabilities browser e2e, which records the
POST body and asserts `ownership: "workspace"` for both the add-account and
reconnect flows. Deleting the field from the web caller turns it red with
exactly that key missing:

```
- "ownership": "workspace",
+ "returnPath": "/workspaces/.../capabilities?...&api_integration_ownership=workspace"
(fail) custom API control center browser acceptance
```

Restored, the suite is **8 pass / 0 fail**.

No non-human first-party path is affected by the personal fence either: no MCP
tool, worker activity, scheduled task, pack, or script creates connections. The
agent-facing tools (`fiken_contact_create`, `fiken_invoice_draft_create`) only
*use* an existing connection.

## 2. Personal ownership requires a managed human

New `apps/api/src/connection-ownership.ts`, applied at every path that mints a
new personal owner: MCP OAuth start, Integration Definition OAuth start, manual
`POST /connections`, social OAuth start, and the personal-only Google Drive and
Atlassian install routes.

`principalKind` is the trusted allow-list — exactly `human_session` passes;
`api_key`, `configured_key`, `agent_attempt`, `service`, and unknown provenance
all fail closed. The delegated-token contract forbids a `human_session` claim
from carrying `serviceInitiator` or exact agent-attempt authority.

**R1 — aligned on the shared core checks; intentionally stricter for reserved machine namespaces.** The helper now shares `contextIntegrity`, exact authenticated/grant subject identity, `principalKind: "human_session"`, and absence of service-initiator provenance with `requireConnectionAuthorityOwner`. It additionally rejects OpenGeni reserved machine-subject namespaces as defence-in-depth, so the predicates are not identical. The status difference is deliberate: **422** rejects an unavailable ownership value; **403** rejects a caller claiming to be the owner.

**R2 — partially declined, with the underlying concern removed.** The review
asked for an allow-list of `user:` + literal `dev`. I did not do that, because
`docs/embedding.md` states `subjectId` "remains opaque to OpenGeni" and that
hosts "must not encode or infer the kind from a subject-id prefix, because the
host owns that namespace" — a trusted embedding host legitimately signs
`human_session` over a non-`user:` subject, and such an allow-list would both
break those hosts and *be* the prefix-based kind inference the embedding
contract forbids. Instead:

- the subject check is now a **reserved-namespace** check (`api_key`,
  `configured`, `worker`) rather than a prefix deny-list, and
- `connection-ownership.test.ts` asserts every machine subject in its list is
  rejected — now including `sandbox:`, `scheduled_task:`, `attempt:`, and
  `service:`, which the re-review found missing. **Corrected claim:** the list is
  hand-maintained on both sides, so it catches a namespace being *removed* from
  the constant and documents the known machine subjects, but it cannot fail on a
  genuinely new namespace introduced elsewhere. That gap is acceptable only
  because this check decides nothing on its own — a new machine namespace still
  arrives with a non-`human_session` `principalKind`. `docs/integrations-design.md`
  previously overclaimed here and now says exactly this, and
- most importantly, it is **no longer the sole signal on any callback** (§3),
  which was the actual force of the objection.

## 3. All five callbacks are fenced — with a signed claim, not a subject shape

**B1 was correct: my original description asserted a fence that did not exist.**
Only 3 of 5 callbacks were fenced; the MCP and social callbacks had nothing.

Rather than add two more subject-shape checks, the mechanism is replaced. A
callback carries signed state and no live principal, so it cannot re-evaluate
`principalKind` — subject shape was always a guess. Every start path that may
persist a personal owner now stamps a `personalOwnerVerified` claim into its
HMAC-signed state, and **all five** callbacks require it:

| callback | before | now |
| --- | --- | --- |
| Integration Definition OAuth | subject shape | signed claim + subject shape |
| Google Drive | subject shape | signed claim + subject shape |
| Atlassian | subject shape | signed claim + subject shape |
| **MCP OAuth** | **unfenced** | signed claim + subject shape |
| **social OAuth** | **unfenced** | signed claim + subject shape |

These are fail-closed enforcement semantics, not a uniform status contract; every callback retains its existing bounded redirect/error projection.

A state minted before the claim existed lacks it and fails closed — precisely
the one `oauthStateTtlMs` in-flight window across a rolling deploy. This is also
why `oauth-client.ts`'s legacy `ownership: … ?? "personal"` decode can no longer
land a machine-owned row; that decode is retained for its documented purpose.

### About the three repaired test fixtures

Three pre-existing tests hand-minted MCP states with **no** `ownership` field,
which the legacy decode reads as `"personal"` — so they were exercising personal
connections all along, and the new callback fence caught them.

**Their fixtures now say `ownership: "personal"` + the claim, which preserves
that original meaning exactly.** They were not converted to `workspace` to make
them pass. I tried `workspace` first and one of them —
*"oauth callback records non-fatal verification failure…"* — failed, because it
asserts on `loadConnectionCredentialForBroker(..., { subjectId, allowSubjectOwned: true })`,
i.e. it genuinely wants a subject-owned row. That failure is what established
the fixtures had been personal, so they were set to explicit personal instead.
The other two do not assert ownership and would have passed either way; they
were changed for the same reason rather than left inconsistent.

### Threat model for the signed-state claim

The claim is a boolean field **inside** the existing HMAC-signed OAuth state
payload, not a separate token. Each property below is covered by a test in
`api-integration-provider-oauth.test.ts` ("the personal-owner claim cannot be
forged, replayed, or moved to another provider"):

| attack | outcome | why |
| --- | --- | --- |
| **Forge** — splice `personalOwnerVerified: true` into the base64 payload of a legacy state | `reason=state_invalid` | `readSignedState` recomputes the HMAC over the encoded payload and compares constant-time; any edit invalidates it. Setting the claim requires `OPENGENI_INTEGRATIONS_STATE_SECRET`. |
| **Replay** — reuse a genuinely verified personal state | first succeeds, second `reason=state_replayed` | `consumeIntegrationOAuthStateNonce` is a single-use insert on the nonce table. |
| **Definition/fingerprint mismatch** — change `definitionId` while retaining the original fingerprint, then re-sign | `reason=state_invalid` | The callback re-derives the current Integration Definition fingerprint and refuses the mismatch. The integrations state signing secret remains root authority and can mint a fully consistent state; this proves mismatch rejection, not isolation from a signing-secret holder. |
| **Pre-deploy states** — a state minted before this change | refused for personal ownership | `personalOwnerVerifiedInState` accepts only an exact boolean `true`; absent, `false`, `"true"`, `1`, and `null` all read false (unit-tested). This is the intended rolling-deploy behaviour, bounded by the 10-minute `oauthStateTtlMs`. |
| **Carry between flows** — feed one flow's state to another flow's callback | refused, but **this was fixed in review, not inherited** — see below | Each callback now parses a flow discriminator first. |

**Correction — the earlier version of this table was wrong about cross-flow
carry.** The re-review executed a probe I had only reasoned about. Google Drive
and Atlassian states carry **no `ownership` field and no provider identity**
(both connectors are personal-only by construction), so the claim was their
*sole* ownership gate — and their return path is byte-identical to
`/workspaces/<ws>/capabilities`, a path the MCP start signs from **caller
input**. The probe: a human's `ownership: "workspace"` MCP state carrying that
return path **sailed past the Atlassian callback's personal-owner fence**
(`reason=http_503` — only missing provider settings stopped it), while the
identical machine-minted state was correctly refused `http_422`.

That was never an escalation — same subject, and the callback independently
rechecks `connections:write`, the same permission the direct install route
requires — and it is pre-existing confusion that this PR narrows rather than
introduces. But the property I claimed did not exist.

It does now. Both states carry an exact `kind` (`google_drive_oauth`,
`atlassian_oauth`), checked before anything else in their parsers, so a state is
bound to the flow that minted it. A pre-cutover state carries no kind and is
refused — the same bounded `oauthStateTtlMs` window the personal-owner claim
already fails closed on, so this adds no new breakage. Pinned by cross-flow
tests on both connectors (sabotage G18/G19), each with a positive control.

**How tightly is the claim bound?** The claim is signature-bound to the exact state payload. Depending on the flow, that payload binds explicit provider/ownership fields or an exact flow kind, along with account/workspace/subject. It does not attenuate the state-signing root authority. It is deliberately **not** bound to a connection id, because on a first connect no connection exists yet. Where a
`connectionId` *is* present (reconnect), the pre-existing fences still apply
unchanged: exact-owner match, version CAS, and provider-principal equality. The
claim never widens what the state could already do; it only removes the ability
to persist a personal owner that no human authorized.

One deliberate detail: the start paths stamp the *principal's capability*
(`personalOwnershipAllowed`), not "this flow is personal". A human starting a
workspace-ownership flow also gets `true`. That is harmless because
`personalOwnerStateAccepted` consults the claim only when
`state.ownership === "personal"`, and `ownership` is itself inside the signed
payload, so it cannot be flipped without the secret.

## 4. Sabotage proof — 20/20 guards pinned, re-run after merging main

The first review disabled each guard and found **4 of 9 produced zero failures**.
Every enforcement point is now pinned. Each row disables exactly one guard and
re-runs the ten suites touching these paths (baseline **144 pass / 0 fail**),
re-run *after* merging `origin/main`:

| # | guard sabotaged | result |
| --- | --- | --- |
| G1 | core rule → `return true` | **RED** 10 fail |
| G2 | `contextIntegrity` requirement | **RED** 1 fail |
| G3 | reserved machine namespaces (list emptied) | **RED** 3 fail |
| G4 | signed-state claim gate → `true` | **RED** 6 fail |
| G5 | provider-oauth 422-on-omission → `"workspace"` | **RED** 1 fail |
| G6 | provider-oauth start principal fence | **RED** 1 fail |
| G7 | provider-oauth callback fence | **RED** 1 fail |
| G8 | MCP start principal fence | **RED** 1 fail |
| G9 | MCP callback fence *(was absent)* | **RED** 1 fail |
| G10 | google-drive callback fence | **RED** 1 fail |
| G11 | atlassian callback fence | **RED** 1 fail |
| G12 | social callback fence *(was absent)* | **RED** 1 fail |
| G13 | social start fence | **RED** 1 fail |
| G14 | drive/atlassian install start fences | **RED** 1 fail |
| G15 | manual `POST /connections` fence | **RED** 1 fail |
| G16 | Gmail/Slack personal-only default | **RED** 5 fail |
| G06 | reserved-namespace **term inside the predicate** *(re-review finding)* | **RED** 2 fail |
| G17 | social start claim hardcoded `true` *(re-review finding)* | **RED** 1 fail |
| G18 | google-drive flow-kind discriminator | **RED** 1 fail |
| G19 | atlassian flow-kind discriminator | **RED** 1 fail |

Two findings from re-running this after the merge, both worth stating plainly:

- **G06 was a guard that looked pinned but wasn't.** The enumeration test's
  helper moved only `grant.subjectId` and left `authenticatedSubjectId`, so the
  anti-substitution term fired first and the reserved-namespace term never
  executed. The helper now moves both, and substitution keeps its own explicit
  case.
- **G3 and G18 initially reported zero failures.** G18 was real — the flow-kind
  test existed only for Atlassian, so the Drive equivalent was added. G3 was a
  **false green in my own harness**: patches were applied with a bare
  `str.replace`, so a match string that had gone stale silently no-op'd and a
  perfectly good guard read as unpinned. The harness now asserts the patch
  changed the file. A measurement that fails open is worse than no measurement.

The callback fences are pinned by minting the genuine rolling-deploy artefact —
a real signed state with the `personalOwnerVerified` claim stripped — plus a
positive control re-signing the identical payload with the claim restored, so
the claim is provably the only difference.

One deliberate green: `assertOwnershipAllowed` in the provider-oauth start is
inert by construction, because no curated Definition resolves to a personal-only
profile. It is kept as pre-wiring for a future Definition targeting Gmail or
`slack.com`, and now carries a comment saying so, so it is not read as a gap.

## Gmail and Slack hosted MCP

Unchanged and re-verified. They are personal-only by profile
(`allowedOwnership: ["personal"]`), which is the opposite rule, so they are
handled by going *through* the profile: `assertOwnershipAllowed` still runs
before the principal fence, a personal-only profile still resolves an omitted
ownership, and a non-human principal is **refused** rather than downgraded to
the workspace ownership those profiles forbid.

## Existing rows — corrected

**R4 was right and my original description was wrong in one direction.** Machine-owned
personal rows **are** runtime-resolvable today:
`personalConnectionDelegationSourceForGrant` returns a subject source for an
`api_key` grant, so an api_key-owned connection resolves for that api_key. They
are inert only in the sense that a machine is never a frozen initiating human.
They also remain listable and readable, and credential-broker refresh-token
renewal is untouched. No schema change, no data migration, nothing rewritten.

What they lose is **any re-consent path**: interactive OAuth start refuses the machine principal, callbacks fail closed through their existing flow-specific redirect/error projections (for example Integration Definition OAuth reports `connection_conflict` and social OAuth reports `not_authorized`), and migration 0256 makes the owner column immutable, so the owner cannot be converted in place. **Remediation when such a connection's
refresh token finally fails:** create a replacement connection — workspace-owned,
or personal connected by the human who should own it — repoint the capability at
it, then revoke the stale row. Survey SQL is in the changeset. *(Production
numbers requested from the coordinator; I will report whether they change this
answer.)*

## `configured` mode — called out (now a read-before-upgrading changeset section)

`configured` product-access-mode deployments **lose personal connections
entirely** (Gmail, hosted Slack MCP, Google Drive, Atlassian). The shared
configured key resolves to `principalKind: "configured_key"`, and its subject
comes from a caller-supplied `x-opengeni-subject` header that proves nothing.
Intended: a shared operator key cannot supply the per-human consent a personal
connection represents. `local` mode is unaffected — its bootstrap `dev` subject
is a `human_session` and genuinely is the sole operator (the review independently
verified `dev` is unreachable outside `local`).

This is now the **first item** in the changeset's "Read before upgrading"
section, naming `OPENGENI_PRODUCT_ACCESS_MODE=configured` explicitly, so a
self-hosted operator meets it before upgrading rather than after.

## Left alone, deliberately

- **`google-drive.ts` / `atlassian.ts` ownership.** No `?? "personal"` exists;
  their start requests carry no `ownership` field and they always write
  `subjectId: state.subjectId`. Deliberately personal-only act-as-a-person
  connectors. Defaulting them to workspace would widen access. Only the
  non-human-principal hazard applied, and that is fixed.
- **`fiken.ts`** hardcodes `subjectId: null`. Already correct.
- **`oauth-client.ts` legacy state decode.** Retained for its documented
  purpose; now genuinely covered by the signed-claim fence (§3).
- **`PATCH /connections/:id { subjectId }`** — pre-existing, not a bypass
  (migration 0256's immutability trigger holds and the row stays workspace-owned)
  but it answers with an opaque 500. Filed as a follow-up: #1640.

## Verification

- `typecheck`, `lint`, `format:check` (run last), `check:docs-refs`,
  `check:public-hygiene` (post-commit), `check:migration-ordinals` — all clean.
- Full `apps/api` suite: **1326 pass / 3 skip / 17 fail** against a **freshly
  re-established `origin/main` baseline** (`13c3af4f`, which includes migration
  0302 and #1639), run in a separate clean worktree with its own `bun install`:
  main is 1303 pass / 3 skip / 17 fail, and the two sorted failure-name sets are
  **identical**. All 17 are the pre-existing sandbox desktop/viewer-stream
  suites.
- `origin/main` **merged** into the branch (not rebased), which also clears the
  `Validate Helm chart` red check — that assertion (`single-node profile did not
  render Garage`) exists only on main via #1639 and was running against an
  isolated pre-#1639 head.


benyhh (commented):
> **Note:** GitHub rejects `--request-changes` and `--approve` on a PR authored by the same account, so this is posted as a comment. **Treat it as an explicit REQUEST CHANGES.**

## Verdict: **REQUEST CHANGES** — do not merge as-is

Independent adversarial review of `e83f076` against merge-base `6e200c9`, in a clean detached worktree with its own `bun install`. I did not take any claim in the description on faith; every one is checked against source below, and I ran sabotage experiments to find out which of the nine new enforcement points are actually pinned by a test.

**Bottom line:** I found **no exploitable privilege escalation** — the central rule is correct, fails closed on unknown provenance, and cannot downgrade Gmail/Slack. But I am not willing to stake production access control on it in its current shape, for four reasons: **one factual claim in the description is wrong**, **two of the five callbacks that can persist a personal owner are unfenced**, **four of the nine new enforcement points have no failing test** (verified by sabotage), and **the change does widen access on one path**, contrary to its own framing.

---

## Blocking

### B1. The description's claim about `oauth-client.ts:1388` is false — that callback has no fence

> *"`oauth-client.ts:1388` `connectionOwnership(payload.ownership) ?? "personal"` is a legacy signed-state decode … Left as-is; **the callback-side subject-shape fence covers the machine case**."*

There is no callback-side subject-shape fence in `oauth-client.ts`. The three fences added are in `provider-oauth.ts` (`requireProviderOAuthOwner`), `google-drive.ts`, and `atlassian.ts`. Complete call-site list on this head:

```
apps/api/src/integrations/provider-oauth.ts:731   isPersonalConnectionOwnerSubject(state.subjectId)
apps/api/src/integrations/atlassian.ts:1514       isPersonalConnectionOwnerSubject(state.subjectId)
apps/api/src/integrations/google-drive.ts:2013    isPersonalConnectionOwnerSubject(state.subjectId)
```

Meanwhile `apps/api/src/integrations/oauth-client.ts:1402` still decodes `ownership: connectionOwnership(payload.ownership) ?? "personal"` and `:713` persists `subjectId: ownerSubjectId` from it, with no subject-shape check anywhere on that path. `apps/api/src/integrations/social-oauth.ts:319` writes `subjectId: state.ownership === "personal" ? state.subjectId : null`, also unfenced.

So the MCP OAuth callback and the social OAuth callback can both still land a machine-owned personal Connection from an in-flight state, for one `oauthStateTtlMs` window across a rolling deploy — the *exact* hazard the other three fences exist to close. The changeset also says the rule is applied to five start paths "plus their callbacks", which is not what shipped.

Practical severity is low (bounded by the state TTL; post-deploy states have already passed the start fence). But it is the one place where the description asserts a guard that does not exist, and the treatment is inconsistent across five structurally identical callbacks. Either add the two missing fences, or state plainly in the body and changeset which callbacks are deliberately not fenced and why.

### B2. Four of the nine enforcement points are unpinned — sabotage produces zero failures

I disabled each new guard in turn and re-ran the relevant suites.

| Guard | Sabotage | Result |
| --- | --- | --- |
| `isPersonalConnectionOwnerPrincipal` core rule | `return true` | **RED** — 6 fail (unit + `connections routes > a non-human principal cannot own a personal connection on any create path`) |
| MCP OAuth start fence (`oauth-client.ts:380`) | pass `true` for `personalOwnershipAllowed` | **RED** — 1 fail |
| `defaultOwnershipFor` (Gmail/Slack personal-only) | force `"workspace"` | **RED** — 5 fail |
| provider-oauth start fence + workspace default | (covered by the two new DB tests) | RED |
| `google-drive/install` + `atlassian/install` start fences | (covered by the new route test) | RED |
| **`provider-oauth` callback `requireProviderOAuthOwner`** | `if (Math.random() < 0) …` | **GREEN — 0 fail** |
| **`google-drive.ts` callback fence** | `if (false && …)` | **GREEN — 0 fail** |
| **`atlassian.ts` callback fence** | `if (false && …)` | **GREEN — 0 fail** |
| **`social/oauth/start` personal fence** | `if (false && …)` | **GREEN — 0 fail** |

Run across `google-drive.test.ts`, `google-drive-oauth-isolation.test.ts`, `atlassian-helpers.test.ts`, `api-integration-provider-oauth.test.ts`, `connections-routes.test.ts`, `connection-ownership.test.ts`, `app.test.ts`, `social-api.test.ts`, `social-oauth.test.ts`, `social-connector-flow.test.ts` — i.e. every file that touches those code paths (`grep -l` on the callback/handler symbols returns exactly that set).

A security guard with no failing test is unpinned; the next refactor deletes it silently. This is a security-boundary change, so I'd want a test for every one of the nine points, particularly the three callback fences that exist *specifically* for a rolling-deploy window nobody will remember in six months.

---

## Strong recommendations

### R1. `isPersonalConnectionOwnerPrincipal` diverges from the codebase's own predicate for the same question

`apps/api/src/connection-authority-owner.ts:15` already answers "is this the exact human owner of a personal Connection?", one file away:

```ts
if (
  !access.contextIntegrity ||                                  // <-- the new helper omits this
  access.authenticatedSubjectId !== access.grant.subjectId ||
  access.grant.principalKind !== "human_session" ||
  access.grant.serviceInitiator ||
  access.grant.serviceInitiatorContext ||
  access.grant.subjectId.startsWith("api_key:")
) throw new HTTPException(403, …);
```

`apps/api/src/routes/connections.ts:722` (the Slack-bot personal-destination check) also requires `contextIntegrity`. The new helper drops it. `contextIntegrity` (`packages/core/src/access/index.ts:84`) is the anti-substitution invariant: every workspace grant agrees on `principalKind`/delegated/`serviceInitiator`, and the selected grant's account resolves to exactly one matching account grant. The reachable divergence is a grant whose `accountId` has no matching entry in `context.accountGrants` (e.g. a surviving `workspace_memberships` row in an organization where the membership is no longer active) — `requireConnectionAuthorityOwner` refuses it, the new helper lets it mint a personal Connection, which then lands on `legacy_user`. That is a narrow version of the very hazard this PR exists to close.

Two helpers answering the same authority question with different strictness (and different status codes: 403 vs 422) is exactly how a boundary rots. Either reuse `requireConnectionAuthorityOwner`'s predicate or add `!access.contextIntegrity` and a test for it. No test currently exercises `contextIntegrity: false` on either helper.

### R2. `NON_HUMAN_SUBJECT_PREFIXES` is a deny-list, so the "belt-and-braces" fails **open**

```ts
const NON_HUMAN_SUBJECT_PREFIXES = ["api_key:", "configured:"] as const;
export function isPersonalConnectionOwnerSubject(subjectId: string): boolean {
  return !NON_HUMAN_SUBJECT_PREFIXES.some((p) => subjectId.startsWith(p));
}
```

Every other subject shape reads as human — including `worker:first-party-mcp` (`packages/runtime/src/index.ts:5135`), and any future `service:`/`system:`/`sandbox:` prefix. Today that is safe only because `principalKind` carries the decision; the prefix check adds nothing beyond the two cases it names, and it is the *sole* signal on the three callbacks (where a machine-minted state carrying, say, a `worker:` subject would pass). Given the file's own comment calls this "belt-and-braces for a trusted embedding host that signs a human claim over a machine subject", an allow-list (`user:` prefix, plus the exact literal `dev`) would actually deliver that and would fail closed on every future subject shape. Note `packages/api` already has ~six sites constructing `principalKind: "human_session"` grants for Slack-originated actors (`slack-interactions.ts`), so "a human `principalKind` on a non-`user:` subject" is not hypothetical inside this codebase.

### R3. This **does** widen access on one path; the description says it never does

The prompt's framing — and the PR's — is "only ever narrows or makes explicit". That is not true of the headline change. On `POST /v1/workspaces/:workspaceId/integrations/oauth/start` with `ownership` omitted:

- **before:** `subject_id = <caller subject>` — visible only to that subject (`connectionSubjectVisibility`, `packages/db/src/index.ts:7952`)
- **after:** `subject_id = NULL` — **workspace-owned**, resolvable by every member and every agent turn in the workspace

The affected Definitions (`packages/capabilities/src/integration-definitions.ts:323`) are `google-drive`, `microsoft-outlook-mail`, `microsoft-outlook-calendar`, `microsoft-outlook-contacts`, `microsoft-onedrive` — all `www.googleapis.com` / `graph.microsoft.com`, none matching a built-in profile, so all resolve to `DEFAULT_OAUTH_PROFILE` → `workspace`. A newly connected **Outlook mailbox** that previously stayed with the connecting subject is now shared workspace-wide.

I accept the product argument: `AGENTS.md` and `docs/integrations-design.md` both document workspace as the default, and `apps/web/.../use-api-integration-accounts.tsx:229` already defaults to `"workspace"`, so the server was contradicting its own UI. Only direct API/SDK callers change. But `ApiIntegrationOAuthStartRequest.ownership` is `.optional()` in the SDK, and for those callers this is a silent narrow→broad flip. The fail-closed form is a **422 "ownership is required"** on omission (at least for a deprecation window) rather than silently picking the broader value. At minimum, the changeset should say in as many words that affected callers' *new* connections become workspace-shared; "only direct API/SDK callers change behaviour" undersells it.

### R4. Machine-owned personal rows lose their only re-consent path with no remediation

Verified: they stay listable/readable (`connectionSubjectVisibility` returns own-subject rows) and — contrary to the description — they **are** runtime-resolvable. `personalConnectionDelegationSourceForGrant` (`packages/core/src/domain/personal-connection-delegations.ts:53`) returns `{ kind: "subject", subjectId: grant.subjectId }` for an `api_key` grant (it only returns `none` for `agent_attempt`/`service`/`serviceInitiator`), so an api_key-owned personal Connection resolves today for that api_key. So "an unusable Connection … that no delegation snapshot can select" is not accurate for the API-key case; the real reason they are inert is that a machine is never a *frozen initiating human*.

That matters operationally: such rows are working integrations that now have no supported way to re-consent — start 422s, the callback fence 422s, and the DB refuses an owner change (see R5). When the stored refresh token finally fails, the integration is dead. The survey SQL is good; please pair it with a stated remediation (recreate as workspace-owned / an operator conversion), because "that is intentional" is not a migration path.

### R5. Adjacent, pre-existing: `PATCH /connections/:id { subjectId }` is a 500, not a 422

I went looking for a seventh creation path and found one: `apps/api/src/routes/connections.ts:794` still honours `payload.subjectId` via `writableSubjectId`, with no personal-owner principal fence. I reproduced it end-to-end (delegated `service` principal, `connections:write`, POST workspace-owned then PATCH a personal owner onto it):

```
[REVIEW machinepatch] POST(workspace) status 201
[REVIEW machinepatch] PATCH status 500 {"error":{"status":500,"code":"internal_error",…}}
[REVIEW machinepatch] rows [ { subjectId: null } ]
```

**The fence holds** — migration `0256_connection_authority_delegation.sql:221` raises `connection owner authority is immutable` (SQLSTATE `23514`) and the row stays workspace-owned. So this is *not* a bypass. But it is a personal-ownership path that answers with an opaque 500 for humans and machines alike, which contradicts the "explicit 422, never a silent fallback, never a 500" principle this PR is establishing. Worth a follow-up (reject `subjectId` in `UpdateConnectionRequest`, or 422 at the route).

---

## What I tried to break, and what held

**Claim 4 — `principalKind` as trusted provenance: holds.** `DelegatedAccessTokenPayload.principalKind` is **required** (`DelegatedAccessPrincipalKind`, not `.optional()`), covered by the token HMAC, and the `superRefine` forbids `human_session` alongside `serviceInitiator` or any exact-attempt claim (`packages/contracts/src/index.ts:2169`). The worker mints only `service`/`agent_attempt` — including the case that would be a real hole, `signFirstPartyDelegatedBearer` passing a **human** `subjectId` with a machine `principalKind` (`packages/runtime/src/index.ts:5132`); the helper requires both and correctly refuses. `api_key` → `principalKind: "api_key"`; account-scoped API keys fall through to `getWorkspaceGrant` with **no** provenance, so `principalKind` is `undefined` → refused. Absent/unknown provenance fails closed, and sabotage confirmed the `principalKind: undefined` case is tested.

**Claim 4 — can a caller influence it?** No. It is derived per auth path (`packages/core/src/access/index.ts`), never read from a request field. `hostedHumanSessionPrincipalKind` only *upgrades* to `human_session` when mode is `managed` **and** every context grant is already non-delegated human_session.

**Claim 3 — the `dev` judgement: sound.** `dev` is minted only by the `local` bootstrap (`resolveAccessContext`, `bootstrapWorkspace` with `accountExternalSource: "opengeni:local"` → `principalKind: "human_session"`). `configured` mints `configured:<x-opengeni-subject>` → `configured_key` → refused on both signals. There is no path that produces the literal subject `dev` in `configured`/`managed`. Asserting `dev` via a delegated token requires the first-party delegation secret, i.e. the worker/trusted-embedding-host boundary that already exists. And in `local` mode the API is unauthenticated by default (`authRequired` defaults `false`), so `dev` genuinely *is* the sole operator — allowing it grants nothing that mode does not already grant. **Note for the record:** this does mean `configured`-mode self-hosted deployments lose personal Gmail/Slack/Drive/Atlassian connect entirely, with no workspace-owned alternative for the personal-only profiles. Defensible (`x-opengeni-subject` is caller-controlled and proves nothing), but the description mentions only local/self-hosted-*local*; `configured` deserves an explicit line.

**Claim 5 — Gmail/Slack cannot be downgraded: holds.** `assertOwnershipAllowed(profile, ownership)` runs **before** `assertConnectionOwnershipAllowedForPrincipal` on both start paths, so a personal-only profile 422s the whole flow rather than falling back. No curated Definition resolves to a personal-only profile today (`builtInOAuthProfileFor` over the five Definitions returns `null` → `DEFAULT_OAUTH_PROFILE`), so the newly added `assertOwnershipAllowed` in `provider-oauth.ts` is currently inert and cannot break an explicit `workspace` request; if a Definition ever targets `slack.com`, the `providerDomains: ["slack.com"]` match keeps the fence. Inverting `defaultOwnershipFor` went red in 5 tests.

**Claim 2 — Drive/Atlassian correctly left personal-only: holds.** `GoogleDriveOAuthStartRequest` (`packages/contracts/src/google-drive.ts:201`) and `AtlassianOAuthStartRequest` (`packages/contracts/src/atlassian.ts:89`) carry no `ownership` field; every write in both connectors is `subjectId: state.subjectId` / `input.subjectId`; there is no shared helper with a hidden default. Defaulting them to workspace would indeed have widened access. The author read this correctly.

**Claim 1 — did it widen anywhere else?** I walked all six start paths, the PATCH route, both manual-create routes, `fiken.ts` (`subjectId: null`, correct), and the reconnect derivations. `existingOwnership` short-circuits before the default on **both** OAuth start paths, so no reconnect flips ownership. `writableSubjectId` still 403s a cross-subject write. `createConnectionSubjectId` guards the deprecated `subjectId` field too (the new check fires on `subjectId !== null`, covering both spellings). The only widening is R3.

**Claim 7 — 422 shape: holds.** `HTTPException(422)` matches `assertOwnershipAllowed`'s existing convention. On the MCP path it passes through `oauthStartApiError` (`oauth-client.ts:1690`), which preserves an `HTTPException` cause's status and message as `ApiHttpError(422, code: "validation_failed", retryable: false)` — not a 500. The browser callbacks convert to a redirect carrying `http_422`, which is a refusal, not a fallback. Only R5 produces a 500, and that is pre-existing.

**Claim 8 — Phase D: clean.** The helper reads `principalKind`, `authenticatedSubjectId`, `grant.subjectId`, and the two service-initiator fields. Nothing infers human authority from `created_by`, connection attribution, a default workspace, resource name, or current access.

**Claim 6 — existing rows: verified, with the correction in R4.** `personal-connection-delegations.ts` and the credential broker are untouched by the diff (14 files, none of them those).

---

## Checks

All run on the exact head commit, clean worktree:

- `typecheck` — clean (all projects)
- `lint` (`oxlint --deny-warnings`) — clean
- `format:check` — clean (2765 files)
- `check:docs-refs` — passed
- `check:public-hygiene` — passed (post-commit, clean tree)
- `check:migration-ordinals` — ok (310 migrations, base highest 0301)

**Baseline claim independently verified — it is exactly right.**

| | pass | skip | fail |
| --- | --- | --- | --- |
| merge-base `6e200c9` | 1303 | 3 | 17 |
| head `e83f076` | 1312 | 3 | 17 |

`diff` of the two sorted failure-name sets (timings stripped): **identical, 17 = 17.** All 17 are the pre-existing sandbox desktop/viewer-stream suites (`P1.4`, `P3.2`, `P4.2`) and are unrelated. The +9 accounts exactly for the 6 new unit tests + 2 new provider-oauth tests + 1 new route test.

---

## Could not verify

- **Real-world population of machine-owned personal rows.** No production database access, so I cannot say whether R4 affects zero rows or many. The survey SQL is the right instrument; please run it before merging and put the number in the PR.
- **`microsoft-*` OAuth end-to-end.** The new DB tests exercise only the `google-drive` Definition; the Outlook Mail default flip (R3) is reasoned from the shared `providerOAuthProfile` resolution, not executed.
- **Behaviour under an actual rolling deploy** with genuinely in-flight pre-fence states (B1) — reasoned from the state TTL and decode path, not simulated.

---

## To get to approve

1. Fence the MCP OAuth and social callbacks, or correct the PR body + changeset to say exactly which callbacks are covered (**B1**).
2. Add a failing test for each of the four unpinned enforcement points (**B2**).
3. Add `!access.contextIntegrity` (or reuse `requireConnectionAuthorityOwner`) plus a test (**R1**).
4. Flip `isPersonalConnectionOwnerSubject` to an allow-list (**R2**).
5. State the widening plainly in the changeset, and say why a 422-on-omission deprecation window was rejected (**R3**).
6. Run the survey SQL and state the remediation for existing machine-owned rows (**R4**).

None of these needs a rebase; `main` moving does not invalidate this review.

benyhh (commented):
> **Note:** GitHub rejects `--approve` on a PR authored by the same account, so this is posted as a comment. **Treat it as an explicit APPROVE.**

## Verdict: **APPROVE** — mergeable

Independent re-review of head **`c899c65f`** against current `main` **`13c3af4f`** (merge-base `6e200c9`), in two fresh detached worktrees with their own `bun install`. I re-targeted mid-review when the head moved from `adf288b3`; everything below is measured on `c899c65f`.

Every blocking item from the previous review is genuinely fixed, and the fix is better than what I asked for. The signed-state claim is a real mechanism, not a re-labelling: it rides **inside** the HMAC-covered payload, it is derived from the live principal at exactly one decision point, and I could not forge, replay, or carry it to a machine principal. I ran my own 27-point sabotage matrix (not the author's 16) — **24 red, 3 green**, and none of the three greens is a live hole.

I found two things the PR body/docs overstate, and I proved one of them with an executed probe. Neither is exploitable. One of them is in a *shipped* doc and I'd fix it in this PR; I would not block the merge on it.

---

## 1. The signed-state claim — forge / replay / binding

**Forge: closed.** `createSignedState` (`packages/github/src/index.ts:142`) JSON-stringifies the *whole* payload — claim included — base64url-encodes it, and HMAC-SHA256s the **encoded string**; `readSignedState` recomputes and compares with `timingSafeEqual` on equal-length buffers. `personalOwnerVerified` is therefore inside the signature, not appended alongside it. Setting it requires `OPENGENI_INTEGRATIONS_STATE_SECRET`. The new forge test (splice the field into the base64, keep the old signature) fails with `state_invalid` **because of the signature**, not incidentally — I confirmed the reader returns `null` before it ever parses the JSON.

**Replay: closed, and atomically.** `consumeIntegrationOAuthStateNonce` (`packages/db/src/index.ts:11992`) is an `insert … onConflictDoNothing({ target: nonce }).returning()` against a table whose `nonce` column is the **PRIMARY KEY** (`0042_integration_oauth_state.sql:21`). That is genuinely single-use under concurrency — two racing callbacks, one row, one `true`. It is also *global*, so a nonce burned in one flow cannot be reused in another. The expired-row sweep in the same transaction cannot resurrect a nonce, because an expired nonce implies an expired state, which the per-flow `oauthStateTtlMs` check (10 min) rejects first.

**Binding: real, but narrower than the PR body claims — see §3.** The claim is bound to whatever else is in that payload. For MCP OAuth that is `ownership` + `providerDomain` + `mcpUrl` + `resource` + client/AS identity; for Integration-Definition OAuth it is `ownership` + `definitionId` + `definitionFingerprint`, and I verified the fingerprint is re-derived from the *current* Definition and compared at the callback (`provider-oauth.ts:289`), so the provider-swap test is a genuine negative control. (It does not, and cannot, contain a *secret holder*: re-signing with a consistent `definitionId` **and** its matching fingerprint would pass. That is fine — per `docs/embedding.md`'s trust model the delegation/state secrets are root authority — but the threat-model row overstates what it proves.)

**Confusion (one human flow, one machine flow, concurrent): closed.** The claim has exactly one source of truth, `isPersonalConnectionOwnerPrincipal(access)`, evaluated on the live principal at each start route. A machine's state carries `false` in every flow. I walked every `createSignedState(requireIntegrationsStateSecret(...))` call site — six of them — and no machine-reachable path can produce a `true` claim.

**The design is also *necessary*, which I checked rather than assumed.** `getWorkspaceGrant(db, subjectId, workspaceId)` without a provenance hint returns a grant with **no** `principalKind` (`packages/db/src/index.ts:2083-2106`), so a callback genuinely cannot re-derive the principal. Subject shape really was the only alternative, and it really was a guess. Replacing it was the right call.

## 2. Rolling deploy — fails closed, and the failure is sane

Pre-claim states lack the field; `personalOwnerVerifiedInState` accepts only `=== true`, so absent / `false` / `"true"` / `1` / `null` / `{}` all read false (unit-pinned). Bounded by the 10-minute state TTL. What the user actually sees:

| callback | outcome | fence runs before nonce burn? |
| --- | --- | --- |
| MCP OAuth | `302 …integration_oauth=error&stage=state_verify&reason=state_invalid` | yes |
| Integration Definition | `302 …integration_oauth=error&reason=connection_conflict` | yes |
| Google Drive | `302 …google_drive=error&reason=http_422` | yes |
| Atlassian | `302 …atlassian=error&reason=http_422` | yes |
| social | `302 …social_oauth=error&reason=not_authorized` | no (nonce burned first) |

All five are refusals, all five are 302s, none is a 500 and none loops. Retrying after the deploy completes works. Two cosmetic notes: `connection_conflict` renders as *"The connection changed. Start again."*, which is accidentally decent advice but the wrong reason code (`state_invalid` is what actually happened); and the social path burns the nonce before the fence, so that user must restart rather than re-hit the callback. Neither is worth a revision.

## 3. **Finding — the claim does cross flows into Drive/Atlassian. Proven, not argued.**

The PR body says the claim is "bound to the state's subject, account, workspace, **provider identity and ownership**", and the threat-model table says feeding one flow's state to another flow's callback is "refused … Drive requires an exact `returnPath`". **Both are wrong for the two personal-only connectors.**

Google Drive and Atlassian states carry **no `ownership` field and no provider identity** — for them the claim is the *sole* ownership gate. And their "exact" return paths are byte-identical to each other:

```
google-drive.ts:110  GOOGLE_DRIVE_RETURN_PATH = (ws) => `/workspaces/${ws}/capabilities`
atlassian.ts:80      ATLASSIAN_RETURN_PATH    = (ws) => `/workspaces/${ws}/capabilities`
```

…and the MCP OAuth start signs a **caller-supplied** `returnPath` (`safeReturnPath` only rejects absolute/protocol-relative paths). The Slack-bot install and Fiken starts mint that same path too.

I wrote a probe against the real routes rather than reasoning about it. A human starts `POST /connections/oauth/start` with `ownership: "workspace"` and `returnPath: "/workspaces/<ws>/capabilities"`, then feeds the returned state to `GET /v1/integrations/atlassian/callback`:

```
[PROBE human   -> atlassian callback] 302 …/capabilities?atlassian=error&reason=http_503
[PROBE machine -> atlassian callback] 302 …/capabilities?atlassian=error&reason=http_422
```

`http_503` is "no Atlassian client configured" — i.e. the human's **workspace-flow** state sailed past the personal-owner fence, past the `connections:write` grant recheck, and past the nonce consume, and only stopped at provider settings. The machine's identical state (claim `false`) is refused at exactly the fence with `http_422`.

**This is not an escalation and it does not block.** The row that would be created is owned by the same authenticated human the claim verified, and `requireCallbackGrant` independently re-checks `connections:write` — the same permission `/connections/atlassian/install` requires. So the human gains nothing they could not do directly, and the machine is closed. It is also *pre-existing* state confusion that this PR **narrows**: before it, a Slack-bot-install or Fiken state (identical shape, no claim) could have driven the Atlassian callback into a personal row; now it 422s.

But it does mean the disclosed "principal capability, not flow" detail is not harmless in the strict sense claimed — it is harmless because of the *permission recheck*, not because the claim can't reach a personal decision. Please correct the two statements. **Follow-up (not this PR):** give every state payload a purpose discriminator and check it, the way `social-oauth.ts:695` already does with `payload.kind !== "social_oauth"`. That closes the whole class.

## 4. **Finding — the enumeration test does not enumerate every machine subject OpenGeni mints**

`docs/integrations-design.md` (shipped) states: *"`connection-ownership.test.ts` enumerates every machine subject OpenGeni mints and asserts each is rejected, so a new machine namespace fails that test rather than silently passing the list."* The test's own header comment repeats it. It is not true. The list has three entries; OpenGeni mints at least these machine subjects:

```
packages/core/src/access/index.ts:290                 api_key:<id>            ✅ listed
packages/core/src/access/index.ts:394                 configured:<header>     ✅ listed
packages/runtime/src/index.ts:5135                    worker:first-party-mcp  ✅ listed
packages/runtime/src/sandbox/codemode-authority.ts:41 sandbox:<attemptId>     ❌ NOT listed
apps/worker/src/activities/scheduled-tasks.ts:886,1782 scheduled_task:<id>    ❌ NOT listed
packages/db/src/session-queue-commands.ts:2358,2645   attempt:<attemptId>     ❌ NOT listed (audit-only)
```

`sandbox:<attemptId>` is not hypothetical — it is a real `signDelegatedAccessToken` **grant** subject. The mechanism cannot deliver the property it claims: the list is hand-maintained, so nothing *makes* a new machine namespace fail; the test only re-asserts the three someone remembered.

**Not exploitable.** `sandbox:` tokens carry `principalKind: "agent_attempt"` and only `codemode:call`, so they are refused twice over by the decisive gate. The reserved list is defence-in-depth against a delegation-secret holder, who is root authority anyway. **But this is the mechanism offered as the replacement for the allow-list I asked for**, and its shipped description is wrong, which per `AGENTS.md` is a bug. Two lines: add `sandbox` and `scheduled_task` to `RESERVED_MACHINE_SUBJECT_NAMESPACES` and to the test list. I'd do it in this PR; I won't block on it.

## 5. Claim 4 (declining the allow-list) — I agree, and I was wrong to ask for it

I read `docs/embedding.md` myself. Lines 131-144: *"`TurnInitiator.kind` is `subject | service`; `subjectId` **remains opaque to OpenGeni**. Hosts must not encode or infer the kind from a subject-id prefix, **because the host owns that namespace**."* A `user:`/`dev` allow-list would (a) break every embedding host that mints opaque human subjects — and the new test's `acme-employee-4471` case is exactly that host — and (b) be OpenGeni inferring a principal kind from a host-owned prefix, which is the thing the contract forbids. **The author read the contract correctly and the decline is right.**

The deny-list replacement is acceptable: it is the same interpretation but in the fail-closed direction. A host *can* legitimately mint `worker:alice`; that human then gets a 422 whose message talks about API keys and services. Denial, not escalation — the right failure mode, with a confusing message. **Small ask:** `docs/embedding.md` now needs one line declaring `api_key`, `configured`, and `worker` (plus whatever §4 adds) reserved subject namespaces. Today that reservation is documented only in `integrations-design.md`, which is not where a host integrator looks.

## 6. Claim 3 (the 422) — verified, and the caller sweep is complete

Verified from source, not from the description: `DEFAULT_OAUTH_PROFILE.allowedOwnership = ["workspace", "personal"]` (`oauth-profiles.ts:117`), no curated Definition matches a built-in profile (`builtInOAuthProfileFor` matches on `mcpUrls` / `providerDomains`; the five Definitions are `www.googleapis.com` / `graph.microsoft.com`), so `soleAllowedOwnership → null → 422`. Gmail and hosted Slack (`allowedOwnership: ["personal"]`) still resolve, and a reconnect still short-circuits on `existingOwnership`. It over-corrects nothing: the endpoint refuses where it used to guess, and refusal cannot widen.

I redid the first-party sweep myself, whole repo, all file types, `--exclude-dir=node_modules`, on both `startApiIntegrationOAuth` and the raw path. Complete hit list: the route + its label, the contract, the SDK method + its README row, `apps/api/test/*`, `docs/capabilities.md` (prose), the e2e **mock** (`test/e2e/custom-api-control-center.browser.e2e.ts:500` intercepts the request, never issues one), and one runtime caller — `use-api-integration-accounts.tsx`, whose hook signature is `ownership = "workspace"` (line 229) applied client-side before the body is built (line 290). The Google Drive hook is the only other consumer and goes through that default. **No `examples/`, script, CLI, MCP tool, worker activity, scheduled task, or pack constructs this request.** The 422 is safe to ship.

## 7. Claim 5 (the three repaired fixtures) — the evidence holds, and I checked it independently

This was the "fixture edited until green" risk, and the author's evidence is real. `loadConnectionCredentialForBroker` with `allowSubjectOwned: true` sets `subjectPredicate = connectionExactSubject(input.subjectId)` (`packages/db/src/index.ts:11541-11543`) — it matches **only** a subject-owned row; a workspace-owned row is not a fallback, it is a miss. So the *"oauth callback records non-fatal verification failure…"* fixture, which asserts `loaded?.credential` at line 2484-2493, genuinely requires a personal row and would have failed under `workspace`. That is decisive proof the fixtures were personal all along (the hand-minted states omitted `ownership`, which the legacy decode reads as `"personal"`). Adding `ownership: "personal"` + the claim preserves the original semantics; no assertion was converted. The other two fixtures test token-endpoint redirect leakage and operator client-key normalisation and are ownership-agnostic.

---

## My own sabotage: 27 guards, 24 red, 3 green

Not the author's 16 — my own decomposition, one sub-term at a time, run over 14 suites (`connection-ownership`, `connections-routes`, `api-integration-provider-oauth`, `api-integrations-routes`, `google-drive-oauth-isolation`, `google-drive`, `atlassian-helpers`, `social-api`, `social-oauth`, `social-connector-flow`, `oauth-profiles`, `connection-authority-owner`, `connection-authorities-routes`, `app`). Baseline **213 pass / 0 fail**. Each row disables exactly one thing and reverts.

| # | sabotage | result |
| --- | --- | --- |
| G01 | core predicate → `true` | RED 9 |
| G02 | drop `contextIntegrity` | RED 1 |
| G03 | drop `authenticatedSubjectId === grant.subjectId` | RED 1 |
| G04 | drop `principalKind === "human_session"` | RED 5 |
| G05 | drop `!serviceInitiator` / `!serviceInitiatorContext` | RED 1 |
| **G06** | **drop `isPersonalConnectionOwnerSubject(grant.subjectId)` from the predicate** | **GREEN 0** |
| G07 | `RESERVED_MACHINE_SUBJECT_NAMESPACES` → `[]` | RED 2 |
| G08 | `personalOwnerVerifiedInState` accepts anything | RED 6 |
| G09 | `personalOwnerStateAccepted` → `true` | RED 6 |
| G10 | `assertConnectionOwnershipAllowedForPrincipal` → no-op | RED 3 |
| G11 | manual `POST /connections` fence | RED 1 |
| G12 | `google-drive/install` route fence | RED 1 |
| G13 | `atlassian/install` route fence | RED 1 |
| G14 | MCP start claim → hardcoded `true` | RED 1 |
| G15 | provider-oauth start claim → hardcoded `true` | RED 1 |
| G16 | social start personal fence | RED 1 |
| **G17** | **social start claim → hardcoded `true`** | **GREEN 0** |
| G18 | MCP start `assertConnectionOwnershipAllowedForPrincipal` | RED 1 |
| G19 | provider-oauth start `assertConnectionOwnershipAllowedForPrincipal` | RED 1 |
| G20 | provider-oauth 422-on-omission → `"personal"` | RED 1 |
| **G21** | **provider-oauth start `assertOwnershipAllowed`** | **GREEN 0 (inert by construction)** |
| G22 | MCP callback fence | RED 1 |
| G23 | provider-oauth callback fence | RED 1 |
| G24 | Google Drive callback fence | RED 1 |
| G25 | Atlassian callback fence | RED 1 |
| G26 | social callback fence | RED 1 |
| G27 | `defaultOwnershipFor` → always `"workspace"` | RED 5 |

**All four previously-unpinned points (provider-oauth callback, Drive callback, Atlassian callback, social start) are now pinned**, and the two callbacks that had *nothing* (MCP, social) are pinned too. The three greens:

- **G06** — the reserved-namespace term inside `isPersonalConnectionOwnerPrincipal` is unpinned. The enumeration test's `withGrant({ subjectId })` helper mutates only `grant.subjectId` and leaves `authenticatedSubjectId: "user:alice"`, so the **substitution** guard masks it and the namespace check is never exercised in the predicate. (`refuses a machine subject even under a forged human claim` has the same masking — only its second assertion pins anything.) One-line fix: set both fields in the loop. Low severity: the deny-list *is* pinned in the standalone function and on the callback path (G07/G09 red), and the only threat it addresses is a delegation-secret holder.
- **G17** — hardcoding the social start's claim to `true` reddens nothing, because the social flow's own fence (G16) still 422s a machine personal request, and the social callback ignores the claim for workspace ownership. It only matters via the §3 cross-flow reach. Worth one test.
- **G21** — the `assertOwnershipAllowed` newly added to provider-oauth start is currently unreachable: no curated Definition resolves to a personal-only profile. Green by construction, not a gap; it is the correct forward guard.

## Checks

All on `c899c65f`, clean worktree:

- `typecheck` — clean, all projects
- `lint` (`oxlint --deny-warnings`) — clean
- `format:check` — clean, 2765 files
- `check:docs-refs` — passed
- `check:public-hygiene` — passed (post-commit, clean tree)
- `check:migration-ordinals` — ok (310 migrations, base `origin/main` highest **0302**, so #1631 landing does not collide)

**Baseline, established independently on pristine `origin/main` `13c3af4f`:**

| | pass | skip | fail |
| --- | --- | --- | --- |
| `origin/main` `13c3af4f` | 1303 | 3 | 17 |
| head `c899c65f` | 1323 | 3 | 17 |

`diff` of the two **sorted failure-name sets**: **identical, 17 = 17.** All 17 are the pre-existing sandbox desktop/viewer-stream suites (`P1.4`, `P3.2`, `P4.2`). +20 tests, all new. (An earlier run on `adf288b3` showed an 18th, `(fail) (unnamed)` in `capability-discovery-mcp.test.ts` — `password authentication failed for user "opengeni_app"`. That was contention with another agent's concurrent `packages/db` run reprovisioning roles on the shared test Postgres, not this PR; it did not reproduce.)

### CI: the one red check is a base-vs-head artefact, not a defect here

`Deployment artifacts / Validate Helm chart` fails with:

```
single-node profile did not render Garage
```

That assertion exists **only on `main`** — `.github/workflows/ci.yml:1192`, added by #1639 (`feat(storage): default bundled S3 fixture to Garage`, merged as `13c3af4f`). It does **not** exist on this PR's head, which branched from `6e200c9`. GitHub evaluates the workflow from the base for `pull_request` events while this repo's checkout step pins `ref: github.event.pull_request.head.sha`, so main's new Garage assertion runs against the head's pre-#1639 chart, which still renders MinIO. The same job **passed** on the previous head (`adf288b3`, run 32259613579), and `c899c65f` changed only two test files and the changeset. Per the moving-main contract this is not a source defect and does not justify a rebase — but the merge authority should expect that check to stay red until the base-owned assertion and the head's chart line up.

---

## Could not verify

- **Production population of machine-owned personal rows.** No production DB. The survey SQL in the changeset is the right instrument; the number is still outstanding, and the PR body says so.
- **A real rolling deploy** with genuinely in-flight pre-claim states. Reasoned from the TTL and decode path plus the synthetic legacy-state tests, not simulated on two live versions.
- **Live `microsoft-*` OAuth.** The 422 is now executed for both `google-drive` and `microsoft-outlook-mail` Definitions in the new test, so the shared-profile reasoning is covered, but no real Microsoft round-trip.
- **The concurrency behaviour of the nonce insert under genuine parallel callbacks** — I verified it structurally (PK + `onConflictDoNothing` + `returning`), not by racing two requests.

## Follow-ups (none blocking)

1. Add `sandbox` and `scheduled_task` to `RESERVED_MACHINE_SUBJECT_NAMESPACES` and to the enumeration test, or soften the `integrations-design.md` sentence — today it is wrong (**§4**). I'd do this one in this PR.
2. Correct the PR body's binding table: for Drive/Atlassian the claim is **not** bound to ownership or provider identity, and cross-flow reach is real for a same-subject human (**§3**).
3. Pin G06 (one line: set `authenticatedSubjectId` alongside `subjectId` in the enumeration loop) and G17.
4. One line in `docs/embedding.md` reserving the machine subject namespaces, since that is the doc that promises hosts the namespace is theirs (**§5**).
5. Give every OAuth state payload a purpose discriminator and check it, as `social-oauth.ts` already does (**§3**).
6. `connection_conflict` → `state_invalid` for the provider-oauth legacy-state refusal.
7. `PATCH /connections/:id { subjectId }` 500 — already filed as #1640. Confirmed again here: migration `0256`'s trigger (`connection owner authority is immutable`, `23514`) holds and the row stays workspace-owned, so it is not a bypass.

Would I stake production access control on this? Yes. Every path that can mint a personal owner — manual create, MCP OAuth, Integration-Definition OAuth, social OAuth, Google Drive, Atlassian — is fenced at start on the live principal and at callback on a signature-covered claim, with `contextIntegrity` and unknown provenance failing closed, and nothing in the change widens any principal's access.

benyhh (commented):
REQUEST CHANGES exact head 4927a1c0b9078490daa9bc557d30117dfdc4c5ba

The authority implementation itself is fail-closed across manual, MCP, Integration Definition, social, Google Drive, and Atlassian creation/callback paths. The signed claim is exact-boolean and signature-covered, Drive/Atlassian have flow discriminators, reconnect ownership is preserved, the current-main merge tree is clean, and the focused authority/Fiken run is 132 pass / 0 fail (including queued revoke/reconnect with one target call). Two release-contract artifacts still need correction before this exact head is approvable:

1. `.changeset/no-silent-personal-connection-ownership.md` explicitly calls the new 422 behavior breaking for direct API/SDK callers, but declares a `minor` bump for the public `@opengeni/api-router` package, currently 1.0.0. That release metadata must be `major` (or the PR must cite a concrete repository policy that permits a known breaking HTTP-surface change in a minor release).
2. The changeset and `docs/integrations-design.md` say the new owner predicate matches `requireConnectionAuthorityOwner` exactly and only the status differs. It does not: the new helper deliberately adds the broader reserved-machine-namespace check, while the sibling helper only has the `api_key:` subject check. Describe the shared core checks (`contextIntegrity`, exact subject, `human_session`, and no service provenance) and the intentional stricter defense-in-depth here instead of claiming predicate identity. While correcting that paragraph, avoid saying every callback fence itself "422s": the callback surfaces intentionally normalize the refusal differently (for example provider OAuth redirects with `connection_conflict`, social with `not_authorized`).

No source-authority widening or current-main conflict found; this request is limited to making the published upgrade contract match the code that will ship.

benyhh (commented):
APPROVE exact head e070f350b7cc584d88e9923865fcb349abf90ad6

The two release-contract blockers from review 4980033488 are fully corrected: `@opengeni/api-router` now carries a major changeset for the documented breaking 422, and the changeset, design doc, source comment, and PR body now distinguish the shared caller-integrity checks from the deliberately stricter reserved-machine-namespace defence. Callback refusals are accurately described as flow-specific bounded projections rather than uniformly HTTP 422.

I re-reviewed the full authority diff. Manual create, MCP OAuth, Integration Definition OAuth, social OAuth, Google Drive, and Atlassian all fail closed at start for a non-human personal owner and at callback through an exact-boolean HMAC-covered claim; Drive and Atlassian also enforce exact flow discriminators. Ambiguous Integration Definition ownership is refused rather than widened, reconnect ownership is preserved, existing-row behavior/remediation is accurately documented, and no new source-authority issue is introduced by the repair.

Verification on this exact head: the focused seven-file authority/integration suite is 132 pass / 0 fail. Against current `origin/main` d634e4447, the merge tree is conflict-free; the only touched-path overlap is AGENTS.md in separate hunks, with no material semantic incompatibility. Live CI currently has 16 passed, 16 pending, 0 failed (plus expected skips); this approval is code/release-contract approval, and merge still requires the remaining checks to finish green.

benyhh (commented):
APPROVE exact head 5d4b29166a233a1e6a86e56e5f9092d4f2b34fe9

This head is the clean two-parent merge of previously approved e070f350b7cc584d88e9923865fcb349abf90ad6 with current main d634e4447bc2b64e312838e01f3d8a6100c50b0e. Its tree exactly matches a fresh `git merge-tree --write-tree d634e4447 e070f350b` result (`7dd1431bd129b03c62e4c2082ab18cf53360f1c7`), so there is no manual-resolution drift. Every PR-owned file other than the independently touched AGENTS.md is byte-identical to e070f350; the merged AGENTS.md retains both the reviewed connection-ownership invariant and main’s separate delivery/SuperGrok updates.

The cumulative authority and release-contract corrections remain intact: major `@opengeni/api-router` changeset, accurate shared-vs-stricter predicate language, flow-specific callback error projections, exact-boolean signed owner claims, and Drive/Atlassian flow discriminators. The coordinator reports the focused suite after this merge at 132 pass / 0 fail and format/diff checks passing; that matches the 132/132 focused suite I independently ran on e070f350.

GitHub currently reports no checks registered yet for this new exact head. This is exact-head code/release-contract approval only; merge must wait for the new head’s required CI to appear and finish green.

PR #1638, opened 2026-08-19
fix: resolve personal-workspace authority in the connection grant layer

benyhh:
## Summary

- make personal-connection authority recognize a managed human's owner-only personal-workspace grant
- deny delegated bearer grants access to personal connection delegation in both personal and shared workspaces
- preserve exact account/workspace pairing and failure-safe subject-GUC restoration around authority probes
- converge on the canonical workspace-authority resolver and public helper name landed by #1641
- integrate cleanly with the connection-ownership protections landed by #1637

## Authority boundary

Managed personal workspaces intentionally have no durable `workspace_memberships` row. The connection layer now asks the canonical named-subject authority oracle, which recognizes either an ordinary live workspace membership or the exact active organization membership's `personal_workspace_id` pointer.

The oracle is not caller authorization. Every call site must supply either the authenticated principal's already-authorized subject or a frozen delegation owner reached through an already-authorized workspace grant. Delegated bearer metadata is denied before this path, and the positive Better Auth provenance gate from #1641 remains intact on the session surface.

## Integrated prerequisite resolution

This head includes current `main` at `2027bc3ef`:

- #1641's extracted `subjectHasLiveWorkspaceAuthorityInScope`, public `namedSubjectHasLiveWorkspaceAuthority`, positive managed-cookie provenance, and session-surface behavior are retained.
- #1638's delegated-source denial, connection-authority call sites, account/workspace pairing, subject-GUC restoration, and tests are retained.
- #1637's connection-ownership security behavior is unchanged.
- stale documentation claiming session list, pin, and drafts remain broken has been removed.

## Breaking changes

The unique `@opengeni/core` public contract changes remain a major changeset: `PersonalConnectionDelegationSource`'s subject variant now requires `accountId`, and delegated personal-connection behavior is narrowed. The `@opengeni/db` helper rename/major release is no longer duplicated here because it landed in #1641.

## Verification

- connection, Google Drive, and contract authority suites: 54 passed
- session-surface and tenancy suites: 26 passed
- workspace-isolation integration: 1 passed
- impacted TypeScript projects (`contracts`, `db`, `core`, `worker`) pass
- formatting, lint, workspace/billing static guard, changeset status, and diff checks pass
- `origin/main` is an ancestor of this exact head

Two whole-repository typecheck attempts were terminated by host memory saturation; all directly impacted projects passed independently, and the exact pushed head is subject to the normal full CI gate before merge.

No migration or SQL change is introduced by this PR.


benyhh (commented):
## Verdict: **REQUEST CHANGES** (posted as a comment — `--approve` is unavailable on own-account authorship)

I would not stake production access control on this **as written**. Not because the fix is wrong — the bug is real, I reproduced it, and the functional change is correct and well-pinned. But the PR's central security claim is **false**, and this PR writes that false claim into `docs/organization-tenancy.md` and into a doc comment that instructs *every future seam* to adopt this function. That is how the next authority hole gets written.

Nothing here is a live, exploitable leak today. Everything below is either a disproven claim, a contract inconsistency, or a latent trap. The delta to approval is small.

---

## 1. BLOCKING — "owner-only by construction" is not true. The guard is self-defeating.

The PR body, the commit message, the `packages/db/src/index.ts` doc comment, and the new `docs/organization-tenancy.md` section all assert:

> the pointer branch reads `list_self_organization_memberships`, a SECURITY DEFINER seam that raises `42501` unless the requested subject is exactly the transaction's own subject GUC ... so no account/organization administrator, API key, service, or delegated bearer can reach a different human's personal workspace through it, **whatever subject a caller supplies**.

The SQL guard is (`0263_organization_membership_lifecycle.sql:439-444`):

```sql
IF p_subject_id IS NULL
  OR p_subject_id IS DISTINCT FROM opengeni_private.current_subject_id()
  OR p_subject_id NOT LIKE 'user:%'
THEN RAISE EXCEPTION ... ERRCODE = '42501';
```

But `subjectHasLiveWorkspaceAuthority` **sets that GUC from its own argument, one statement earlier** (`packages/db/src/index.ts:17502`):

```ts
await setSubjectRlsContext(scopedDb, input.subjectId);   // set_config('opengeni.subject_id', <argument>, true)
...
sql`select list_self_organization_memberships(${input.subjectId}) as result`
```

`setSubjectRlsContext` (`packages/db/src/database.ts:856`) is a plain `set_config` available to the ordinary `opengeni_app` role. The predicate `argument IS DISTINCT FROM current_subject_id()` is therefore **always false** for any subject the caller names. The function checks that the caller filled in the same value twice. It proves nothing about who the caller *is*.

**Proven, not argued.** Against the real shared test Postgres, as `opengeni_app` with RLS engaged, from a process that never authenticated as anyone:

```
PROBE1 memberships for victim: [{"id":"977285b9-…","role":"owner","status":"active",
  "subjectId":"user:victim-b19560e4-…","organizationId":"fbf697e1-…",
  "personalWorkspaceId":"7addb0a2-ea8d-48fc-970b-54c4aad48a5b",
  "authorizationRevision":1,"personalRetentionUntil":null}]
PROBE1 leaked personalWorkspaceId: [ "7addb0a2-ea8d-48fc-970b-54c4aad48a5b" ]
```

I read a victim's complete organization-membership row — role, status, authorization revision, and the private `personalWorkspaceId` — by doing nothing but naming them:

```ts
await tx`select set_config('opengeni.subject_id', ${victim.subjectId}, true)`;
await tx`select list_self_organization_memberships(${victim.subjectId}) as result`;
```

And `PROBE2`: `subjectHasLiveWorkspaceAuthority(db, {accountId, workspaceId: victim.personalWorkspaceId, subjectId: victim.subjectId})` → **`true`**, from a caller with no session, no cookie, no grant.

So `subjectHasLiveWorkspaceAuthority` is an **arbitrary-subject oracle**, not an authorization. Its safety is *entirely* caller discipline — precisely the property the PR says it does not depend on.

**To be fair to the change:** all five new call sites do pass a subject they are entitled to ask about (the grant's own subject, or the frozen `ownerSubjectId` of a delegation the caller already holds), and I could not construct a cross-user read through any of them. This is a wrong *rationale*, not a live hole.

But the rationale is the thing being institutionalised. The doc comment says "Every seam that needs 'does this subject still hold workspace authority here' must call this function," and the new docs section repeats "owner-only *by construction*, not by caller discipline." A future author will read that and pass an attacker-influenced subject, because the docs promise the function is safe against exactly that.

**Asks:**
- Correct the claim in all four places (PR body, commit message, `index.ts:17478-17489`, `docs/organization-tenancy.md`). State plainly: *this function answers for any subject named; the caller must prove entitlement to ask.* The existing hedge ("Callers must still pass only a subject they are entitled to ask about") contradicts the sentence above it — keep the hedge, drop the "by construction" guarantee.
- Better: make it structural for real. Split into `subjectHasLiveWorkspaceAuthority(db, {…, subjectId})` (the oracle, explicitly named as such) and a `callerHasLiveWorkspaceAuthority` that derives the subject from an already-established authenticated context instead of minting the GUC from its own argument.

Related (out of scope, worth an issue): the same self-set-GUC pattern makes `list_self_organization_memberships` a general read primitive for any application-role connection. `ensureManagedAccessForUserWithOrganizationMemberships` uses it the same way (`index.ts:1481-1484`). The SECURITY DEFINER wrapper is not buying the isolation its name implies anywhere.

---

## 2. BLOCKING (small fix) — a delegated bearer *does* inherit the widening, at the seam you changed

Your stated reason for leaving `getWorkspaceGrant` alone:

> it is also the fallback inside `accessGrantAuthorization`, where a delegated bearer carrying `principalKind: "human_session"` and a `user:<id>` subject ... would inherit any widening.

That reasoning is sound, and I agree with the non-change (see §3). But the widening is now present one layer inward, at a seam you *did* change.

`personalConnectionDelegationSourceForGrant` (`personal-connection-delegations.ts:99-107`) excludes only `agent_attempt`, `service`, and `serviceInitiator`. It does **not** exclude `grant.metadata?.delegated === true`. And `delegatedAccessContext` (`packages/core/src/access/index.ts:339-389`) builds `workspaceGrants` **entirely from the signed token payload with no database validation**:

```ts
workspaceGrants: [{ workspaceId: payload.workspaceId, accountId: payload.accountId,
                    subjectId: payload.subjectId, principalKind: payload.principalKind,
                    metadata: { delegated: true, … } }]
```

`DelegatedAccessPrincipalKind` permits `human_session` (`contracts/src/index.ts:2056`), and the refinements at :2168-2176 only forbid it from carrying machine-authority claims — nothing binds `subjectId`/`workspaceId` to real rows.

So a `human_session` delegated token minted for `user:<victim>` + the victim's personal workspace:
- **before:** `getWorkspaceGrant` → `null` → zero personal connections frozen;
- **after:** `subjectHasLiveWorkspaceAuthority` pointer branch → `true` → **the victim's personal connections are frozen into the session.**

`CLAUDE.md` is explicit: *"Bearer/delegated principals, API keys, and account or organization administrators receive no personal-workspace access through that exception."* This PR grants exactly that to the delegated-bearer case while documenting the opposite.

**Weighing it honestly:** this needs the first-party delegation secret, agent-attempt tokens are already filtered to `{kind:"none"}`, and a secret-holder could already impersonate `user:<victim>` in any *shared* workspace the victim belongs to. So it is not a new trust-boundary crossing — it is the same trust extended into the victim's private space, against an explicit written prohibition. Medium, and a one-line fix:

```ts
if (grant.principalKind === "agent_attempt" || grant.principalKind === "service" ||
    grant.serviceInitiator || grant.metadata?.delegated === true) {
  return { kind: "none" };
}
```

(or gate the pointer branch on a canonical managed-cookie grant). Either way, please add a test.

---

## 3. The deliberate non-change to `getWorkspaceGrant` — you are right, keep it

I tried to argue the other side and could not. Widening `getWorkspaceGrant` would flow straight into `accessGrantAuthorization`'s fallback for API-key and delegated principals, which is the forbidden widening. Two implementations of one rule *is* a smell, but here they answer genuinely different questions: `getWorkspaceGrant` returns a **grant with permissions** for a *request principal*; `subjectHasLiveWorkspaceAuthority` returns a **boolean** about a *named subject*. That is not divergence, it is two functions.

The residual risk is naming, and §1 is the fix: as long as one is called "THE canonical live-workspace-authority resolver" and the other is described as "not an authority answer," someone will swap them at an authorization site. Make the oracle's name say it is an oracle.

---

## 4. MEDIUM (latent) — the seam clobbers the caller's subject GUC and does not restore it

`withRlsContext` (`database.ts:427-455`) saves and restores **only** `opengeni.account_id` and `opengeni.workspace_id` (`restoreRlsContextSettings`, :405). `subjectHasLiveWorkspaceAuthority` sets `opengeni.subject_id` inside that scope, and — as the function's own comment notes — "SET LOCAL survives successful savepoint release."

Proven (`PROBE3`), inside a caller's transaction:

```
before:  user:caller-c2b1f440-…
after:   user:other-249ea3b0-…      <-- subject GUC CLOBBERED
afterAccount: cd10fad0-…            <-- account_id correctly restored
```

The subject the rest of that transaction runs as becomes the **probed** subject — at three of the five call sites, a frozen `ownerSubjectId` read out of stored data.

**Currently inert.** I traced every production caller: all five new sites, plus the original 0281 `apps/api/src/sandbox/viewer.ts:747` mint recheck, receive top-level pool handles (`deps.db` from `app.ts:274`, `services().db` from `activity-services.ts:38`), so `isTransactionHandle` is false, the function opens its own transaction, and the GUC dies at COMMIT. That is presumably why 0281 never surfaced this.

But the parameters are typed plain `Database`, nothing rejects a tx handle, and this PR's docs now direct every future authority probe here. `apps/api/src/scheduled-task-deletion.ts:112` already injects a `tx` into an `ApiRouteDeps`-shaped object today (it does not reach this seam — but it shows the shape exists). Please either add `opengeni.subject_id` to `restoreRlsContextSettings`, or have `subjectHasLiveWorkspaceAuthority` restore the prior subject itself. Cheap, and it closes the trap permanently.

---

## 5. MEDIUM — the new tests only exercise a *cross-organization* intruder

Both negative tests build their "intruder" with a second `provisionManagedHuman()` / `seedOwnerWithDriveConnection()`, and each of those calls `ensureManagedAccessForUser` with a fresh user id — which mints a **fresh managed account**. So `intruder.accountId !== owner.accountId`, and the deny is reached trivially: the intruder has no organization membership in the owner's organization at all.

The case `CLAUDE.md` is actually about — a **same-organization** non-owner: a co-member, an organization admin, an account admin — is never tested. I read the path and it does hold (a same-org intruder's own membership resolves, but its `personalWorkspaceId` is their own workspace, and 0219 forbids a `workspace_memberships` row on a personal workspace, so both branches are false). But that is the load-bearing property of the whole feature and it has no failing test.

Please add: two humans in one organization via the 0263 invitation lifecycle, then assert `subjectHasLiveWorkspaceAuthority(owner.personalWorkspaceId, intruder.subjectId) === false` for intruder roles `member`, `admin`, and `owner`.

**Sabotage results** (`SABOTAGE_ALWAYS_TRUE=1`, authority check forced `true`) — the guard is pinned, but unevenly:

```
(fail) another human's subject never freezes a delegation in someone else's personal workspace
(fail) a different managed human gets no authority over someone else's personal workspace
 37 pass  2 fail
```

Note what did **not** go red: `another human never resolves a publication target in someone else's personal workspace`. That test still passes with the membership check disabled, because the deny is enforced downstream by connection ownership, not by the seam under test. The `getMembership` wiring in `resolveGoogleDrivePublicationTarget` has mock-level coverage (`getMembership: async () => false`) but no real-resolver test that fails when it is bypassed.

---

## 6. MINOR — `accountId` and `workspaceId` come from different objects in one place

`scheduled-tasks.ts:718-723` passes `workspaceId: input.existing.workspaceId` alongside `source: …ForGrant(input.grant)` → `accountId: input.grant.accountId`. Everywhere else the two are consistent by construction (`grant.workspaceId === workspaceId`).

The old `getWorkspaceGrant(db, subjectId, workspaceId)` ignored the account entirely, so it was immune. The new call scopes RLS by `grant.accountId` and then reads memberships for `existing.workspaceId`; if those ever diverge, the membership row is filtered out by RLS and **every personal connection is silently dropped** from the updated task. Deny-side, not a leak, and I could not make them diverge — but deriving the account from the workspace row (`rlsContextForWorkspace`) instead of from the grant would make it structurally impossible.

---

## What I verified and what held

**The bug is real.** I restored the old bare-membership port behind an env flag and reproduced your "before" exactly, against real Postgres:

```
(fail) the owner-only live-authority seam resolves the personal workspace        Received: false
(fail) a personal social connection freezes a delegation …                       Expected length: 1  Received length: 0
(fail) resolves the owner's frozen publication target through the real ports      Received: null
```

**Stricter, as claimed — confirmed.** The new resolver adds `selfOrganizationMembership.status === "active"` filtering on top of the membership row, which the bare join never did. A suspended/revoked organization membership is now correctly false.

**Phase D respected.** The only authority fact read is `organization_memberships.personal_workspace_id`. No `created_by`, connection attribution, default workspace, resource name, or current access is consulted anywhere on the new path.

**Non-`user:` short-circuit is safe.** API keys (`api_key:*`), configured (`configured:*`), and service subjects cannot own an organization membership and now return the plain membership answer — no behaviour change, no `42501`.

**No over-broad SQL disjunct exists.** I audited every `organization_memberships` reference across all 31 migrations that touch it (~30 authority predicates in 0241/0249/0252/0253/0254/0256/0258/0262/0264/0275/0279/0280/0297). Every one binds `organization_memberships.subject_id` to a specific validated subject, filters `status='active'`, and binds the organization. No RLS policy or SECURITY DEFINER function uses `personal_workspace_id` in an access predicate without a subject binding. Two notes: `0253:1347-1359` deliberately drops 0252's extra `personal_workspace_id = origin_workspace_id` requirement (intentional common-personal-resource widening, still owner-fenced — deserves an explicit nod somewhere); and `0256:538-551` is only owner-bound via the outer query's `subject_id`/`account_id` pins, which reads sloppily and is worth tightening.

**Unfixed seams — I spot-checked five, chosen myself. All deny; none leak:**

| Seam | Behaviour for a personal workspace |
|---|---|
| `listSessionsForSubject` (`index.ts:27375-27394`) | `if (!membership && subjectId.startsWith("user:")) throw SessionListAccessError` → **denies the owner**. |
| `setSessionPin` (`index.ts:27821-27830`) | `if (!membership) throw SessionPinAccessError` — unconditional → **denies everyone**. |
| `listWorkspaceMembers` (`index.ts:2111-2123`) | Plain enumeration of membership rows; zero rows → empty roster. Not an authorization decision. |
| `withCodexAppsRequestAuthorization` (`index.ts:18682-18697`) | `canManageCodexApps(undefined)` → `Array.isArray(undefined)` is false → **throws**. Fails closed. |
| `saveNewSessionDraftInTransaction` (`new-session-drafts.ts:113-132`) | `if (!membership) throw NewSessionDraftAccessError` → **denies the owner**. |

Plus `fork_session_content` (`0289:238-246`) independently confirmed as deny-only. Your characterisation — broken feature, not authority hole — holds for everything I checked.

One structural note for the follow-up: `listSessionsForSubject` and `saveNewSessionDraftInTransaction` both *skip* the membership fence for non-`user:` principals ("API-key and delegated service subjects have no workspace-membership row"). Today the route-level grant is the only thing keeping an API key out of a personal workspace, and these seams provide no second fence. Fine now; please do not widen the route-level grant without revisiting them.

**Independence from #1631 / #1637.** Confirmed independent.
- **#1637** (`codex/no-silent-personal-ownership`): **zero file overlap**. It owns exactly the OAuth-callback cluster you deliberately left whole — consistent with your note.
- **#1631** (`codex/personal-workspace-session-owner`): touches `packages/db/src/index.ts` only at :17826 (a comment string in `TENANCY_PARITY_UNVERIFIABLE`), far from your :17457-17530. **No code conflict.** Expect a textual conflict in `docs/organization-tenancy.md` — you both append to the Slice B section. One semantic note: #1631 repairs `guard_session_authority_write` (0225) *with* the personal-workspace pointer via migration 0302, which makes your new doc bullet listing 0225 among "SQL seams without the personal-workspace disjunct" stale on merge. Whichever lands second should fix that line.

**Verification — all reproduced on my machine, none taken on faith:**

| Check | Result |
|---|---|
| `typecheck` | **31/31 projects clean**, exit 0 |
| `lint` (`oxlint --deny-warnings`) | clean |
| `format:check` | clean, 2765 files |
| `check:docs-refs` | passed |
| `check:migration-ordinals` | ok — 310 migrations, base `origin/main` highest 0301 |
| `check:public-hygiene` | passed |
| `packages/core/test` + `apps/worker/test` | **1525 pass / 0 fail** (you said 1526 — off by one, immaterial) |
| `packages/db` tenancy 0218/0219/0263/0298 + session-pins 0054 | **46 pass / 0 fail** |
| `apps/api/test` on this branch | **1303 pass / 17 fail** — matches your number exactly |
| `apps/api/test` on `origin/main` @ `6f61d6ee` | **1303 pass / 17 fail**, and `diff` of the sorted failing-test names against this branch is **empty — identical failure sets** |

Your baseline claim checks out, and I did not take it on faith. All 17 failures come from one file, `apps/api/test/sandbox-desktop-stream.test.ts`, which passes **17/17 in isolation** on this branch. Genuine full-suite resource contention, present identically on `main`.

Worth saying out loud, though: that file is the **0281 stream-token mint suite** — the closest existing regression coverage to the function you are generalising, including a test literally named *"0281 — a managed PERSONAL-workspace human mints (no membership row by design); suspension degrades."* It is not this PR's fault that it is red in the full run, but it does mean the single most relevant guard on this change is invisible in CI's full-suite signal. Fixing that contention is worth its own issue.

Also confirmed the new DB tests really run rather than no-op through their `if (!client) return` guards — 11 `expect()` calls on a live `opengeni_app` connection with FORCE RLS engaged.

---

## Summary

The bug is real, the diagnosis is right, the fix is the right fix, the blast-radius survey is honest and — where I checked it — accurate. That is most of the work and it is good work.

What blocks approval is that the *justification* is wrong in a way that outlives this PR. The SECURITY DEFINER seam does not constrain the caller; it constrains the caller to be self-consistent. Shipping that sentence into `docs/organization-tenancy.md` as the canonical rule, next to an instruction that every authority probe should adopt this function, is the part I would not stake production access control on.

Fix §1 (correct the claim, ideally rename the oracle), §2 (one-line delegated filter + test), §4 (restore the subject GUC), and §5 (a same-organization non-owner test), and I would be glad to approve.

benyhh:
## Review addressed — pushed as `9731ab62`

You were right on every count, and I reproduced all three probes myself against real Postgres as `opengeni_app` with RLS engaged before touching anything. `PROBE1` returned the victim's full membership row including `personalWorkspaceId`; `PROBE2` returned `true` from a caller with no session; `PROBE3` showed `user:caller-…` → `user:other-…`. The rationale I shipped was wrong, and thank you for not letting it land as canon.

---

### §1 — the false claim, corrected and renamed

The guard compares its argument to a GUC that the function itself sets from that same argument, through an ordinary `set_config` any `opengeni_app` connection may issue. It constrains the caller to be self-consistent, not to be anyone.

**Renamed** `subjectHasLiveWorkspaceAuthority` → **`namedSubjectHasLiveWorkspaceAuthority`**, so the name states what it answers about. Its doc comment now opens:

> ORACLE, NOT AN AUTHORIZATION. Answers "does the subject you *named* hold live workspace authority here" for **any** subject the caller cares to name. It does not, and cannot, establish who the caller is.

…then explains the self-set-GUC mechanism, cites the empirical result, and names the two entitled subject shapes. `docs/organization-tenancy.md` carries the same correction, keeps the "callers must prove entitlement locally" hedge as the *only* claim, and states the false version explicitly so the next reader learns the shape rather than rediscovering it. It also generalises your related finding: `*_self_*` definer routines here are not caller checks — including `ensureManagedAccessForUserWithOrganizationMemberships`, which uses the same pattern.

**On `callerHasLiveWorkspaceAuthority` (your "better" option):** feasible, but I recommend it as a follow-up and documented the reasoning rather than half-doing it. Three of the five call sites legitimately ask about a **non-caller** subject (a frozen delegation `ownerSubjectId`), so a caller-derived variant is an *additional* API for the caller-subject sites, not a replacement — shipping it now would leave two functions and still leave the oracle reachable. The fix that actually removes the hazard is at the SQL layer: have `list_self_organization_memberships` read the GUC instead of accepting the subject as a parameter. That is a migration touching every caller of a shipped definer routine, and it does not belong inside a bug fix.

### §2 — delegated-bearer hole closed

Took the one-line filter. `personalConnectionDelegationSourceForGrant` now collapses `grant.metadata?.delegated === true` to `{kind:"none"}`, with a comment recording that this was moot before the pointer existed (`getWorkspaceGrant` returned `null` for a personal workspace regardless), which is exactly why the filter belongs with the pointer.

Two tests: a unit test that a `human_session` **delegated** grant yields `{kind:"none"}` while the canonical managed-cookie grant still yields a subject source; and a DB test that forges precisely what `delegatedAccessContext` builds for a victim's personal workspace, asserts it freezes nothing, then flips only `metadata.delegated` off and asserts it freezes 1 — so the test proves the *bearer* is being denied, not the workspace.

### §4 — subject GUC restored

`namedSubjectHasLiveWorkspaceAuthority` now reads `opengeni.subject_id` before probing and restores it in a `finally`. I fixed it in the seam rather than in `restoreRlsContextSettings`: `withRlsContext` is used everywhere, its comment deliberately scopes restoration to the tenant pair, and widening a shared primitive is not something I want riding along in a security fix. The seam is the thing that mints the subject GUC from an argument, so the seam is what should clean up.

Two permanent regression tests (not scratch probes): one asserts both subject and account survive a probe inside a caller transaction, one asserts the probe still answers correctly when handed a transaction handle — so the restore cannot be "fixed" by breaking the tx path.

### §5 — same-organization denial matrix

You were right that a fresh account per human proved only cross-tenant denial. Built the real case through the 0263 invitation lifecycle. For each of **member, admin, owner**, two tests:

- the co-member is denied the owner's personal workspace, with `accountId` / `status: active` / `role` asserted first so the test cannot pass by accidentally landing in a different organization — and, in the same test, the co-member **does** resolve `true` for their own personal workspace, so the deny is specific rather than blanket;
- the co-member freezes **no** delegation in the owner's personal workspace even with a connection planted there under their own subject.

Renamed the old test to `a different-organization managed human …` and added `expect(intruder.accountId).not.toBe(owner.accountId)` so it is honest about which case it covers.

**Drive coverage.** Your catch that the Drive intruder test survived sabotage was correct — it was denied downstream by connection ownership. Added an isolated test whose planted connection genuinely **belongs** to the intruder and sits in the owner's personal workspace, so `getConnection` succeeds and the probe is the only fence. It carries a control assertion (probe forced `true` → resolves) proving every downstream check passes, then asserts the real resolver denies.

**Fresh sabotage results.** Forcing the probe `true` at both layers (`ownerStillBelongsToWorkspace` and the db oracle):

```
 10 pass  9 fail
```

up from your `37 pass / 2 fail`. The nine: 3 same-org authority + 3 same-org freeze + cross-org authority + cross-org freeze + **the Drive probe-isolation test**.

### §6 — account/workspace pair

Fixed at the flagged site: `workspaceId` now comes from `input.grant`, the same object `personalConnectionDelegationSourceForGrant` reads `accountId` off, so the two cannot drift. The task is loaded via `getScheduledTask(db, grant.workspaceId, …)`, so this is the same workspace. I skipped the `rlsContextForWorkspace` derivation — it throws on an unknown workspace where the old path returned a deny, and I did not want to introduce a new failure mode for a deny-side issue neither of us could trigger. Both traps are now recorded in the docs section.

### Coordination

- **0225 bullet updated** for #1631's migration 0302 — it now names only `fork_session_content` (0289) and 0234, and says explicitly that 0225's `guard_session_authority_write` is repaired by 0302 and should not be re-added.
- Expect the `docs/organization-tenancy.md` textual conflict you predicted; whichever lands second resolves it.
- Untouched: the session-facing seams (`listSessionsForSubject`, `setSessionPin`, `saveNewSessionDraftInTransaction`) and the OAuth-callback cluster.

### Verification

`typecheck` 31/31 clean · `lint` · `format:check` · `check:docs-refs` · `check:public-hygiene` · `check:migration-ordinals` (310, base highest 0301).

`packages/core/test` + `apps/worker/test` → **1536 pass / 0 fail**. Targeted suites → 50 pass / 0 fail.

One note in your favour, found while verifying: `session-pins.test.ts` + `migration-0263` run together produced 2 failures on the "personal-state fence" race in one run, then passed on repeat runs and passes 28/28 alone — the same contention class as the `apps/api` 17, and the pre-change baseline for that combo is also green. Not caused by this change, but recorded rather than quietly re-rolled.

Both new DB test files use `acquireSharedTestDatabase` (the `opengeni_app` role, FORCE RLS engaged) with explicit `180_000` timeouts including `beforeAll`/`afterAll`. Format ran last. No migration; still independent of #1631 and #1637.

benyhh (commented):
## Verdict: **REQUEST CHANGES** — narrow. The security substance now holds; three artifacts do not.

Posted as a comment (`--approve` is unavailable on own-account authorship).

I re-reviewed `9731ab62` from scratch in a clean worktree off `origin/main` @ `6f61d6ee`, re-ran every check and suite, and re-ran the sabotage myself rather than trusting the numbers. **The code is now something I would stake production access control on.** All four of the previous round's blocking asks are genuinely fixed, and I verified each one by breaking it and watching a test go red.

What I am withholding approval for is not code. It is three artifacts, all cheap, and the first of them is the exact failure mode this round was called to prevent.

---

# Blocking

## B1 — The PR body still contains the disproven claim, and `main` squash-merges.

`origin/main` is a chain of single-parent `... (#NNNN)` commits — this repo squash-merges, so **the PR description becomes the permanent commit message on `main`.** The description still says, verbatim:

> **It is owner-only structurally, not by caller discipline.** The pointer branch reads `list_self_organization_memberships`, a SECURITY DEFINER seam (`0263`, lines 431-458) that raises `42501` unless the requested subject is *exactly* the transaction's own subject GUC **and** matches `user:%`. The only fact it can ever return is "subject X owns personal workspace W", for the exact X named. Whatever subject a caller supplies, no admin, API key, service, or delegated bearer can reach a *different* human's personal workspace.

Your own fix commit message calls this false, in as many words. Last round's ask #1 listed four places to correct it: **PR body**, commit message, `index.ts`, `docs/organization-tenancy.md`. Three of four are done and done well. The one left undone is the one that lands in `main`'s history and is the first thing anyone reading `git log` on this change will see.

The whole argument last round was "the claim outlives the PR." Leaving it in the squash message means it literally does, in the most durable place available. Rewrite the "How the exception stays owner-only" section of the description to match what `docs/organization-tenancy.md` now says.

## B2 — No changeset, for a breaking rename of a published package's public API.

`@opengeni/db` is publicly published — `packages/db/package.json` has no `private`, and carries `"publishConfig": {"access": "public", "provenance": true}` at v2.0.0. This PR renames the exported `subjectHasLiveWorkspaceAuthority` → `namedSubjectHasLiveWorkspaceAuthority` with **no back-compat alias**, and changes runtime behaviour (B3). There is no `.changeset/` entry in the diff.

Repo convention is unambiguous — every recent commit touching `packages/db/src/index.ts` carries one (`c19fad8b` → `session-snapshot-and-pin-visibility.md`, `1c78ed0f` → `composer-authority.md`, `51123b41` → `late-bind-remembered-tools.md`). CI will not catch this: `ci.yml`'s changeset step only rejects changesets that target *only* ignored packages; it does not require one. So this ships in whatever bump an unrelated changeset happens to produce, and an external consumer gets a removed export with no release note.

The rename is right — please just declare it. Either a changeset with the appropriate bump, or a changeset *plus* a deprecated `export { namedSubjectHasLiveWorkspaceAuthority as subjectHasLiveWorkspaceAuthority }` alias.

## B3 — The delegated filter is correct, but it is broader than the PR says, and the widening is undisclosed and untested.

The filter itself is sound and I could not defeat it (see V2 below). But its blast radius is not the personal workspace — it is **every** workspace.

`personalConnectionDelegationSourceForGrant` now returns `{kind:"none"}` for any grant with `metadata.delegated === true`. That collapses the source before the workspace is ever consulted, so a delegated `human_session` grant now freezes **zero** personal connections **anywhere**, including ordinary shared workspaces where it previously worked through a real `workspace_memberships` row.

That is a documented surface. `docs/embedding.md:227`: *"The host still mints its ordinary user-facing delegated token."* An embedding host minting `ogd_` tokens for its real users loses personal X/Reddit/Atlassian/Drive connection delegation across the board, silently.

I think the change is **defensible** — a delegated grant's `subjectId` is unvalidated token payload, so trusting it to select someone's personal connections in a shared workspace is the same defect one room over, and closing it entirely is the better call. But right now:

- the PR body, the code comment, and `docs/organization-tenancy.md` all describe this as the *personal-workspace* fix;
- nothing in `docs/embedding.md` mentions it;
- no test covers the shared-workspace delegated case. The new test's control (`metadata: {}` → length 1) proves the filter is subject-specific, not that the shared-workspace lane was intentionally dropped. I grepped: no test anywhere pairs `delegated: true` with `freezePersonalConnectionDelegations`. That is why 1536/1536 stayed green — the regression is invisible, not absent.

Please either say explicitly that this is intended (docs/embedding.md note + changeset + a test asserting a delegated grant freezes nothing in a *shared* workspace), or scope the filter to the pointer branch so the shared-workspace lane keeps working.

---

# Verified — the four previous blockers, each broken to confirm it is pinned

## V1 (was §1) — the rationale. **Held.**

The rename is complete and clean: zero references to the old name survive anywhere in the tree, including docs, `AGENTS.md`, and the SDK.

The doc comment now opens *"ORACLE, NOT AN AUTHORIZATION"*, explains **why** the definer guard is self-defeating (`set_config` from its own argument, ordinary `opengeni_app` privilege, "only verifies that the caller supplied the same value twice"), records the evidence, and enumerates the two entitled subject shapes. `docs/organization-tenancy.md` does the same at length and generalises correctly: *"Do not treat any `*_self_*` definer routine here as a caller check without re-reading its guard."*

I read it as a future implementer would. It does prevent the failure mode: someone reaching for this helper with a request-derived subject hits *"Passing an attacker-influenced or merely looked-up subject is a vulnerability, not a policy decision"* before they hit anything encouraging. "Callers must prove entitlement locally" is **not** hand-wavy here, because the two entitled shapes are named concretely and the module-level comment in `personal-connection-delegations.ts` shows the pattern. Ask satisfied.

Two inaccuracies survive in the corrected prose (see N1, N2).

## V2 (was §2) — the delegated hole. **Closed, and airtight for every token shape the contract permits.**

I attacked this specifically, because `metadata.delegated === true` is one field and the turn-branch short-circuit sits *above* it. Enumerating `DelegatedAccessTokenPayload`'s `superRefine` (`packages/contracts/src/index.ts:2143-2226`):

| principalKind | attempt claims | reaches `kind:"subject"`? | outcome |
|---|---|---|---|
| `human_session` | **forbidden** (`exactAttemptClaimCount !== 0` → issue) | yes, always | caught by the new filter |
| `agent_attempt` | **all four required** | no — always `kind:"turn"` | turn branch probes only a frozen `ownerSubjectId` from a persisted row in the grant's own workspace |
| `service` | `turnId`/`attemptId`/`executionGeneration` forbidden → no turn branch | yes | caught by the pre-existing `principalKind === "service"` filter |

So a delegated bearer cannot route around the filter via the turn branch — `human_session` is structurally barred from carrying `turnId`. `delegatedAccessContext` is the only place in the tree that builds `workspaceGrants` inline, and it always stamps `delegated: true`. API-key subjects are `api_key:<id>` and `configured` subjects are forced to a `configured:` prefix, so both short-circuit the oracle's non-`user:` branch. The cross-workspace fallback (`accessGrantAuthorization` → `getWorkspaceGrant` when the token's workspace ≠ the requested one) yields a grant without the delegated marker, but `getWorkspaceGrant` is a bare membership join and returns `null` for a personal workspace, so the pointer branch stays unreachable there.

I could not get a bearer through. One fragility, N3.

## V3 (was §4) — the GUC restore. **Closed, and I agree with not widening `withRlsContext`.**

`priorSubjectId ?? ""` is semantically equivalent to unset: `opengeni_private.current_subject_id()` is `nullif(current_setting(...), '')`, and the two raw readers I found (`0255:102,313`, `0269:621`) only ever equality-compare, where `''` and `NULL` both fail against a real subject. It also matches the existing precedent in `0239:579` (`COALESCE(previous_subject_id, '')`).

Your reason for keeping it seam-local rather than adding `subject_id` to `restoreRlsContextSettings` is fine, but be aware of what it leaves: I count ~30 other `setSubjectRlsContext` sites inside `withRlsContext` scopes (`organization-membership-lifecycle.ts`, `knowledge-source-sync.ts`, `documents/src/index.ts`, `capability-integrations.ts`, `user-resource-authority.ts`, `company-brain-governed-writes.ts`) that still leak on a transaction handle. Your doc scopes the claim to *"two mechanical traps on this path"*, which is accurate — I am flagging it as a follow-up, not as a defect in this PR.

## V4 (was §5) — the same-organization matrix. **Real, and better than asked for.**

Verified against source, not the diff:

- `joinOrganizationAs` uses the genuine 0263 lifecycle — `createOrganizationInvitation` + `acceptOrganizationInvitation` with real revisions and operation ids, not a direct row insert;
- `expect(coMember.accountId).toBe(owner.accountId)` — **one account, two humans**, asserted, not assumed;
- `status === "active"` and `role === role` asserted, so the denial is not coming from a dead membership;
- the **positive control is present**: each role also asserts `namedSubjectHasLiveWorkspaceAuthority(owner.accountId, coMember.personalWorkspaceId, coMember.subjectId) === true`. The denial is specific, not blanket;
- and the co-member's pointer is read back through `listSelfOrganizationMemberships` — the same seam the resolver uses — rather than being fabricated.

Six tests (3 roles × {authority, freeze}), and every one of them reddens under sabotage.

---

# My independent sabotage numbers

Not taken on faith. I forced the probe true at each layer independently, in a clean worktree, against real Postgres as `opengeni_app` with FORCE RLS engaged. Baseline for the two new files: **19 pass / 0 fail**.

| # | What I broke | Result | Which tests reddened |
|---|---|---|---|
| **S1** | `namedSubjectHasLiveWorkspaceAuthority` → `return true` | **10 pass / 9 fail** | the 2 cross-org/other-human freeze+authority tests, all 6 same-org matrix tests, and the new isolated Drive test |
| **S2** | `ownerStillBelongsToWorkspace` (core) → `true` | 15 pass / **4 fail** | the 4 freeze-path tests only — correctly *not* the Drive path, which does not route through core |
| **S3** | `defaultPorts.getMembership` → `async () => true` | 33 pass / **1 fail** | **only** `an intruder-owned connection ... is denied by the authority probe alone` |
| **S4** | delete the `metadata.delegated === true` filter | 45 pass / **2 fail** | both delegated-bearer tests |
| **S5** | delete the GUC restore | 15 pass / **1 fail** | `probing a named subject does not leak that subject into the caller's transaction` |

**S1 reproduces your 10/9 exactly**, including the failure *set*, not just the count.

**On the Drive path specifically — your claim checks out, and I verified the control rather than trusting it.** The new test does this:

```ts
// Control: with the probe forced true the row IS reachable.
expect(await resolveGoogleDrivePublicationTarget(db, ws, [plantedDelegation],
  { getConnection: getConnectionMetadata, getMembership: async () => true })).not.toBeNull();
// Real resolver: denied by the probe.
expect(await resolveGoogleDrivePublicationTarget(db, ws, [plantedDelegation])).toBeNull();
```

The planted connection is genuinely `subjectId: intruder.subjectId` inside `owner.personalWorkspaceId`, so `getConnection` succeeds and every ownership/scope/destination check downstream passes. The control proves the *only* remaining fence is the probe. **S3 is the decisive evidence**: swapping the real resolver out of `defaultPorts` reddens exactly one test in the entire Drive corpus — this one. Before this PR that substitution would have been completely invisible. The path is pinned for the first time, as claimed.

Also worth stating plainly: `another human never resolves a publication target ...` still does **not** redden under S1, because it is denied downstream by connection ownership. Your test comment says so explicitly rather than implying broader coverage. I appreciate that.

---

# Claim 6 (declining `callerHasLiveWorkspaceAuthority`) — sound, and if anything understated

I checked each call site rather than accepting the count:

| Call site | Subject | Caller-scoped? |
|---|---|---|
| `freezePersonalConnectionDelegations` (subject branch) | `source.subjectId` = the grant's own subject | **yes** |
| `authorizedSocialConnectionsForGrant` (turn branch) | `delegation.ownerSubjectId` | no |
| `authorizedAtlassianConnectionsForGrant` (turn branch) | `delegation.ownerSubjectId` | no |
| `agent-turn.ts:8118` `delegatedOwnerHasMembership` | frozen `ownerSubjectId` | no |
| `google-drive-publication.ts` `defaultPorts.getMembership` | frozen `delegation.ownerSubjectId` | no |
| `viewer.ts:747` (pre-existing 0281) | `input.subjectId`, already gated on `subjectDelegated !== true` | **yes** |

So it is **four** of five non-caller, not three (N1). Either way the conclusion is right: a caller-scoped variant is an *addition*, not a replacement, and adding an API surface you cannot migrate the majority of your call sites onto is worse than not adding it in a bug fix. Not rationalisation.

---

# The concurrent `subjectHasLiveWorkspaceAuthorityInScope` variant

`origin/codex/personal-workspace-session-surface` introduces exactly the caller-scoped variant this PR declined, and also repairs `listSessionsForSubject:27400` and `setSessionPin:27857` — two of the seams your doc lists as still broken.

**Declining to adopt it here remains defensible.** It cannot serve four of your five call sites, and it does not exist on `main` yet.

But there is a real conflict, and it is not merely textual:

1. **Direct code conflict.** Both branches rewrite the same ~70 lines (`packages/db/src/index.ts:17457-17530`) in incompatible directions: you rename to `namedSubjectHasLiveWorkspaceAuthority` and extract `probeNamedSubjectWorkspaceAuthority`; the surface branch keeps the old name and extracts `subjectHasLiveWorkspaceAuthorityInScope`. Whichever lands second must reconcile deliberately, or the repo ends up with the arbitrary-subject oracle under **two** names.
2. **The surface branch does not carry your GUC restore.** If it merges second and resolves naively toward its own version, V3 silently regresses.
3. **Doc staleness both ways.** Your new section says a caller-scoped variant *"is deliberately not added here"* and lists `listSessionsForSubject`/`setSessionPin` as unfixed; both go stale the moment that branch lands. (Your handling of #1631 — *"0225's `guard_session_authority_write` … is repaired separately by migration 0302; do not add it back to this list"* — is exactly the right pattern; please do the same for this one.)
4. **Semantic caution, worth raising now.** The surface branch's doc says the in-scope variant *"asserts that the subject matches the transaction's already-applied, authenticated scope, so it cannot name a third party."* That is the same shape of over-claim this PR just spent a round correcting: `setSubjectRlsContext` accepts any string from any `opengeni_app` connection, so "already-applied, authenticated scope" is a **convention**, not a proof. It is a genuine improvement — it forces the caller to have gone through scope establishment — but it is defence in depth, not structure. That branch's design (a positive `PersonalWorkspaceOwnerException` flag from `grantIsCanonicalManagedHumanSession`) is also strictly better than this PR's blocklist, and worth converging on later.

---

# Nits

**N1.** `docs/organization-tenancy.md`: *"three of this path's five call sites legitimately ask about a non-caller subject."* It is four of five (table above). Understating it weakens your own argument.

**N2.** Same paragraph: *"the deeper fix — having `list_self_organization_memberships` read the GUC instead of taking the subject as a parameter — is a migration with repo-wide blast radius."* **That fix would not make it structural.** The caller sets the GUC via ordinary `set_config`; reading it instead of a parameter changes nothing about who may set it. This is the same reasoning error the surrounding section exists to correct, and it points the follow-up author at a design that will feel safe and won't be. The actual structural fix is to stop letting the application role set `opengeni.subject_id` freely (a definer-only setter, or an attested subject) — a much larger conversation. Please reword.

**N3.** The `metadata.delegated === true` check sits **after** the `sessionId`+`turnId` turn-branch short-circuit. Today that is airtight only because `DelegatedAccessTokenPayload`'s `superRefine` forbids `human_session` from carrying attempt claims (V2). That coupling is load-bearing and undocumented. A comment at the turn branch, plus a contract test asserting `human_session` + `sessionId` is rejected, would keep a future refinement change from silently reopening this.

Relatedly, the module comment says agent-attempt/service/delegated grants *"are all collapsed to `{kind:"none"}` before they can reach here."* Strictly they are collapsed to `none` **or `turn`**. The consequence you assert is true; the mechanism as stated is not.

**N4.** `finally { set_config(...) }` in the probe will itself throw `25P02 current transaction is aborted` and **mask the original error** if the probe's SQL fails inside a caller's transaction. Unreachable in practice (the definer guard can't fire — the function sets the GUC to the same value — and the `user:` prefix is pre-checked), but a `.catch(() => undefined)` on the restore would cost nothing.

**N5.** `scheduled-tasks.ts` — the `input.grant.workspaceId` fix is right and I verified the comment's premise at all four callers (HTTP `requireAccessGrant(c, deps, workspaceId)` guarantees `grant.workspaceId === param`; MCP loads with `requireScheduledTask(db, grant.workspaceId, id)`). But `settingsWithEnabledCapabilityMcpServers` twenty lines above still reads `input.existing.workspaceId`. Same value today; inconsistent for the same reason you just fixed.

---

# Verification — everything reproduced, nothing taken on faith

Clean worktree at `9731ab62`, `bun install` from scratch, baseline worktree at `origin/main` `6f61d6ee`.

| Check | Result |
|---|---|
| `typecheck` | **exit 0** |
| `lint` | **exit 0** |
| `format:check` | **exit 0** |
| `check:docs-refs` | **exit 0** |
| `check:migration-ordinals` | **exit 0** |
| `check:public-hygiene` (post-commit) | **exit 0** |
| `packages/core/test` + `apps/worker/test` | **1536 pass / 0 fail** |
| `packages/db/test` (whole directory) | **1504 pass / 0 fail** |
| `apps/api/test` on this branch | 1303 pass / **17 fail** |
| `apps/api/test` on `origin/main` @ `6f61d6ee` | 1303 pass / **17 fail** |
| sorted failing-test-name diff, branch vs main | **empty — identical failure sets** |

All 17 come from `apps/api/test/sandbox-desktop-stream.test.ts` and are present identically on pristine `main`. **No new failures.** (You reported 1526 for core+worker; I get 1536 — you added tests since. Immaterial.)

**Flaky-test claim (item 7) — I could not reproduce it, and I looked hard.** `session-pins.test.ts` + `migration-0263-organization-membership-lifecycle.test.ts` combined: **51 pass / 0 fail, five consecutive runs**, plus **two more runs executed concurrently with a full `packages/db/test` sweep** to force contention — still 51/51. And that full `packages/db/test` sweep itself finished 1504/0 *while* being hammered.

So I cannot confirm your 2-failure observation, but I can say it is not an ordering or isolation bug this change introduced, on three independent grounds: (a) it does not reproduce under deliberately worse contention than a normal run; (b) the whole `packages/db` directory is green under that same contention; (c) mechanically, this PR's only `packages/db/src/index.ts` change is the two hunks at `17457-17530` — `setSessionPin`, `list_self_organization_memberships`, the 0263 lifecycle functions, and every migration are untouched, so there is no surface through which those suites could newly interact. Contention is the plausible explanation; if it recurs, please capture the failing assertion text rather than the count.

**Also verified:** the new DB tests genuinely execute (they acquire a real `opengeni_app` connection with FORCE RLS, not a no-op through `if (!client) return`); `getWorkspaceGrant` is byte-for-byte unchanged; the only `packages/db` hunks in the whole PR are the two around the oracle; the modified Drive/delegation tests are mechanical signature adaptations with no weakened assertions (`getMembership: async () => ({}) as never` → `async () => true` is the same truthiness, now type-honest).

---

## Summary

The engineering is done. The bug was real, the fix is right, the rationale is now honest and instructive, the delegated hole is closed against every token shape the contract permits, the GUC trap is closed, the same-organization matrix is genuine with a positive control, and — the part I cared most about — the Drive path is pinned by the probe for the first time, which I proved by removing the probe from `defaultPorts` and watching exactly that one test and no other go red.

Fix **B1** (one paragraph in the PR description — it becomes `main`'s commit message), **B2** (a changeset; `@opengeni/db` is publicly published and you removed an export), and make a call on **B3** (say the shared-workspace delegated narrowing is intended and test it, or scope the filter), and I would approve without reservation.

benyhh:
## B1, B2, B3 addressed — `1615f26e`

### B1 — PR body rewritten

You were right that this was the one place that mattered most, and I had left it. The description's "How the exception stays owner-only" section is gone. In its place is **"Why this helper is an oracle, not an authorization"**, which opens by saying the previous claim was false, states the mechanism (the function sets the GUC from its own argument, so the guard only checks the caller supplied the same value twice), cites the empirical result, and carries the generalisation about `*_self_*` definer routines. Verified: zero occurrences of the old phrasing remain in the body.

**N1** corrected to four of five, in both the body and `docs/organization-tenancy.md`.

**N2** — you are right, and this one was worth catching. My "deeper fix" would not have made anything structural, for precisely the reason the surrounding section exists to explain: `setSubjectRlsContext` is an ordinary `set_config`, so reading the GUC instead of a parameter only changes *how* the caller names its subject. I had reproduced that mechanism, written it down, and then reached for a design that violates it one paragraph later. The doc now says so explicitly and names the actual structural fix — stop letting the application role set `opengeni.subject_id` at all (definer-only setter, or an attested subject fixed at authentication) — and flags it as a much larger conversation rather than a follow-up someone can pick up casually.

### B2 — changeset

`.changeset/personal-workspace-connection-authority.md`, **major** for both `@opengeni/db` and `@opengeni/core` (`changeset status` confirms it is picked up). It declares the rename, the required `accountId` on `PersonalConnectionDelegationSource`, and the delegated behaviour change, each with the consumer-facing consequence.

**No back-compat alias, deliberately.** Re-exporting `subjectHasLiveWorkspaceAuthority` would put the misleading name back in the public API — the exact label this PR spent two rounds removing. A hard rename with a release note is the smaller cost.

### B3 — kept broad, and now stated

Kept the broad denial. Reasoning matches yours: a delegated payload's subject is signed token content with no row behind it, so borrowing someone's private provider credentials in a *shared* workspace is the same defect one room over, and scoping the filter to personal workspaces would leave it standing.

Now stated rather than implied:

- **`docs/embedding.md`** gains a "Delegated tokens carry no personal-connection authority" subsection under the host-minted-token paragraph you pointed at: says plainly that this is a behaviour change, that sessions created through a delegated token get workspace-owned connections only, why, and what a host should do instead.
- **`docs/organization-tenancy.md`** records it as an intended widening.
- **The PR body** has its own section rather than describing this as the personal-workspace fix.
- **The missing test exists.** You were right that no test anywhere paired `delegated: true` with `freezePersonalConnectionDelegations` — that is why 1536/1536 stayed green through a behaviour change. The new test uses the *legacy shared* workspace with a real membership row, asserts the canonical cookie grant freezes 1, then flips only `metadata.delegated` and asserts 0.

Sabotage on the filter (removing it): **3 fail** — the two personal-workspace delegated tests plus the new shared-workspace one.

### Nits

- **N3** — comment at the turn branch marking the ordering load-bearing and naming the exact refinement it depends on, plus `packages/contracts/test/delegated-access-token-attempt-claims.test.ts` pinning that a `human_session` payload cannot carry `sessionId`/`turnId`, with a header saying to move the filter above the turn branch if that ever changes. Also fixed the module comment: delegated/agent grants resolve to `none` **or `turn`**, as you said — the consequence held but the stated mechanism did not.
- **N4** — restore wrapped in `.catch(() => undefined)` so it cannot throw `25P02` and replace the original diagnostic.
- **N5** — `settingsWithEnabledCapabilityMcpServers` now reads `input.grant.workspaceId` too.

### Merge hazard recorded

`docs/organization-tenancy.md` now carries a merge note next to the mechanical traps: converge on one name for the oracle, drop the two repaired seams from the still-broken list, and — stated explicitly — **whichever change lands second must preserve the subject-GUC restore, because that branch does not carry it.** I also noted that its `PersonalWorkspaceOwnerException` flag is the better shape and worth converging on, and that its "cannot name a third party" claim needs the same correction this section makes. Not resolving it from my side.

### Flake — dropping it

Understood, and thank you for the thoroughness. Five runs plus two under forced contention is far better evidence than my single observation, and your three mechanical grounds are convincing. Treat my report as noise; I am not chasing it.

### Verification

`typecheck` 31/31 · `lint` · `format:check` · `check:docs-refs` · `check:public-hygiene` · `check:migration-ordinals` (now reports base highest **0302**, so `main` has moved and this branch still carries no migration) · `check:migration-rls-backfills`.

`packages/core/test` + `apps/worker/test` + `packages/contracts/test` → **1975 pass / 0 fail**. Format ran last; no rebase, no merge from `main`.

benyhh (commented):
APPROVE — exact head `62c518b2824953ac6def5a2761fecbfe45ce7427`.

Independent integrated-head review found no blocking defect.

Authority/security audit:
- The head retains #1641's single canonical rule in `subjectHasLiveWorkspaceAuthorityInScope` and its explicitly named public oracle `namedSubjectHasLiveWorkspaceAuthority`; no second resolver or stale old export survives.
- The session surface remains positively gated by `AccessGrantAuthorization.canonicalManagedHumanSession`, stamped only after Better Auth cookie verification. The inherited tests prove delegated bearers, API keys, services, same-organization members/admins/owners, account admins, and unauthenticated callers do not receive the personal-workspace exception.
- Connection delegation denies `metadata.delegated === true` before a subject source can be produced. The turn-branch coupling is closed by the signed-token contract: human-session tokens cannot carry attempt claims, service tokens cannot carry a turn, and exact agent-attempt tokens inherit only already-frozen authority. New contract tests pin this.
- Every new named-subject query is locally entitled: either the authenticated grant subject or an already-frozen delegation owner. Same-organization member/admin/owner matrices include positive controls for each caller's own personal workspace.
- Account/workspace pairs come from one authority object at production call sites: grant/grant in core scheduled paths, turn input/input in the worker, and the explicit workspace tuple for Drive.
- Subject-GUC restoration is preserved on success and failure. Successful and non-database-failure paths explicitly restore; an aborted SQL path suppresses only a masking `25P02`, then nested savepoint/outer rollback restores scope. Real-Postgres tests prove pre-existing and unset subject contexts survive successful probes.
- #1637 is preserved byte-for-byte in the integrated tree: current `main` is an ancestor and this PR has no diff in #1637's ownership/callback files.
- Changesets are accurate: `@opengeni/core` is major for the public `accountId` and delegated-behaviour breaks; landed #1641 is the sole `@opengeni/db` major for the helper rename.
- Documentation accurately distinguishes oracle from authorization, discloses shared-workspace delegated narrowing, and preserves the positive Better Auth provenance boundary.

Independent exact-head verification:
- 73/73 focused tests passed across session surface, connection authority, Drive publication, and delegated-token contracts.
- Contracts, DB, core, and worker TypeScript checks passed independently.
- Changed-file format/lint, workspace/billing static guard, changeset status, and `git diff --check` passed.
- `origin/main` at `2027bc3ef518cc48fd898693c4c367804b3b91f1` is an ancestor; GitHub reports MERGEABLE.

CI snapshot at review time: source/type contracts, deployment artifacts, impacted E2E shards, multiple unit/real-service shards, and native/WASM smoke receipts are green; remaining exact-head jobs are still running. This is source approval, not a waiver of the required all-green exact-head CI gate.

PR #1641, opened 2026-08-19
fix: resolve personal-workspace authority on the session surface

benyhh:
## The bug, proven against the real routes

A managed human's personal workspace deliberately has **no `workspace_memberships` row** — `0219_organization_tenancy_managed_human_provisioning.sql` raises `42501` if one appears. Their access is the `organization_memberships.personal_workspace_id` pointer. Three session seams fenced on a bare membership probe and therefore denied the one human who always belongs.

Driven through the real Hono routes with a real Better Auth cookie session (`auth_users` / `auth_identities` / `auth_sessions` + `synchronizeCanonicalHumanLoginBindings`, then `GET /v1/access/me` running the real managed provisioning lifecycle), on `origin/main` at `6f61d6ee`:

```
(fail) GET  /v1/workspaces/:id/sessions          in the owner's own personal workspace  → 403
(fail) PUT  /v1/workspaces/:id/sessions/:id/pin  in the owner's own personal workspace  → 403
(fail) PUT  /v1/workspaces/:id/new-session-draft in the owner's own personal workspace  → 403
```

The premise is pinned by its own test: `select count(*) from workspace_memberships where workspace_id = <personal>` is `0`, while the same human's legacy Better Auth workspace has exactly `1` — and the legacy workspace returns `200` for the same request. So this is the personal workspace specifically, not a broken fixture.

The list 403 is the highest-impact one: `sessions:read` is in `managedPersonalWorkspacePermissions`, so the route authorizes and the workspace renders — then lists nothing and shows an access error.

### After

```
apps/api/test/personal-workspace-session-surface.test.ts     18 pass 0 fail
packages/db/test/personal-workspace-session-tenancy.test.ts   7 pass 0 fail
```

(The `packages/db` file pins the two SQL seams as *known defects* rather than asserting a fix — see below.)

### One implementation of the rule, and it is not an oracle

`packages/db/src/workspace-authority.ts` (new) holds `subjectHasLiveWorkspaceAuthorityInScope` — the single implementation of "does this subject still hold workspace authority here": a `workspace_memberships` row whose owning organization membership is active, **or** an active organization membership whose `personal_workspace_id` is exactly this workspace. The exported `subjectHasLiveWorkspaceAuthority` (`packages/db/src/index.ts`) now delegates to it, so there is no second copy.

Two differences from the exported function matter:

1. **It never sets `opengeni.subject_id`.** It runs on the caller's already-scoped transaction and raises if the requested subject differs from that scope.
2. Non-`user:` subjects short-circuit to the plain membership answer instead of tripping the seam's `42501`.

### ⚠️ That raise is a consistency check, NOT an authorization

Review corrected me here, and the correction matters more than the code. `withWorkspaceSubjectRls(db, workspaceId, subjectId, …)` sets `opengeni.subject_id` **from the same argument** the resolver later compares against. At all three seams both sides come from one variable, so the check is a **tautology** there — my own G2 sabotage proves it, reddening only a synthetic mismatched-scope test and nothing real.

So this PR ships **one guard, not two**. The provenance stamp does all the authorization work. The resolver answers "does subject X hold authority here" and establishes nothing about who the caller is.

What the in-scope variant still earns — stated precisely rather than dressed up as a second guard:

- it makes the **arbitrary-subject oracle shape unrepresentable** at these seams, because it has no way to scope a transaction itself;
- it keeps the authority read **inside the caller's transaction, snapshot, and advisory fence** (`lockSessionPersonalStateShared`/`Exclusive`, the draft's `FOR KEY SHARE`), so it serialises with `removeWorkspaceMember()`;
- it keeps the corrected rule to **one implementation**, shared with the oracle;
- the raise is a **tripwire** against a future refactor that makes scope and argument diverge.

I judged that worth keeping. It is not worth calling a security guard, and `AGENTS.md`, `docs/organization-tenancy.md`, both doc comments, and the changeset have been corrected — they had been calling the inherited scope "authenticated", which re-encodes in normative repo guidance exactly the false claim that got #1638's docs rejected.

### ⚠️ `namedSubjectHasLiveWorkspaceAuthority` is an arbitrary-subject oracle

The exported helper sets `opengeni.subject_id` **from its own argument**, then consults `list_self_organization_memberships`, whose `42501` guard compares the requested subject against that GUC — which the function just set to the caller's parameter. Its guard is therefore satisfied for *whatever subject the caller names*. It is **not** owner-only by construction; it is owner-only only if the caller can prove the subject is its own. Passing it a request-derived subject is a vulnerability.

Its behaviour as an oracle is unchanged and deliberate — some callers legitimately ask about a non-caller subject. What changed here: it is **renamed** so the hazard is visible at the call site, its docstring says so plainly, and it now **restores the subject GUC** it used to leak (see below).

**Follow-up, not done here:** migrate the call sites where the subject *is* the caller onto the in-scope variant, then lock down or delete what remains. Of #1638's five sites, two can move; three legitimately cannot.

### Provenance, not grant shape

Reusing the resolver alone reopened the hole in a new place: with only the resolver call, a delegated bearer whose host-signed token named the owner's `user:` subject went **403 → 200** on the owner's personal workspace. My denial test caught it before merge.

Inspecting the grant is not a sound gate — a delegated bearer chooses its own `principalKind`, `metadata.delegated`, `serviceInitiator`, and `subjectId`. So the gate asks **how the request authenticated**:

- `canonicalManagedCookieContexts`, a `WeakSet<AccessContext>` stamped **only** inside the one branch of `resolveAccessContext` that verified a Better Auth cookie.
- Surfaced as `AccessGrantAuthorization.canonicalManagedHumanSession` = stamped **and** `grant.subjectId === context.subjectId` **and** `user:`-prefixed.
- The three db seams take an explicit `personalWorkspaceOwnerException` flag that **defaults to absent** — every caller that has not proven provenance gets the historical bare-membership fence verbatim.

**Fails closed by construction.** Membership is only ever *added* at the cookie branch, so every other path — delegated bearer, API key, local bootstrap, configured key, and any path added later — is absent without anyone maintaining an exclusion list. A request that never resolves a context has no context to test (covered: unauthenticated → `401`).

**Keyed on the `AccessContext`, deliberately not the `Request`.** A reused, wrapped, or retried request object cannot carry a stale positive into a differently-authenticated resolution, and the non-cached `requireFreshAccessGrant` path is stamped correctly too. (Keying on `Request` would also have been strictly weaker than the existing `accessContextByRequest` cache's own assumption.)

## Per-seam subject origin (final code)

| Seam | Subject origin | Request-controllable? | Gate |
|---|---|---|---|
| `listSessionsForSubject` ← `GET /v1/workspaces/:id/sessions` | `grant.subjectId` | **Yes** on the delegated path — `delegatedAccessContext` builds the whole context from the token | `authorization.canonicalManagedHumanSession` |
| `setSessionPin` ← `PUT …/sessions/:id/pin` | `grant.subjectId` | **Yes**, same | `authorization.canonicalManagedHumanSession` |
| `saveNewSessionDraftInTransaction` ← `PUT …/new-session-draft` | `grant.subjectId` | **Yes**, same | `authorization.canonicalManagedHumanSession`, threaded through `saveActorNewSessionDraft` (defaults `false`) |
| `listSessionsForSubject` ← Slack App Home | Slack user-link subject | No cookie session | flag never set → behaviour byte-identical |

## Denial matrix

Same-organization cases use the **real 0263 invitation lifecycle** (`createOrganizationInvitation` → `acceptOrganizationInvitation`, which provisions the invitee's own personal workspace under the same account), not a hand-built row — so "same organization" is genuine rather than an artefact of minting a fresh account per human.

Every non-owner principal is asserted against **all three seams**, so a widening at one cannot hide behind another's denial.

| Principal | List | Pin | Draft |
|---|---|---|---|
| **Owner, own personal workspace** | 200 | 200 | 200 |
| Same-org **co-member** (0263 invite) | 403 | 403 | 403 |
| Same-org **ADMIN** (0263 invite, `role='admin'` asserted) | 403 | 403 | 403 |
| Organization **OWNER** (`role='owner'` asserted) | 403 | 403 | 403 |
| Human in a **different organization** | 403 | 403 | 403 |
| **Account-admin** API key (`account:read`,`account:admin`) | 403 | 403 | 403 |
| **Delegated bearer**, substituted `user:` subject = the owner's | 403 | 403 | 403 |
| **Service initiator** (delegated, `principalKind: "service"`) | 403 | 403 | 403 |
| **Unauthenticated** | 401 | — | — |
| **Workspace-scoped API key** on the personal workspace *(unconstructible in production — see below)* | 200 | 403 | 200 |

The co-member's *own* personal workspace still returns `200` in the same test, so these denials are the exception being owner-scoped rather than the co-member being broken.

### The one row that is not a denial — and why

A workspace-scoped API key minted **on** the personal workspace returns **200 / 403 / 200**, unchanged by this PR (verified on pristine `main` before touching anything).

My first explanation for why that is safe was **wrong**, and review caught it: I wrote that `managedPersonalWorkspacePermissions` includes `api_keys:manage`. It does not — `AGENTS.md` says so one paragraph above the text I edited. Driving the real route, `POST /v1/workspaces/:id/api-keys` with the owner's own cookie is **403 `missing permission: api_keys:manage`**, and that route is the **only production caller of `createApiKey`**.

So the principal is **unconstructible in production**; my test mints it below the route. Pinning the three statuses alone asserted that 200 is *correct* without asserting the unreachability that makes it *safe* — it would have stayed green on the day that hole opened. The pin is now the unreachability itself:

- a test asserts the **route refuses to mint** such a key and that `api_keys:manage` is absent from the personal permission set;
- the status pin is explicitly labelled "reachable only below the route, safe only because the route denies it";
- sabotage confirms it is load-bearing — adding `api_keys:manage` to the permission set reddens it. (My first sabotage attempt silently patched a *different* permissions array and proved nothing; I retargeted it rather than accept the green.)

Also recorded, from review: such a key **outlives the authority** that would have created it. After suspending the organization membership the key still returns 200 while the owner's cookie does not. Observed as a **500**, not a clean 403 — the lifecycle seam raises `assert_active_managed_human_organization_membership` and nothing converts it. Verified pre-existing on pristine `main` with the same probe, so it is recorded here, not fixed.

## Sabotage results — 8 guards, 12 reddening events across 10 distinct tests

Re-run in full **after** the `main` merge and **after** the rename, since either can quietly detach a guard from its test.

| Sabotage | Tests reddened |
|---|---|
| **G1** provenance stamp forced `true` *(the only authorization)* | 2 — delegated bearer, service initiator |
| **G2** in-scope consistency check removed | 1 — the synthetic mismatched-scope test, and nothing else. **This is the evidence that it is not a second guard.** |
| **G3** resolver `personal_workspace_id` equality dropped | 3 — organization **OWNER**, same-org **ADMIN**, same-org member |
| **G4** `personalWorkspaceOwnerException` ignored at all 3 seams | 2 — delegated bearer, service initiator |
| **G5** non-`user:` short-circuit removed | 1 — API-key behaviour pin |
| **G6** membership `status = 'active'` checks dropped | 1 — suspended organization membership |
| **G7** `api_keys:manage` added to the personal permission set | 1 — the route-refuses-to-mint unreachability pin |
| **G8** subject-GUC restore removed | 1 — the GUC-leak regression test |
| **Fix removed** (exception never applies) | 7 — all 3 positive seams + 4 owner-scope tests |

Note on G1: the same-org admin and organization owner do **not** redden under G1, because today they are denied one layer earlier — `accessGrantAuthorization` finds no grant on another member's personal workspace. The guard that owns those cases *at the seam* is **G3**, and it reddens exactly there. Both layers are pinned: the G3 tests call `listSessionsForSubject` / `setSessionPin` / `saveNewSessionDraftInTransaction` **directly** with `personalWorkspaceOwnerException: true` — the strongest thing the API layer could ever assert.

## The two SQL seams — verified broken, deliberately not fixed here

Both confirmed by driving the real adapters against real Postgres:

```
transition_session_visibility → 42501 "session visibility transition requires active membership"
fork_session_content          → 42501 "session fork source workspace access is unavailable"
```

They are `SECURITY DEFINER`, so **no application-layer fix exists** — a migration is the only option. Not shipped here because the ordinal would collide with **#1631**'s `0302`, which is already rewriting session-ownership behaviour in these same functions.

Rather than leave a red test, `packages/db/test/personal-workspace-session-tenancy.test.ts` **pins the current wrong behaviour** by its exact PostgreSQL message, with the replacement disjunct recorded in the file. A third test runs the same two operations in an ordinary workspace and expects success, so the pins record a personal-workspace defect rather than a generally broken seam. Whoever ships the migration inverts the two pins.

**No production caller — verified, and machine-enforced.** `test/session-visibility-contract-surface.test.ts` already pins exactly this: no file under `apps/api/src`, `apps/worker/src`, `apps/web/src`, `packages/core/src`, `packages/runtime/src`, `packages/sdk/src`, `packages/react/src`, or `packages/contracts/src` may name either SQL function or either TS wrapper, with a 5-file allowlist (the two migrations, the adapter, `provision-roles.ts`, `runtime-posture.ts`). It passes. An independent sweep reproduced it: every caller of `transitionSessionVisibility` (11 files) and `forkSessionContent` (2 files) is a `packages/db` test; no other SQL object, trigger, view, policy, or `DO` block calls them; there is no `/visibility` or `/fork` route (`PATCH …/sessions/:id` is title-only — its schema is `{ title }`), no MCP tool, and no SDK/React/web client call. The alternate write path is closed too: `opengeni.session_visibility_write_capability` has **zero** non-test minters, so production cannot move a session off `workspace_shared` by either route. `AGENTS.md:100` documents the same property.

**Exact one-line disjunct for whoever picks this up** — the pattern `0258_three_scope_document_knowledge_authority.sql:404` already uses. In all three places `actor_membership` is the actor's own `organization_memberships%ROWTYPE`, already `SELECT`ed with `status = 'active'`, so this is owner-only by construction:

```sql
-- 0225 transition_session_visibility (~line 590), wrapping the existing check:
IF actor_membership.personal_workspace_id IS DISTINCT FROM p_workspace_id THEN
  <existing NOT EXISTS (SELECT 1 FROM workspace_memberships ...) RAISE>
END IF;

-- 0289 fork_session_content, source check (~line 238):      ... IS DISTINCT FROM p_source_workspace_id
-- 0289 fork_session_content, destination check (~line 355): ... IS DISTINCT FROM p_destination_workspace_id
```

`transition_session_visibility` additionally requires a non-null `owner_organization_membership_id`, which is precisely what #1631's `0302` repairs at the mint. That seam needs **both** halves; this disjunct alone is not sufficient.

## Deliberately not widened

`getWorkspaceGrant` is unchanged. It is the fallback inside `accessGrantAuthorization` (`packages/core/src/access/index.ts`), where a delegated bearer carrying a `user:` subject, or an API-key principal, would inherit any widening for a workspace outside its token. That is exactly the widening `AGENTS.md` forbids.

**Phase D:** the only authority read is the membership's own `personal_workspace_id`. Nothing is inferred from `created_by`, connection attribution, a default workspace, a resource name, or current access.

## Verification

`typecheck` (31 projects clean) · `lint` · `format:check` (run last) · `check:public-hygiene` (post-commit) · `check:docs-refs` · `check:migration-ordinals` (310 migrations, base `origin/main` highest 0301 — **no migration added**) · `test/session-visibility-contract-surface.test.ts` (5 pass, 6774 assertions — the inertness pin still holds).

Suites against real Postgres, with the baseline **re-established on the new pristine `origin/main` `13c3af4f`** in a separate worktree (the old baseline went stale when `main` moved):

| Suite | Baseline (pristine main `13c3af4f`) | This branch |
|---|---|---|
| `packages/core` | 359 pass / 0 fail | 359 pass / 0 fail |
| `packages/db` | — | 1521 pass / 0 fail |
| `apps/api` | 1303 pass / 18 fail | **1322 pass / 17 fail** |

The `apps/api` failure-name `diff` against that new baseline has **zero** `>` lines — no failure this branch introduces. The single difference is one *baseline-only* failure, `(fail) (unnamed) … a beforeEach/afterEach hook timed out for this test` at exactly 5000 ms — the same class of default-timeout flake under concurrent machine load documented below, which happened to land on the baseline run and not on the branch run. The +19 passing tests are exactly this PR's new tests.

New DB-backed tests use `acquireSharedTestDatabase` (the `opengeni_app` app role, not superuser) so FORCE RLS actually engages, and carry explicit `180_000` timeouts including `beforeAll`/`afterAll`.

## Coordination

- Branched from `origin/main` `6f61d6ee`, and **`origin/main` `13c3af4f` is merged in** (not rebased). That brings in #1631's migration 0302 and #1639. The merge is required, not tidying: the PR was `CONFLICTING`, which means this repo gives it **no CI run at all**, and the moving-main contract permits mutating a candidate for an actual merge conflict.
- Did **not** touch #1637's files (`provider-oauth.ts`, `google-drive.ts`, `atlassian.ts`) or #1638's (`personal-connection-delegations.ts`, `google-drive-publication.ts`).

### Why two resolvers exist, rather than one replacing the other

#1638 keeps an arbitrary-subject resolver because three of its five call sites legitimately ask about a **non-caller** subject — the frozen owner of a delegation the caller already holds. A caller-scoped API cannot express that, so the in-scope variant is an **addition**, not a replacement.

My three seams are the opposite case: for `listSessionsForSubject`, `setSessionPin`, and `saveNewSessionDraftInTransaction` the subject **is** the caller. The two functions encode two genuinely different questions — "does this *named* subject have authority" vs "does the subject this transaction is already scoped to have authority" — not an accident of refactoring.

### The GUC restore, and a correction to an earlier version of this PR body

An earlier revision of this section said "the two PRs do not touch the same primitive" and prescribed: *take #1638's name and docstring, keep this PR's delegating body*. **That was wrong and dangerous, and it is retracted.**

`withRlsContext` restores `account_id`/`workspace_id` but **never** `subject_id` (`packages/db/src/database.ts`). #1638's GUC-restore fix therefore lives **inside the body of the public helper** — the exact body this PR replaces with a delegation. Following that recipe literally would have deleted their fix, leaving no failing test and no conflict marker behind. Since this repo squash-merges, the recipe would have become the permanent commit message.

**This PR now carries the restore itself**, so the trap is gone regardless of resolution order: the helper captures the prior subject GUC and restores it after probing, using `""` as the canonical unset (`current_subject_id()` is `nullif(…, '')`, the 0239 precedent) so a transaction that had no subject ends with none rather than a literal empty one. A regression test asserts both the preserved-subject and still-unset cases; sabotage **G8** confirms it reddens when the restore is removed.

### Adopted from #1638: the rename

The public helper is renamed `subjectHasLiveWorkspaceAuthority` → **`namedSubjectHasLiveWorkspaceAuthority`**. The previous pair gave the *dangerous* function the short inviting name and the *safe* one the qualifier — backwards. The `named` prefix now says at the call site that you are naming a subject, not proving you are one.

Breaking for `@opengeni/db` consumers with **no alias**, deliberately: re-exporting the old name would put the misleading label back in the public API. **This PR declares the required `@opengeni/db` major at the first landing of the rename.** After this PR merges, the companion #1638 changeset should retain only its separate `@opengeni/core` major.

### Expected conflict with #1638

Now much smaller, because the name, the restore, and the corrected docstring all converge. What remains in `packages/db/src/index.ts` is one function whose body both sides rewrote to the same shape. **Do not resolve by deleting either side's GUC restore** — the restore must survive resolution. Also `docs/organization-tenancy.md`, where both add a section.

Deliberately **not** adopted from #1638, by agreement: its delegated-source blocklist, and any convergence of the authorization shape. `PersonalWorkspaceOwnerException` + the `canonicalManagedHumanSession` positive stamp stays as-is; convergence belongs in a separately reviewed follow-up, not inside a conflict resolution.

### Review minors addressed

- **F3** — `saveNewSessionDraftInTransaction` now derives the account from `accountIdInRlsScope(db)` like the other two seams, instead of the caller-supplied `input.accountId`. Not exploitable before (a mismatched account could only narrow the answer), but it removes a caller-controlled input from an authority decision for free.
- **F4** — the orphaned doc blocks in `workspace-authority.ts` are fixed; each helper now carries exactly one, and `accountIdInRlsScope` is documented.
- **F7** — `accountIdInRlsScope` is no longer published. The module is re-exported by explicit name rather than `export *`, so a helper that trivially satisfies the in-scope consistency check stays inside the package.
- The two surviving "authenticated" descriptions of the inherited scope are gone; both remaining uses of the word now explicitly *deny* that anything authenticated the subject.

### Follow-up this PR does not do

Migrate #1638's five call sites onto `subjectHasLiveWorkspaceAuthorityInScope` where the subject is the caller, and lock down or delete the remaining oracle. Two of its five sites can move; three legitimately cannot and need the oracle to stay, with its warning intact.


benyhh (commented):
# Adversarial security review — verdict: REQUEST CHANGES (do not merge as-is)

**The runtime fix is sound. I attacked the provenance boundary and could not break it.** What I am blocking on is not code: it is that this PR writes an overstated security guarantee into `AGENTS.md` and `docs/organization-tenancy.md` — overstated in *precisely* the way #1638 was caught for — and justifies its one deliberate non-denial with a claim that is factually false. Both are text/test fixes, not architecture changes.

Everything below is independently reproduced in a clean clone at `5727bf1e` vs pristine `origin/main` `6f61d6ee`, against real Postgres. I did not take the diff's word for anything.

---

## What I verified and agree with

**The bug is real.** I rebuilt the PR's test file against pristine `main` (stripping every PR-only symbol) and ran it there. 7 tests red on main, with exactly the failure set the PR claims:

```
(fail) GET  /v1/workspaces/:id/sessions          in the owner's own personal workspace  → 403
(fail) PUT  …/sessions/:id/pin                   in the owner's own personal workspace  → 403
(fail) PUT  …/new-session-draft                  in the owner's own personal workspace  → 403
(fail) a SAME-organization co-member … (their OWN personal workspace also 403s on main)
(fail) same-org ADMIN / MEMBER / SUSPENDED at the db seam
```

**Sabotage: my numbers match yours exactly.** 6 guards, run independently, each applied alone to a clean tree, both new suites (`apps/api` + `packages/db`, 23 pass baseline):

| Guard | My result | Reddened tests |
|---|---|---|
| G1 provenance stamp forced true | **2** | delegated service initiator; delegated bearer w/ substituted subject |
| G2 in-scope subject assertion removed | **1** | the arbitrary-subject-oracle test |
| G3 `personal_workspace_id` equality dropped | **2** | same-org ADMIN (db seam); same-org MEMBER (db seam) |
| G4 `personalWorkspaceOwnerException` ignored at all 3 seams | **2** | delegated service initiator; delegated bearer |
| G5 non-`user:` short-circuit removed | **1** | workspace-scoped API-key regression pin |
| G6 membership `status='active'` dropped | **1** | suspended organization membership |
| Fix removed | **7** | 3 positive seams + 4 owner-scope tests |

9 reddening events across 7 distinct tests. Confirmed. **And your own caveat is correct and important**: the same-org ADMIN does *not* redden under G1 — only the two delegated tests do — and it *does* redden under G3, at the db seam, via the direct-seam tests. G3 genuinely owns that case at the seam. Defence in depth is real here, not asserted.

**The `WeakSet<AccessContext>` stamp holds up.** I tried to get a non-cookie principal to carry a stamped context and failed:
- `canonicalManagedCookieContexts.add` has exactly one call site, inside the one `deps.managedAuth` branch.
- `ensureManagedAccessForUser` builds and returns a **fresh object literal per call** inside its transaction — no memoization, no module-level cache, no shared instance. So a stamp can never travel to a second resolution.
- Every earlier `return` in `resolveAccessContext` (local bootstrap, configured delegated/API-key/bootstrap, managed delegated, managed API-key) leaves its context unstamped. A bearer that *fails* to verify falls through to the cookie branch, but the context produced there is the cookie human's own — not the bearer's claimed subject. Not an escalation.
- The only per-request cache (`accessContextByRequest`) is keyed on `Request` and returns the same stamped object for the same authenticated request; `requireFreshAccessGrant` bypasses it and re-stamps correctly.
- The second half (`grant.subjectId === context.subjectId` + `user:` prefix) closes the remaining gap: for a stamped cookie context every reachable grant already carries that subject, and no cookie human can obtain a grant on *another* human's personal workspace, because there is no `workspace_memberships` row for `getWorkspaceGrant` to find and Slice B only appends the caller's own pointer.

Keying on the context rather than the `Request` is the right call and the reasoning in the doc comment is correct.

**No widening.** `personalWorkspaceOwnerException` has exactly 3 producers, all in `apps/api/src/routes/sessions.ts` / `packages/core/src/application/new-session-drafts.ts`. The Slack App Home caller of `listSessionsForSubject` never passes it, so that path is byte-identical. The resolver can return `true` only when the *scoped* subject's own active organization membership pointer equals the target workspace, so no principal reaches a workspace it could not reach on `main` beyond the intended owner. `getWorkspaceGrant` is untouched.

**The deferred SQL seams have no production caller.** I reproduced this independently, repo-wide, across *all* file types (not just `.ts`), excluding `node_modules`/`.git`. Only: the two migrations, `packages/db/src/session-tenancy.ts` (adapter), `provision-roles.ts`, `runtime-posture.ts`, one prose comment in `organization-membership-lifecycle.ts`, and 14 `packages/db` test files. No route, MCP tool, SDK, React, or web call. `test/session-visibility-contract-surface.test.ts` passes (5 pass / 6774 assertions). Deferring is justified, and pinning them as `PINNED DEFECT` with the exact SQL message and the replacement disjunct is the right call rather than leaving a red test.

**Phase D holds.** The only authority read is `personalWorkspaceId` off the subject's own membership row, filtered by organization and `status='active'`. Nothing from `created_by`, a default workspace, connection attribution, or current access.

**Checks, all on the PR head:** `typecheck` all projects clean · `lint` exit 0 · `format:check` clean · `check:docs-refs` pass · `check:migration-ordinals` pass (no migration added) · `check:public-hygiene` pass.

**Baseline, diffed by name and not by count.** Full `apps/api` suite run on both trees on the same machine:

```
pristine main 6f61d6ee : 1303 pass / 3 skip / 17 fail   (1323 tests, 159 files)
this branch  5727bf1e  : 1319 pass / 3 skip / 17 fail   (1339 tests, 160 files)
diff <(sorted main failure names) <(sorted PR failure names)  →  IDENTICAL
```

Exactly as claimed. I also read into the 17: they are the P1.4 / P3.2 / P4.2 sandbox-viewer, desktop-stream and lease suites, and the P4.2 ones fail on `RangeError: Maximum call stack size exceeded` from a self-recursive `getSessionAuthorityEpoch` mock in `apps/api/test/selfhosted-stream-mint.test.ts:76` — a broken test double, unrelated to this change. Worth noting that one of them is literally named *"a managed PERSONAL-workspace human mints (no membership row by design); suspension degrades"* — I checked, and its redness is the mock recursion, **not** a fourth unfixed personal-workspace seam. It is still red for a dumb reason and someone should fix that mock.

**Organization OWNER — the one principal your matrix omits — is denied.** I ran it myself rather than reasoning about it:

```
attacker organization role                                  = owner
owner → member's personal workspace, GET …/sessions (route) = 403
… listSessionsForSubject   direct seam, flag FORCED true    = SessionListAccessError
… setSessionPin            direct seam, flag FORCED true    = SessionPinAccessError
… saveNewSessionDraft…     direct seam, flag FORCED true    = NewSessionDraftAccessError
```

---

## F1 — MERGE BLOCKER (text): the in-scope resolver's guarantee is vacuous at every current call site, and `AGENTS.md` now says otherwise

`subjectHasLiveWorkspaceAuthorityInScope` compares `current_setting('opengeni.subject_id')` against `input.subjectId` and raises on mismatch. But look at who sets that GUC:

```ts
// packages/db/src/database.ts — withWorkspaceSubjectRls
const context = await rlsContextForWorkspace(db, workspaceId);
return await withRlsContext(db, context, async (scopedDb) => {
  await setSubjectRlsContext(scopedDb, subjectId);   // ← from the caller's own argument
  return await fn(scopedDb);
});
```

and then at all three seams:

```ts
withWorkspaceSubjectRls(db, workspaceId, options.subjectId, async (tx) =>
  … subjectHasLiveWorkspaceAuthorityInScope(tx, { subjectId: options.subjectId, … })
```

The scope is set **from the same request-derived string** the assertion later compares against. `scopedSubjectId !== input.subjectId` is a tautology at `listSessionsForSubject`, `setSessionPin`, and `saveNewSessionDraftInTransaction`; it can never fire in production. Structurally this is the *same* circularity as #1638's — the GUC is written from the caller's argument — just moved one stack frame outward.

The clearest evidence is your own sabotage: **G2 reddens exactly one test**, and that test is the synthetic one that hand-builds a mismatched scope. Not a single route-level test depends on the assertion, because nothing can reach it.

Concretely, this makes the following statements wrong:

- **`docs/organization-tenancy.md` (new section):** *"Two properties keep it owner-only, and both are **structural** rather than conventions."* Property 1 is a convention — it holds only because every caller happens to pass the same subject it scoped with.
- **`AGENTS.md` (new paragraph):** *"it never sets `opengeni.subject_id` and raises if the requested subject differs from the transaction's **authenticated** scope."* Nothing authenticates that scope. `withWorkspaceSubjectRls` takes an arbitrary string.
- **`packages/db/src/workspace-authority.ts`:** the bold heading *"**Owner-only by construction.**"* — the very next sentences correctly walk it back (*"Callers must still keep the exception off a principal's own authorization path"*), which is the honest statement. The heading contradicts it.

Why this is worth blocking on rather than a nit: this repo treats `AGENTS.md` as normative, and the paragraph you added is now the guidance a future author will read before touching this seam. As written, a reasonable reader concludes the in-scope variant is safe to call with a request-derived subject *without* the provenance flag — which is exactly the mistake #1638 was caught making, re-encoded in the project's own instructions. The trap gets *more* dangerous, not less, once the resolver is the blessed one.

To be fair to the design: even an entirely unconstrained oracle here leaks nothing on its own. The only fact it can return is *"subject X owns personal workspace W"*, and that becomes authority only when a caller treats it as *"I am X"*. The `WeakSet` provenance stamp is a genuinely sound boundary for that, and it is the **only** one. So the correct framing is *one* real guard plus a hygiene assertion — not "two independent structural guards".

**Asks:**
1. Reword the `AGENTS.md` and `docs/organization-tenancy.md` sentences to say what is true: the resolver refuses to *name* a third party under a scope the caller did not establish, and *the provenance flag is the boundary*. Drop "structural" and drop "authenticated" (or say "the scope the caller established").
2. Drop or requalify the **"Owner-only by construction"** heading in `workspace-authority.ts`; the paragraph under it already says the right thing.
3. Consider making the assertion non-vacuous later by having the seams pass the *scope-derived* subject rather than re-passing their own argument — then the assertion becomes a real read-back rather than a comparison of a value with itself. Not required for this PR.

---

## F2 — MERGE BLOCKER (test + body): claim 4's justification is factually false, and the pin blesses the wrong status

You asked to be judged hardest here, so:

**The status claim is true.** Pristine `main` really does return **200 / 403 / 200** for a workspace-scoped API key on a personal workspace — my re-derived baseline test on `6f61d6ee` passes that case unmodified. You did not change it.

**The stated reason is false.** The PR body says:

> The owner can mint such a key because `managedPersonalWorkspacePermissions` includes `api_keys:manage`, so this is the owner delegating to their own key.

`managedPersonalWorkspacePermissions` (`packages/db/src/index.ts:1160-1190`) does **not** contain `api_keys:manage`. `AGENTS.md` says so explicitly, in the paragraph immediately above the one this PR edits: the personal-workspace grant *"does not carry `workspace:admin`, `members:manage`, or `api_keys:manage`, so it cannot delegate access to another principal."*

I drove the real route to be sure:

```
POST /v1/workspaces/<personal>/api-keys   owner's own cookie  → 403 missing permission: api_keys:manage
POST /v1/workspaces/<legacy>/api-keys     owner's own cookie  → 201
```

`POST /v1/workspaces/:workspaceId/api-keys` is the *only* production caller of `createApiKey`, and it is gated on `api_keys:manage`. Nobody but the owner holds any grant on a personal workspace (no membership row for `getWorkspaceGrant`, no context grant for anyone else), and the owner's grant lacks the permission. **So the principal the test pins cannot be constructed through any production path today** — the test mints it by calling `createApiKey()` at the DB layer directly.

This cuts both ways.

*Good news:* the thing I went looking for is not there. An account- or organization-admin cannot mint a key on a co-member's personal workspace and read their sessions — I checked, it 403s at `requireAccessGrant`. Main is not leaking.

*Bad news:* the pin now asserts **200 is correct** for a principal whose sole safety property is unreachability, and it does not assert the unreachability. The day someone adds `api_keys:manage` to `managedPersonalWorkspacePermissions`, or adds an account-admin minting path, this "regression guard" goes **green** while blessing exactly the case `AGENTS.md` forbids ("API keys … receive no personal-workspace access through that exception"). A guard that stays green when the hole opens is worse than no guard.

And such a key would genuinely outlive the authority. I checked:

```
key list, membership active            → 200
cookie list, AFTER org membership suspended → 403   (correct)
key  list, AFTER org membership suspended → 200   (0263 suspension does not revoke API keys)
```

Pre-existing, currently unreachable, and not caused by this PR — but it is exactly why `200` should not be recorded as "the verified truth" without its premise.

**Asks:** correct the false sentence in the PR body / test comment, and either (a) assert `403` for that principal, or (b) keep the three-status pin and add its premise to the same test, e.g. `expect(managedPersonalWorkspacePermissions).not.toContain("api_keys:manage")` plus the route-level `403` on the mint. Then the pin fails loudly the day it becomes reachable, which is the only condition under which it matters.

---

## Smaller findings

**F3 (minor).** `saveNewSessionDraftInTransaction` passes caller-supplied `input.accountId` into the resolver, while `listSessionsForSubject` and `setSessionPin` derive it from `accountIdInRlsScope(tx)`. Same seam, two provenances. Prefer the scope-derived value at all three — it removes a caller-controlled input from an authority decision for free.

**F4 (minor).** `packages/db/src/workspace-authority.ts` has an orphaned doc block — the *"The tenant account of the caller's ALREADY-applied RLS scope…"* comment sits immediately above `subjectIdInRlsScope`, which then has its own comment; `accountIdInRlsScope` ends up undocumented. Cosmetic, but this file is now cited by `AGENTS.md`.

**F5 (coverage only — behaviour verified correct).** The denial matrix covers same-org **member** and **admin** but never the organization **owner**, the top of the 0263 ladder. I ran that case myself (results above): denied at the route and at all three seams with the flag forced. So this is purely missing coverage, not a hole — but it is a two-line addition to `inviteIntoOrganization` + the G3 direct-seam loop, and publishing "administrators" without "owners" invites the wrong reading.

**F6 (nit, perf).** On the owner's personal-workspace path the list/pin seams now cost four extra round trips (`current_setting` × 2, a second membership probe, `list_self_organization_memberships`) — and the membership probe runs twice (outer drizzle + the resolver's raw copy). Only on that path, so acceptable; worth folding the duplicate probe out if this ever gets hot.

**F7 (note).** `export * from "./workspace-authority"` publishes both `subjectHasLiveWorkspaceAuthorityInScope` and `accountIdInRlsScope` on `@opengeni/db` (a non-private, published package). Given F1, an external caller trivially satisfies the assertion. Either don't re-export `accountIdInRlsScope`, or carry the caller obligation onto the public surface.

---

## What I could not verify

- Better Auth itself is stubbed in `buildApp()` (`managedAuth.api.getSession` is a fake). The *real* cookie parsing/renewal is not exercised — everything below it (`validateCanonicalHumanSession`, `ensureManagedAccessForUser`, the real 0219/0263 lifecycle, the real Hono routes, the real RLS) is. That is a reasonable seam and the stamp is placed on the OpenGeni side of it, but "a real Better Auth cookie session" in the PR body is slightly generous.
- I did not re-run `packages/core` (359) or the full `packages/db` (1512) suites; I ran the two new files, the contract test, and the full `apps/api` suite on both branches.
- I did not evaluate the interaction with #1638 or #1631 beyond the merge-sequencing note; the conflict analysis in the body looks mechanical and I have no reason to doubt it.

---

## Bottom line

The bug is real, the diagnosis is right, the provenance stamp is a genuinely sound and well-chosen boundary, the sabotage matrix reproduces exactly, and I could not find a principal that gains access it lacked on `main`. **I would stake production access control on the code.**

I would not stake it on the *text*, and on this seam the text is load-bearing: `AGENTS.md` is how the next author learns what is safe here, and as written it asserts a structural guarantee the code does not have — the same class of overstatement that got #1638 caught. Fix F1 and F2 (both are wording plus one test assertion, no design change) and this is a merge.

*Reviewed independently in an isolated clone; not merged, nothing fixed. `--approve` is unavailable to me on my own authorship, so this is posted as a comment — treat it as **request changes**.*

benyhh (commented):
# Confirmation pass - verdict: **APPROVE the code and the seam text. One required PR-body edit before squash-merge.**

Reviewed head **`1a822a42`** (`codex/personal-workspace-session-surface`), in an isolated clone, against **current** `origin/main` **`13c3af4f`** (not the branch's base `6f61d6ee` - `main` moved, and `0302` from #1631 landed). The author has **not** merged `main`; merge-base is still `6f61d6ee`.

**F1 and F2 are genuinely fixed.** I re-derived both from scratch rather than reading the diff for reassurance, re-ran the whole sabotage matrix independently, and re-established the `apps/api` baseline against current `main`. My numbers reproduce the author's exactly. The corrected text now says the true thing in the places that matter, and the F2 pin is load-bearing from **both** directions.

The one thing I am blocking on is text again, and it is in the **PR body**, which this repo turns into the permanent squash commit message: the stated #1638 merge-resolution recipe would silently delete a security fix from #1638. Details in §6. One paragraph.

---

## 1. F1 - is the correction complete?

**Substantively yes.** I grepped every artifact - `AGENTS.md`, all of `docs/**`, both new doc comments, the changeset, the PR body, and every test name/comment - for surviving claims of two guards, "structural", "by construction", "second guard", "defence in depth", or "authenticated" applied to the inherited scope.

Confirmed corrected and accurate:

- `AGENTS.md:93-110` - "**Neither that resolver nor its exported `subjectHasLiveWorkspaceAuthority` sibling is an authorization**", "a **consistency check, not an authorization**: the scope was itself set from a caller-supplied argument", "the resolvers will answer `true` for whatever subject they are given."
- `docs/organization-tenancy.md:158` - "**The resolver is not a second guard.** … At all three call sites both sides come from the same variable, so the check is a tautology there."
- `packages/db/src/workspace-authority.ts:35-55` - "`# This function is NOT an authorization`", "Nothing here authenticates anybody", "this function will happily answer `true` for any subject you scope a transaction to." The `Owner-only by construction` heading is gone.
- `packages/db/src/index.ts:26759` (the `PersonalWorkspaceOwnerException` docstring) - "**This flag is the authorization.** The resolver it gates is not."
- `.changeset/personal-workspace-session-surface.md:12` - "**The authorization is not that resolver.**"
- PR body - "⚠️ That raise is a consistency check, NOT an authorization", "this PR ships **one guard, not two**".

**Read as a future implementer:** yes, this now prevents the mistake. `AGENTS.md` names the exported oracle as an arbitrary-subject oracle, says "never pass it a request-derived subject", and states positively that the *only* thing that authorizes the exception is `canonicalManagedHumanSession`. A new caller reaching for either helper is steered correctly.

### Two residual uses of "authenticated" for the inherited scope

You said you removed every one. Two survive, both in doc comments:

1. **`packages/db/src/index.ts:17475`** - in the docstring you rewrote in `116fa7cb` and did *not* touch in either correction commit:
   > When the subject IS the caller, use `subjectHasLiveWorkspaceAuthorityInScope` instead: it refuses to set the subject GUC and raises unless the subject matches the transaction's **authenticated** scope.

   This is the exact sentence that steers the next implementer toward the in-scope variant, and it re-asserts the thing F1 was about. Mitigated - they land on the in-scope variant's own header next, which is unambiguous - but it is the one place a hurried reader stops.

2. **`packages/db/src/workspace-authority.ts:121`** - "The **authenticated** subject of the caller's ALREADY-applied RLS scope … Used to **prove** that a caller is asking about its own scope."

In fairness: "authenticated subject" is pre-existing repo vocabulary for `opengeni.subject_id` (`database.ts:807`, `setSubjectRlsContext`'s own error string), so neither is an invention of this PR, and both sit next to explicit "not an authorization" warnings. But #1638 is going to rewrite (1) anyway when it takes the rename, so fixing it now costs nothing and removes the last copy.

**Also still open from my last pass** (all were marked minor, none were addressed - flagging only so they are not lost, not asking for them here):
- **F4** - `workspace-authority.ts:112-119` is still an orphaned doc block: two consecutive JSDoc comments mean only the second attaches, so `accountIdInRlsScope` remains undocumented and the "tenant account" prose now sits on `subjectIdInRlsScope`. The correction commit made this worse, because the block that won is the one carrying "authenticated".
- **F3** - `saveNewSessionDraftInTransaction` still passes caller-supplied `input.accountId` to the resolver while list/pin use `accountIdInRlsScope(tx)`. Not exploitable (I checked: a mismatched account can only make the answer *narrower*, since the pointer equality still has to hold), but it is a caller-controlled input on an authority decision for no gain.
- **F7** - `export * from "./workspace-authority"` still publishes `accountIdInRlsScope` on the non-private `@opengeni/db`. `subjectIdInRlsScope` is correctly private.
- Cosmetic: `docs/organization-tenancy.md:145` opens an ordered list that has only item `1.` - the leftover shape of the two-item list the correction collapsed. Renders oddly; make it a plain paragraph.

## 2. Is the retained justification for the in-scope variant true?

Four claims. Three hold; one is stronger than the code on the path this PR actually adds.

| Claim | Verdict |
|---|---|
| Makes the arbitrary-subject oracle shape **unrepresentable** at these seams | **True.** `subjectHasLiveWorkspaceAuthorityInScope` never calls `setSubjectRlsContext`; it has no way to scope a transaction. |
| Keeps the rule to **one implementation** shared with the oracle | **True.** `subjectHasLiveWorkspaceAuthority` (`index.ts:17482`) is now a five-line scope wrapper delegating to it. |
| The raise is a **tripwire** against future divergence | **True**, and correctly labelled as exactly that everywhere now. |
| Keeps the read inside the caller's transaction/snapshot/**advisory fence**, "so it serialises with `removeWorkspaceMember()`" | **Half true.** See below. |

On serialisation, verified against the source rather than the prose:

- `listSessionsForSubject` takes `lockSessionPersonalStateShared` (`index.ts:27352`) and `setSessionPin` takes the exclusive counterpart (`index.ts:27826`) **before** calling the resolver, on key `session-personal-state:<ws>:<subject>`. `removeWorkspaceMember` takes the exclusive lock on the same key (`organization-membership-lifecycle.ts:304` and again inside `workspace_membership_removal_command`, `0278:369`). So for those two seams the claim is exactly right.
- `saveNewSessionDraftInTransaction` takes **no** `session-personal-state` lock - it has `FOR KEY SHARE` on the membership row plus `lockNewSessionDraftIdentity` (`new-session-draft:…`, a different key). The PR body says this correctly ("the draft's `FOR KEY SHARE`"); `docs/organization-tenancy.md:172-173` drops the qualification and reads as though the advisory fence covers all three.
- More to the point: **on the new personal-workspace branch the claim is vacuous at all three seams.** There is no membership row, so the draft's `FOR KEY SHARE` locks nothing, and `removeWorkspaceMember` is a no-op there by construction (your own code comment at `new-session-drafts.ts:141-147` says so). The mutation that *does* revoke this authority is organization-membership suspension/removal - and I grepped: `session-personal-state` has exactly three producers repo-wide (`0278`, `removeWorkspaceMember`, the two seams). The 0263 suspension path takes none of them. So the authority read the exception adds is **not** serialised against the write that invalidates it.

Consequence is small (a narrow TOCTOU on your own personal workspace; verified separately that suspension does deny on the next request). But it is a benefit claim in normative docs that is stronger than the code, on precisely the path this PR adds - the same *class* of thing F1 was, two orders of magnitude smaller. Worth one honest clause rather than removal.

Separately, the differentiator is slightly mis-attributed. `withRlsContext` on a transaction handle opens a **savepoint on the same backend**, so passing `tx` to the exported oracle would also stay inside the caller's transaction and fence. The real reason not to do that is that the oracle leaves `opengeni.subject_id` clobbered for the rest of the caller's transaction - which is exactly #1638's fix, and exactly what §6 is about.

## 3. F2 - is the new pin real?

**Yes, and it is load-bearing from both directions.** The old pin asserted three status codes; the new one asserts the unreachability. I broke it two ways:

| Sabotage | Result |
|---|---|
| `api_keys:manage` added to `managedPersonalWorkspacePermissions` (line-targeted, verified I hit *that* array and not the adjacent one that ends with the same `"artifacts:publish",` line) | **RED** - `the route REFUSES to mint an API key on a personal workspace` |
| Route permission weakened: `apps/api/src/routes/api-keys.ts:22` `"api_keys:manage"` → `"sessions:read"` | **RED** - same test |

Two independent assertions carry it (`expect(posted.status).toBe(403)` and `expect(managedPersonalWorkspacePermissions).not.toContain("api_keys:manage")`), so neither hole can open silently. This is the right fix.

**Your correction to me is confirmed, on current `main`, not just on the branch.** I drove a stripped probe on pristine `13c3af4f`:

```
MAIN-SEAMS               403 403 403     (the bug is still live on current main)
MAIN-MINT                403 missing permission: api_keys:manage
MAIN-PERMS-HAS-API-KEYS-MANAGE  false
MAIN-APIKEY              200 403 200     (unchanged by this PR)
MAIN-SUSPENDED-COOKIE    500             ← your 500, pre-existing
MAIN-SUSPENDED-KEY       200             ← key outlives the authority, pre-existing
```

So the **500** is real and pre-existing on pristine `main`, exactly as you reported. My "403 (correct)" line in the last review was wrong; yours is right.

One nit on the new test: `expect([404, 405]).toContain(minted.status)` for the `PUT` verb asserts nothing useful (a 404 for any reason passes) and `expect(...).not.toBe(200)` right after `toBe(500)` is redundant. Also, pinning the exact `500` will need a touch the day someone converts that raise - the comment says so, which is enough.

## 4. Independent sabotage matrix

Run from a clean tree, one guard at a time, both new suites (`25 pass / 0 fail` green baseline - 18 `apps/api` + 7 `packages/db`). **I did not use your patches**; I wrote my own anchors and verified each landed where it claimed before running.

| Sabotage | My count | Reddened |
|---|---|---|
| **G1** `isCanonicalManagedHumanSession` → `return true` | **2** | delegated bearer w/ substituted `user:` subject; delegated service initiator |
| **G2** in-scope consistency check removed | **1** | the synthetic mismatched-scope test - nothing else |
| **G3** resolver `personalWorkspaceId === workspaceId` dropped | **3** | organization **OWNER**; same-org **ADMIN** (seam); same-org **MEMBER** (seam) |
| **G4** `personalWorkspaceOwnerException` forced true at all 3 seams | **2** | the same two delegated tests |
| **G5** non-`user:` short-circuit removed | **1** | DB-layer-minted API-key pin |
| **G6** `status = 'active'` dropped in the resolver | **1** | suspended organization membership |
| **G7** `api_keys:manage` added to the personal permission set | **1** | route-refuses-to-mint pin |
| **G7b** *(mine)* route permission check weakened | **1** | same pin |
| **Fix removed** (exception never applies) | **7** | 3 positive seams + ADMIN/MEMBER seam + SUSPENDED + same-org co-member |

**7 guards, 11 reddening events, 9 distinct tests.** Byte-for-byte your numbers. Your G1 caveat reproduces too - the organization OWNER and same-org ADMIN do *not* redden under G1 (denied a layer earlier at `accessGrantAuthorization`) and *do* redden under G3 at the seam, so G3 genuinely owns those cases.

On the G7 flaw you disclosed: I hit the same trap and avoided it by inserting at an explicit line number, then printing the surrounding 12 lines to prove I was inside `managedPersonalWorkspacePermissions` and not the array immediately above it that ends with the same `"artifacts:publish",` token. Your retarget was correct.

## 5. Regressions - baseline re-established against **current** `main`

Both trees on the same machine, same shared Postgres, full `apps/api` suite, compared by **sorted failure names**:

```
current origin/main 13c3af4f : 1303 pass / 3 skip / 17 fail   (1323 tests, 159 files)
this branch         1a822a42 : 1321 pass / 3 skip / 17 fail   (1341 tests, 160 files)
diff <(sorted main fails) <(sorted PR fails)  →  IDENTICAL
```

+18 pass = exactly the 18 new tests. Failure set unchanged despite `main` having moved 7 commits, including `0302`.

The correction commits themselves cannot have regressed anything: `git diff 5727bf1e..1a822a42 -- packages apps/api/src packages/core/src`, filtered to non-comment lines, is **empty**. The two commits are docstrings, comments, and tests only.

Other checks on the head: `typecheck` exit 0 (all projects) · `lint` exit 0 · `format:check` clean · `check:docs-refs` pass · `check:public-hygiene` pass · `check:migration-ordinals` pass (`310 migrations, base origin/main highest 0302`, no migration added) · `packages/core` **359 pass / 0 fail** · `test/session-visibility-contract-surface.test.ts` **5 pass / 6774 assertions** - the inertness pin holds. The full `packages/db` suite was still running when I posted this, with **zero** `(fail)` lines across ~100k lines of output; the 7 new tests in it pass in isolation and inside the merged tree.

The two comment-only edits to `sandbox-desktop-stream.test.ts` and `selfhosted-stream-mint.test.ts` are pure prose (I diffed them) and change no behaviour. The diagnosis in them reads correctly - `realDb` is the same namespace object `mock.module` replaces, so that fallback really is self-recursive, unlike the `realGetSandbox` capture above it - and those tests are red identically on both trees, so they are not a fourth personal-workspace seam. Documenting it in place rather than fixing it here is the right call; I did not chase the failure to ground.

## 6. Merge sequencing - **this is the blocking item**

`main` moved but you did **not** merge it, which is right per the moving-main contract. GitHub reports `CONFLICTING`; I resolved it locally and it is **one additive conflict** in `docs/organization-tenancy.md` (your new seam section vs #1631's new `0302` paragraph - keep both, nothing else conflicts; `AGENTS.md` and `packages/db/src/index.ts` auto-merge). No code conflict.

**Post-merge verification (item 6).** I built the merge locally, installed, and re-ran both new suites on it: **25 pass / 0 fail**. Your two `PINNED DEFECT` assertions still describe actual current behaviour after `0302`:

- `0302` replaces only `guard_session_authority_write`; it does not touch `transition_session_visibility` (`0225`) or `fork_session_content` (`0289`).
- In `0225` the `workspace_memberships` requirement (`0225:589-597`) raises **before** the `owner_organization_membership_id IS NULL` check (`0225:616`), so the pinned message `session visibility transition requires active membership` is stable regardless of whether the owner column is now populated. Your note that that seam needs **both** halves is correct.
- `check:migration-ordinals` on the merged tree: `311 migrations, base origin/main highest 0302`. No collision.
- Merged tree is otherwise clean too: `typecheck` all 31 projects, `lint`, `format:check`, `check:public-hygiene`, `check:docs-refs` all pass.

**The problem is the recipe you wrote for #1638.** The PR body says:

> **Resolution: take their name and their docstring, keep this PR's delegating body** — the extracted rule in `workspace-authority.ts` is unchanged either way…

and

> I also fixed nothing in `withRlsContext`: the in-scope variant never sets the subject GUC, so there is no restore to perform and no shared primitive to widen. **The two PRs do not touch the same primitive.**

I pulled #1638 (`1615f26e`) and read it. Its GUC-restore fix is **not** in `withRlsContext` - it is inside the body of `subjectHasLiveWorkspaceAuthority` itself: read `current_setting('opengeni.subject_id')` before the probe, `try { probe } finally { set_config(prior) }`, with a comment explaining that `withRlsContext` restores only account/workspace and that `SET LOCAL` survives savepoint release. That is precisely the body this PR replaces with a delegation.

So the two PRs *do* touch the same lines, and **following your recipe literally deletes #1638's fix.** The regression is real after this PR lands: `subjectHasLiveWorkspaceAuthority(tx, …)` called on a transaction handle nests as a savepoint, sets `opengeni.subject_id` to the probed subject, and leaves it set for the remainder of the caller's transaction.

Nothing in *this* PR is wrong - none of your three seams call the oracle, and its one surviving production caller (`apps/api/src/sandbox/viewer.ts:747`, the 0281 mint) passes a pool handle, so nothing leaks today. #1638's call sites are the ones that need the restore, which is why #1638's author wrote it. But the PR body is the permanent squash commit message, and as written it instructs the next person to drop a security fix. **Ask: replace that paragraph with "keep this PR's delegation *plus* #1638's save/restore around it (the restore belongs in the wrapper, not the extracted body)", and drop the "do not touch the same primitive" sentence.** One paragraph, no code change.

## 7. What I could not verify

- Better Auth itself is still stubbed in `buildApp()`; everything below it is real. "A real Better Auth cookie session" in the body remains slightly generous - unchanged from last time, and the stamp sits on the OpenGeni side of that seam, so it does not weaken the result.
- Other agents were running full `apps/api` suites against the same shared Postgres during my runs. My baseline and PR runs each completed with a stable 17/17 and an identical failure set, so I do not believe contention affected the result - but the transient you recorded in `migration-0241-atomic-personal-resource-delegation` has an obvious cause, and giving that file explicit timeouts really is worth a separate cleanup.
- I did not re-run the `apps/worker`, `packages/runtime`, or web suites; nothing in the diff reaches them.

---

**Bottom line.** The two blockers I raised are genuinely, verifiably fixed - F1's text now says the true thing in every artifact that matters, and F2's pin fails the day the hole opens, which is the only day it counts. I re-derived the sabotage matrix independently and it reproduces exactly, including the caveat. I would stake production access control on the code and on the seam's text.

Fix the #1638 merge-resolution paragraph in the body before squash-merge, ideally sweep the last two "authenticated"-for-scope phrases (`index.ts:17475`, `workspace-authority.ts:121`), and soften the "serialises with `removeWorkspaceMember()`" clause in `docs/organization-tenancy.md` to say what it does on the personal branch. Then merge, and please carry the #1638 follow-up forward - it is still the most important thing this PR does not do.

*Reviewed independently at `1a822a42` in an isolated clone against `origin/main` `13c3af4f`; not merged, nothing fixed. `--approve` is unavailable on my own authorship, so this is posted as a comment.*

benyhh:
Addendum to my confirmation pass at `1a822a42`: the full `packages/db` suite finished after I posted - **1511 pass / 0 fail** across 265 files (771s). That matches your claimed number exactly, so the last open item in my "could not verify" list is now closed. Nothing else in the review changes.

benyhh (commented):
APPROVE exact head a54b55dddb3d76382b30b4d617aead46b3c4aa9f

Independent production security/release review found no blocking defect. The managed-cookie provenance is positively stamped only after Better Auth verification and remains absent for delegated, API-key, service, configured, local, and future unstamped paths; subject equality plus the live organization-membership pointer keeps the exception owner-only at all three session seams. The public named-subject oracle now preserves the caller's prior `opengeni.subject_id`, and its warning/rename are carried consistently through the sole production call site, docs, and tests. The breaking rename is correctly declared as an `@opengeni/db` major at this first landing; `@opengeni/core` remains patch-only.

Verified locally on the exact head: the two authority suites plus the inert visibility/fork contract are 31 pass / 0 fail; all 31 TypeScript projects, lint, format, public-hygiene, docs-reference, and migration-ordinal checks pass. Against current `origin/main` d634e4447bc2b64e312838e01f3d8a6100c50b0e, `git merge-tree --write-tree` succeeds; the only file touched on both sides since merge-base is `AGENTS.md`, in separate non-conflicting sections. GitHub reports MERGEABLE. At review time CI has 26 passing checks, 7 pending, 4 intentionally skipped, and 0 failures; the remaining checks must still finish green before merge.

benyhh (commented):
APPROVE exact head dcfae8dc865c331ebcd445f23d9cf62e8c8ed819

Cumulative exact-head review found no blocking defect. This head is a clean two-parent merge of the previously approved candidate a54b55dddb3d76382b30b4d617aead46b3c4aa9f with current main d634e4447bc2b64e312838e01f3d8a6100c50b0e: its tree hash exactly equals `git merge-tree --write-tree a54b55ddd d634e4447`. Every candidate-owned authority, test, documentation, and changeset file is unchanged from a54b55ddd (the only touched candidate file with main overlap is `AGENTS.md`, whose main edits are in unrelated release/SuperGrok sections).

The positive Better Auth provenance gate, owner-subject equality, live personal-workspace pointer check, named-subject GUC save/restore, dangerous-helper rename, sole production call-site update, denial matrix, and pinned inactive SQL seams all remain intact. The changeset still declares the breaking `@opengeni/db` rename as major at first landing and `@opengeni/core` as patch.

Fresh local verification on dcfae8dc8: 31 focused authority/tenancy/contract tests pass with 0 failures, and all 31 TypeScript projects typecheck. GitHub reports MERGEABLE. At review time exact-head CI has 17 passing, 15 pending, 4 intentionally skipped, and 0 failing checks; do not merge until the remaining checks finish green.

PR #1642, opened 2026-08-19
fix(agent): assert the read gate against the pump's own witness, not fork rate

benyhh:
## Verdict: the read gate is correct. This is a flaky **and ineffective** test.

`overshoot_frame_closes_the_read_gate_end_to_end` fails intermittently on macos-26, including on `main`. A multi-agent investigation instrumented the real pump while the assertion fired: the retention watermark was pinned at **2** (frame 1 sent, frame 2 retained unsent = exactly 2048 bytes drained) for the full window in **every** failing run. The gate had closed correctly.

## Why it flakes

The test infers "gate closed" from "`progress.txt` stopped growing". But **the producer does not park when the gate closes** — it keeps running until it has filled the 64 KiB pipe, roughly 64 more `fork+exec` of `head`. The test latches its baseline at a fixed ~500 ms offset and requires it to be final.

The plateau arithmetic confirms it: `64 + drained_KiB`. Measured directly — drain 1024 → plateau 65, drain 2048 → plateau 66, drain 3072 → plateau 67. The real CI failure was `left: 66, right: 65`: the producer parked exactly where a *correctly shut* gate puts it, and the baseline was sampled one `fork+exec` early.

## The more serious problem

**That assertion never detected the regression it is named after.** Injecting the actual E3 regression (dropping the caught-up conjunct from `may_read`) makes the producer run all 200 iterations *before* the first sample, so both samples read 200 and the equality passes.

Verified directly on this branch's parent:

```
E3 REGRESSION INJECTED
  panicked at crates/opengeni-agent/src/job.rs:1638   <- assert!(parked_at < 200)
```

Line 1638, never 1643. The flaky assertion was also the blind one.

## The fix

Assert against `job.watermark` — the highest seq the pump has appended, i.e. reads off the pipe, independent of the producer's fork rate. A 1536-byte window with ≤1024-byte frames admits at most two in-window frames plus one overshoot, so a closed gate pins it at ≤ 3.

With the regression injected, the new assertion fails **immediately and deterministically**:

```
panicked at job.rs:1658: the pump read past the window: 48
```

`assert!(parked_at < 200)` is kept as the end-to-end producer-throttle proof — slowness only makes it more true, so it cannot flake.

## Sibling test

`exhausted_window_throttles_the_child_end_to_end` has the identical fixed-sample shape with ~3.7x more margin (18 iterations to park rather than 66). It has not flaked in CI but fails under heavier load. Given the same treatment. It exercises the *window-exhausted* path (allowance 0) rather than the overshoot path, so it is unaffected by that particular regression — correctly, it still passes with it injected.

## What was explicitly NOT done

- **`may_read` (job.rs:797) is untouched.** Every writer of `sent_hi`/`allowance` was enumerated; closure is monotone and permanent, not racy.
- The sleeps were **not** simply enlarged. That preserves the same fork-rate assumption and leaves the E3 guard as weak as it is.
- The E3 assertion was not deleted or downgraded.

## Verification (macOS, the failing platform)

- 15/15 under CPU contention
- 211/211 for the crate
- `cargo clippy --all-targets --all-features -- -D warnings` clean
- `cargo fmt --all --check` clean
- Regression injected and confirmed caught, then reverted; tree clean

## Residual uncertainty

- The `gated_at <= 3` bound is empirical, not proven from the protocol — the invariant bounds drained *bytes* (1536 + 1024 = 2560), not frame count. The watermark was exactly 2 in 24/24 instrumented runs, and the pre-existing `assert_eq!(first.body.payload_len(), 1024)` already depends on full-size reads.
- The "1 in 8" failure rate in the original report is not well supported; two confirmed instances were found in CI history. It is load-dependent and bursty rather than a fixed probability.

### DIFF
--- a/agent/crates/opengeni-agent/src/job.rs
+++ b/agent/crates/opengeni-agent/src/job.rs
@@ -1324,19 +1324,26 @@ mod tests {
         assert_eq!(data_total(&frames), 8192, "exactly one window was emitted");
         job.no_frame_for(Duration::from_millis(200)).await;
 
-        // Pipe (64 KiB) + window are full: the producer must be blocked well
-        // short of its 200 iterations, and stay parked while we withhold acks.
-        tokio::time::sleep(Duration::from_millis(300)).await;
+        // Same shape as the overshoot test: assert against the pump's own
+        // watermark rather than the producer's fork rate. An 8192-byte window
+        // with <=4096-byte frames admits two in-window frames plus one
+        // overshoot. This test has ~3.7x more timing margin than the overshoot
+        // one (18 iterations to park rather than 66), so it has not flaked in
+        // CI - but the defect is identical and it does fail under heavier load.
+        let gated_at = job.watermark.load(Ordering::Relaxed);
+        assert!(gated_at <= 3, "the pump read past the window: {gated_at}");
+        tokio::time::sleep(Duration::from_millis(400)).await;
+        assert_eq!(
+            job.watermark.load(Ordering::Relaxed),
+            gated_at,
+            "the pump must stop reading while credit is withheld"
+        );
+
+        // End-to-end: the producer is blocked in write(2), not finished.
         let parked_at = lines(&progress_path);
         assert!(
             parked_at < 200,
             "producer must be far from done: {parked_at}"
-        );
-        tokio::time::sleep(Duration::from_millis(400)).await;
-        assert_eq!(
-            lines(&progress_path),
-            parked_at,
-            "producer is blocked in write(2) while credit is withheld"
         );
 
         // Credit returns: the child un-blocks and finishes all 200 writes.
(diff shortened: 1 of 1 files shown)


benyhh:
Independent post-fix review: **APPROVE** at exact head `dddcc20e2f252547cb8bec11b028cfbab207e26a`.

No line-specific findings.

Evidence:
- The overshoot test now waits for watermark 2 and cannot pass before the retained overshoot exists.
- Both paths prove a stable blocked-write witness using fixed-size builtin producers, avoiding host fork-rate and pipe-capacity assumptions.
- 50/50 focused executions passed under CPU contention.
- The exact E3 mutation (`may_read = allowance > 0`) failed 25/25 under pressure and also failed with a deliberately slowed producer.
- The sibling dropped-allowance mutation failed immediately.
- Waits are bounded; no leaked producer processes were observed.
- A conflict-free synthetic merge with current main retained the candidate test blob and passed both focused tests without source mutation.

CI classification remains separate: relevant Agent CI lanes are green; remaining pending jobs and the inherited stale-base Garage deployment leaf are being tracked before merge.

PR #1652, opened 2026-08-20
fix(web): fence workspace transitions (OPE-205)

benyhh:
## Summary

- add a display-only tenant boundary before a routed workspace can render mutable console state
- clear workspace/session drafts and caches and fence late session-create, GitHub, MCP, and Slack projections across workspace or principal switches
- keep managed sign-out gated until an ambiguous response is reconciled against the authoritative cookie session
- prevent Slack polling from superseding an in-flight Request or Cancel mutation
- expose the active workspace to assistive technology and mark selected organization/workspace menu items

## Authority boundary

This is UI isolation, not frontend authorization. Server grants remain authoritative. The boundary prevents stale state and actions from the prior workspace from appearing under the destination route.

## Verification

- 27 focused transition/race tests pass
- web TypeScript check passes
- targeted oxfmt and oxlint pass
- production web build passes the documented raw/gzip/file-count bundle contracts; the always-loaded transition boundary uses a narrow 2,010 KiB direct-session raw envelope while the 559 KiB gzip cap is unchanged
- current-main merge tree is clean against `fba437f71`

No changeset is required because this changes only the internal web application. No architecture map update is required because no system component, data flow, invariant, or canonical change-area mapping changed.

## Review repair and residual audit

Independent review found and verified repairs for stale operation classes in session creation, MCP, GitHub, Slack, and managed sign-out. A final narrow repair ensures a queued Slack polling tick cannot steal ownership from an in-flight Request or Cancel.

The audit also identified a separate pre-existing class: generic workspace CRUD callbacks can still late-update the global workspace collection across a principal replacement. That principal-generation audit is intentionally the next OPE-205 slice; this PR does not claim it complete. The current review must treat that boundary explicitly and block this slice if the residual cannot safely land as a follow-up.

## OPE-205 scope

This is slice 1: tenant-safe organization/workspace switching and selected-context accessibility. Personal-workspace labeling, organization administration IA, visibility/fork activation, and personal-resource grant UX remain separate reviewed slices.

### DIFF
--- a/apps/web/src/components/rail/switcher-block.tsx
+++ b/apps/web/src/components/rail/switcher-block.tsx
@@ -189,6 +189,7 @@ export function SwitcherBlock() {
       >
         <button
           type="button"
+          aria-label={`Workspace: ${activeWorkspace?.name ?? "none"}. Switch workspace`}
           className="group flex w-full items-center gap-2 rounded-md border border-border bg-surface-2/50 px-2 py-1.5 text-left transition-colors hover:border-border-strong hover:bg-surface-2 focus-visible:outline-none"
         >
           <Avatar size="sm" className="rounded-md">
@@ -255,6 +256,7 @@ function OrgLine(props: {
         {props.orgs.map((org) => (
           <DropdownMenuItem
             key={org.accountId}
+            aria-current={org.accountId === props.activeAccountId ? "true" : undefined}
             onSelect={() => {
               if (org.accountId !== props.activeAccountId) {
                 rail.openOrg(org.accountId);
@@ -317,7 +319,11 @@ function WorkspaceMenu(props: {
       >
         <DropdownMenuLabel className="text-fg-subtle">Workspaces</DropdownMenuLabel>
         {props.workspaces.map((workspace) => (
-          <DropdownMenuItem key={workspace.id} onSelect={() => props.onSelect(workspace.id)}>
+          <DropdownMenuItem
+            key={workspace.id}
+            aria-current={workspace.id === props.activeWorkspaceId ? "page" : undefined}
+            onSelect={() => props.onSelect(workspace.id)}
+          >
             <span className="flex size-5 items-center justify-center rounded bg-surface-3 text-2xs font-semibold">
               {workspaceInitial(workspace)}
             </span>
--- a/apps/web/src/components/workspace-tenant-boundary.test.tsx
+++ b/apps/web/src/components/workspace-tenant-boundary.test.tsx
@@ -1,0 +1,55 @@
+import { afterAll, describe, expect, test } from "bun:test";
+import { GlobalRegistrator } from "@happy-dom/global-registrator";
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+
+import { WorkspaceTenantBoundary } from "./workspace-tenant-boundary";
+
+GlobalRegistrator.register();
+(globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT =
+  true;
+
+afterAll(() => {
+  GlobalRegistrator.unregister();
+});
+
+describe("WorkspaceTenantBoundary", () => {
+  test("never commits the previous workspace surface under a new route", async () => {
+    const transitions: string[] = [];
+    const container = document.createElement("div");
+    const root = createRoot(container);
+
+    await act(async () => {
+      root.render(
+        <WorkspaceTenantBoundary
+          workspaceId="workspace-b"
+          stateOwnerWorkspaceId="workspace-a"
+          prepareTransition={(workspaceId) => transitions.push(workspaceId)}
+        >
+          <button type="button">Workspace A action</button>
+        </WorkspaceTenantBoundary>,
+      );
+    });
+
+    expect(container.textContent).toContain("Switching workspace");
+    expect(container.textContent).not.toContain("Workspace A action");
+    expect(transitions).toEqual(["workspace-b"]);
+
+    await act(async () => {
+      root.render(
+        <WorkspaceTenantBoundary
+          workspaceId="workspace-b"
+          stateOwnerWorkspaceId="workspace-b"
+          prepareTransition={(workspaceId) => transitions.push(workspaceId)}
+        >
+          <button type="button">Workspace B action</button>
+        </WorkspaceTenantBoundary>,
+      );
+    });
+
+    expect(container.textContent).toContain("Workspace B action");
+    expect(container.textContent).not.toContain("Switching workspace");
+    expect(transitions).toEqual(["workspace-b"]);
+    await act(async () => root.unmount());
+  });
+});
--- a/apps/web/src/components/workspace-tenant-boundary.tsx
+++ b/apps/web/src/components/workspace-tenant-boundary.tsx
@@ -1,0 +1,31 @@
+import { useLayoutEffect, type ReactNode } from "react";
+
+import { LoadingPanel } from "@/components/common";
+
+/**
+ * A display-only fence between the routed workspace and mutable console state.
+ * Server grants remain the authority. This boundary only prevents a newly
+ * routed workspace from rendering with the previous workspace's cached UI
+ * state while the provider clears that state.
+ */
+export function WorkspaceTenantBoundary(props: {
+  workspaceId: string;
+  stateOwnerWorkspaceId: string | null;
+  prepareTransition: (workspaceId: string) => void;
+  children: ReactNode;
+}) {
+  const { children, prepareTransition, stateOwnerWorkspaceId, workspaceId } = props;
+  const ready = stateOwnerWorkspaceId === workspaceId;
+
+  useLayoutEffect(() => {
+    if (!ready) {
+      prepareTransition(workspaceId);
+    }
+  }, [prepareTransition, ready, workspaceId]);
+
+  if (!ready) {
+    return <LoadingPanel label="Switching workspace" />;
+  }
+
+  return children;
+}
--- a/apps/web/src/context-principal-transition.test.ts
+++ b/apps/web/src/context-principal-transition.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  beginWorkspaceOperation,
+  beginWorkspaceTransition,
+  invalidateWorkspaceTransition,
+  ownsWorkspaceOperation,
+  settleWorkspaceOperation,
+} from "./lib/workspace-transition";
+
+const contextSource = await Bun.file(`${import.meta.dir}/context.tsx`).text();
+
+function sourceBetween(start: string, end: string): string {
+  const startIndex = contextSource.indexOf(start);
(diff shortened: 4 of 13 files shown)


benyhh (commented):
REQUEST CHANGES exact head 480eb450fb31b76ad3308839625ba7f0a457ad5d

The route-change display fence is a good direction, but the exact head is not yet tenant-safe across all transitions it claims to cover:

1. **Credential/principal changes do not advance the transition identity or clear all workspace-owned state.** `forgetAccessKey`, `handleManagedAuth`, and `handleManagedSignOut` change `accessKeyVersion` but leave `workspaceTransitionIdentity`, `workspaceStateOwnerId`, `pendingCreateAttempt`, manual repository drafts, integration projections, and the async refresh generations intact (`apps/web/src/context.tsx:1118-1168`). If a different principal signs back into the same workspace route, `workspaceStateOwnerId === workspaceId` opens the boundary immediately, so old drafts can render. More seriously, a session create accepted under the previous credential can resolve afterward and still pass `ownsWorkspaceTransition`, then install that old principal's session at lines 998-1008. Please make credential/subject changes use the same complete invalidation seam (including advancing the operation identity even when the workspace id is unchanged), and cover sign-out/sign-in/access-key replacement plus a late old create in tests.

2. **A superseded session create still mutates the destination workspace's operation UI.** The ownership check protects only the success projection. The `catch` always emits the old workspace's failure toast, and the unconditional `finally` clears global `busy` (`apps/web/src/context.tsx:1018-1027`). Thus an A create settling after switching to B can show an A error in B or re-enable B's composer while a newer B create is still pending. Fence error effects by the accepted transition and make busy settlement identity/token-aware; add an A→B overlap test with both late resolve and late reject ordering.

3. **A late MCP rejection leaks across the route fence.** `refreshWorkspaceMcpServers` does not pass the signal into its requests and checks it only after `Promise.all` resolves. If the old request rejects after cleanup, the route catch checks only `isAbortError(error)`, not `abortController.signal.aborted`, and toasts the old workspace's error in the new workspace (`apps/web/src/routes/workspace.tsx:155-165`). Suppress rejected results after the signal/transition is stale and test that path.

Focused transition tests and the web typecheck pass locally, and the exact head is mergeable with current `main` (`d634e4447`), but the new tests exercise only the pure workspace-id transition and the boundary render; they do not cover these root-provider async/credential races. CI still had the browser/visual and real-service jobs in progress when reviewed.

benyhh (commented):
REQUEST CHANGES exact head 8c874e0834bb264a4f6b81f3facbe63b1ec3163c

The three previously reported races are repaired: same-route principal changes now advance the fence and clear drafts/pending creates; old creates cannot install/toast/settle a newer create's busy state; and stale MCP resolve/reject paths are inert. The new focused tests and web typecheck pass locally. The full cumulative transition surface still has blocking late-operation/state gaps, though:

1. **A delayed GitHub mutation can start a fresh old-workspace refresh after B is already displayed.** `disconnectGitHubInstallation(A)` awaits the unlink, then unconditionally mutates the shared selections and calls `refreshGitHub(A, ..., {sync:true})` (`apps/web/src/context.tsx:1168-1192`). If the user switches to B during the unlink, the transition invalidates requests that were already running, but this continuation starts *after* invalidation, obtains the newest `githubRefreshId`, and can replace B's GitHub status/repositories with A's. The same class exists for a delayed manifest response: `startGitHubAppManifestFlow(A)` can still submit the provider redirect after switching to B (`:1151-1165`). Capture the accepted transition/operation identity at invocation and require it before every later UI effect, redirect, and chained refresh; also make `refreshGitHub` itself reject application for a workspace not owned by the current transition. Add delayed A mutation → switch to ready B tests.

2. **Workspace-local Slack request state survives route-param changes.** `WorkspaceShellRoute` keeps `slackAccessRequest`, `slackAccessError`, and `slackAccessBusy` in component state, but does not clear them when `workspaceId` changes. The route component is reused for a new param, so a pending A request restarts its polling through B's `refreshSlackAccess`; if B is unavailable, the `slackAccessRequest || slackAccessError` branch renders A's Slack-access UI/workspace name instead of B's ordinary unavailable panel (`apps/web/src/routes/workspace.tsx:35-149, 190-260`). Reset/fence that local request state on the workspace transition (or key/remount the shell) and test an A pending request followed by authorized and unavailable B routes.

3. **The sign-out failure path does not perform the authoritative session re-read its comment promises.** On an ambiguous `signOutManaged()` failure it restores `previousAuthSession` directly and only increments `accessKeyVersion` (`apps/web/src/context.tsx:1285-1297`). The managed-session effect that calls `fetchAuthSession()` does not depend on that version, so a sign-out that committed server-side but lost its response reopens from cached auth until the access request later fails; it can also leave `authSession` describing the prior user while the refreshed access context describes another cookie principal. Keep the authenticated tree gated and re-fetch the managed session before restoring it, then load access for that exact result. Cover committed-but-response-lost and definitive failure cases.

Current-main merge-tree is clean at `d634e4447`. Live CI is not green: `Impacted E2E shard 3/3` and `Browser and visual acceptance (knowledge)` are failed, while real-service and interaction browser jobs are still running at review time.

benyhh (commented):
REQUEST CHANGES exact head 059f2afc8baa9f2d7d92b9724e60b0845575aab7

The three blockers from review 4980109729 are otherwise fixed: GitHub refresh/manifest/disconnect effects are invocation-bound, Slack render state and late cross-workspace completions are workspace-owned, and failed managed sign-out stays gated through an authoritative cookie-session read. The 24 focused race tests selected for this audit and the web typecheck pass locally.

One current-scope Slack action race remains:

- **The periodic poll can supersede an in-flight user Cancel operation in the same workspace.** `refreshSlackAccess`, `requestAccess`, and `cancelSlackAccess` all call `beginSlackOperation()` and share `activeSlackOperation`. The pending-state interval continues firing while `slackAccessBusy` is true because the polling effect neither checks nor depends on `slackAccessBusy` (`apps/web/src/routes/workspace.tsx:106-279`). Sequence: a pending request starts `cancelSlackAccess` as operation N; before the POST settles, the existing timer starts polling as N+1; the successful cancel then fails `ownsSlackOperation`, so it does not clear the continuation or navigate, and its `finally` cannot clear busy. I reproduced the ownership ordering against the exact helper (`cancelStillOwns: false`, `timerTickOwns: true`). Pause/tear down polling while a user mutation is busy, or use separate poll and mutation lanes, and add a behavior test where the timer fires during delayed cancel/request completion.

Two-repair scope classification: the disclosed generic workspace CRUD late-upsert class is a **separable tracked follow-up**, not an additional blocker for this slice by itself. It is pre-existing, server authorization remains authoritative, and the new invocation primitives can fence it in the next OPE-205 slice without changing this candidate's transition contract. The Slack race above is different: it is introduced by this repair and directly breaks the action-completion fence this exact head adds, so it must be resolved before approval.

Compatibility: merge-tree is clean against current `main` 9530e19b7 with no material overlap in the changed web files. Live CI is not green: `Impacted E2E shard 3/3` and `Browser and visual acceptance (knowledge)` are failed; real-service and interaction-browser jobs remain in progress at review time.

benyhh (commented):
APPROVE exact head 610dab281fb928253c9abc79412d0af6150107b8

The narrow Slack ownership repair is approvable.

- A poll already in flight is superseded by the later Request/Cancel mutation, so its completion is inert.
- A timer callback queued after the mutation begins is synchronously blocked by `slackMutationBusy.current`; it cannot allocate a newer operation or steal ownership.
- Rendering also tears down polling while `slackAccessBusy` is true, and the current mutation alone clears the ref/state busy fences in its owned `finally`.
- The new behavior tests cover delayed Request and Cancel completion with a queued timer tick and assert projection/continuation/navigation eligibility plus final busy/operation settlement. All 26 focused transition tests and the web TypeScript check pass locally.
- The prior GitHub, principal/session-create, MCP, workspace-owned Slack, and authoritative managed-sign-out fences remain unchanged and intact.

The disclosed generic workspace CRUD late-upsert class remains a safely separable, explicitly tracked OPE-205 follow-up: this head does not worsen it, server authority remains authoritative, and the existing invocation primitive can fence those callbacks independently.

Compatibility is clean against requested `main` 9530e19b7 and the newer current `main` 2027bc3e; the merge-tree has no conflicts and current-main changes have no material overlap with these web files.

CI status at approval time: no failures, `Explain change impact` passed, and all substantive source/unit/E2E/real-service/browser/deployment jobs were still in progress. Merge should wait for required green CI.

benyhh (commented):
APPROVE exact head e5f47425cad761593cf51cc49f7ea7bc580acfbe

This is an accurate, narrow bundle-contract update, not a disguised product regression.

- The only delta from approved head `610dab281fb928253c9abc79412d0af6150107b8` is `scripts/check-web-bundle-budget.ts`; the approved `apps/` and `packages/` product trees are byte-identical.
- The raw direct-session limit changes by exactly 4 KiB, from 2,054,144 bytes (2006 KiB) to 2,058,240 bytes (2010 KiB). The documented configured-build measurement of 2,056,813 bytes is 2,669 bytes over the former cap and leaves 1,427 bytes under the new cap.
- I reran the full production web build at this exact head. It passed the budget gate with a local direct-session graph of 2,056,799 raw bytes, 567,530 gzip bytes, and 19 files. Initial graph, per-file gzip, direct-session gzip/file-count, lazy raw/gzip, and CSS gzip caps all remain unchanged and passed. The 14-byte difference from the documented configured measurement is consistent with the script's documented API-base configuration variance.
- The newly always-loaded tenant transition/invocation/accessibility code is intentional for the display fence, while compression remains independently bounded by the unchanged 559 KiB gzip cap. The comment records that rationale and the measured boundary directly.

Compatibility is clean against current `main` `fba437f71e25b118febe593092718936f933fa11`; the merge-tree has no conflicts or material overlap in the approved web code or budget contract.

CI status at approval time: no failures; several real-service/E2E/deployment/native checks have passed, while source/unit, package-and-bundle, remaining E2E/real-service/browser, and native/WASM checks are still running. Merge should wait for required green CI.

benyhh (commented):
APPROVE exact head f0b6cbd3b5084154163b2efc4c4419a689305545

Final exact-head confirmation passes.

- The sole change from approved `e5f47425cad761593cf51cc49f7ea7bc580acfbe` removes the internal issue identifier from the public bundle-budget rationale. No issue key remains in that public comment.
- Product code under `apps/` and `packages/` is byte-identical, and every numeric budget—including the approved 2,010 KiB direct-session raw limit—is unchanged.
- The full exact-head production build passes. Direct-session output remains 2,056,799 raw bytes, 567,530 gzip bytes, and 19 files locally; all initial, per-file, direct-session gzip/file-count, lazy, and CSS limits pass unchanged.
- The wording remains technically complete: it attributes the intentional always-loaded tenant-transition, invocation-fence, and selected-context code without exposing internal tracker metadata.
- Merge-tree is clean against current `main` `fba437f71e25b118febe593092718936f933fa11`, with no material overlap in product or budget-contract files.

CI status at approval time: no failures. Source/type, all unit and impacted E2E shards, interaction/knowledge browser checks, deployment, most real-service shards, and all completed native/WASM receipts are green. Real-service shard 4, recovery, workbench browser, package/bundle, and Windows native receipt remain in progress. Merge should wait for required green CI.

PR #1659, opened 2026-08-20
fix(ci): enforce native step timeout caps

benyhh:
## Summary

- add native numeric `timeout-minutes` caps to the 13 budget-bearing `run:` steps
- consolidate the mutually exclusive browser runtime installers and cap all 3 local Playwright action callers
- replace semantic shell analysis with a YAML-only contract that counts each native step cap once and enforces job headroom
- correct the Playwright `__dirlock` stale-time comment and preserve the existing retry/process-group coverage

## Budget mapping

Profile-wrapper caps: 11, 11, 16, 21, 31, 13, 21 minutes.
Direct Bun caps: 13, 5, 4, 6, 4, 36 minutes.
Playwright callers: 17 minutes each.

Preserved job needed/cap pairs: plan 12/15, source-contracts 28/35, unit 22/30, integration 32/40, e2e 31/35, test-suite 19/30, browser 32/35, packages 39/55, desktop 37/45.

## Verification

- focused workflow/profile/release/artifact contracts: 107 pass
- explicit cap mutation matrix: 7 pass, 132 assertions
- all scripts: 638 pass, 1 intentional skip, 0 fail
- typecheck: all 31 projects clean
- lint, format check, public hygiene, and diff check: clean
- synthetic merge with current main `c22116fa`: focused 107 pass; scripts 638 pass/1 skip; typecheck clean

### DIFF
--- a/.github/actions/playwright-browsers/action.yml
+++ b/.github/actions/playwright-browsers/action.yml
@@ -98,6 +98,7 @@ runs:
         PLAYWRIGHT_ONLY_SHELL: ${{ inputs.only-shell }}
         ATTEMPT_TIMEOUT_SECONDS: ${{ inputs.attempt-timeout-seconds }}
         ATTEMPTS: ${{ inputs.attempts }}
+        KILL_AFTER_SECONDS: "15"
       run: |
         set -uo pipefail
         only_shell=""
@@ -112,7 +113,7 @@ runs:
           # `|| status=$?` guard the first failing attempt aborts the script,
           # so the retry, the warnings, and the error summary never run.
           status=0
-          timeout --kill-after=15s "${ATTEMPT_TIMEOUT_SECONDS}s" \
+          timeout --kill-after="${KILL_AFTER_SECONDS}s" "${ATTEMPT_TIMEOUT_SECONDS}s" \
             bun x playwright install --with-deps ${only_shell} ${PLAYWRIGHT_BROWSERS} || status=$?
           echo "::endgroup::"
           if [ "${status}" -eq 0 ]; then
--- a/.github/actions/playwright-browsers/action.yml
+++ b/.github/actions/playwright-browsers/action.yml
@@ -137,9 +137,13 @@ runs:
     # step. One step is sufficient because it reduces the re-run to a
     # single-browser download, which has never breached the bound - but it does
     # mean a partial set persists for the life of this Playwright version.
-    # A SIGKILLed install can leave a stale __dirlock behind. Playwright only
-    # treats it as stale after ~5 minutes, so baking it into an immutable cache
-    # would tax every later run against this key.
+    # A SIGKILLed install can leave a stale __dirlock behind. Playwright locks
+    # the registry through proper-lockfile passing only retries/onCompromised/
+    # lockfilePath, so the default `stale: 10000` applies - the lock clears
+    # after ~10 seconds, not the ~5 minutes an earlier revision of this comment
+    # claimed. Removing it is therefore cheap hygiene rather than a large
+    # saving: it keeps a lock out of an immutable cache entry and avoids the
+    # onCompromised path entirely.
     - name: Drop stale Playwright dirlock before saving
       if: ${{ always() && steps.restore.outputs.cache-hit != 'true' }}
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,7 @@ jobs:
           bun-version-file: .bun-version
 
       - name: Install exact dependency tree
+        timeout-minutes: 11
         env:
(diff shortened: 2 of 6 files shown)


benyhh:
Independent review result: **REJECT** at exact head `348de2bf59bde6155341039ea2d8f569a52a047e`.

Both non-author reviews confirmed the same two blockers:

1. **P1 — assembled/ANSI timeout spellings evade the cap-required lint.** Executable forms including `time$''out`, `$'time\x6fut'`, `time${EMPTY}out`, and short static concatenations return no violation when uncapped. The repair must remain lexical and conservative: ambiguous spelling composition should require a native cap, not be semantically evaluated.
2. **P1 — Playwright bound is not structurally tied to runtime wiring.** Mutating the environment mapping or replacing `${ATTEMPT_TIMEOUT_SECONDS}s` in the command leaves the calculated 15.25-minute bound and current tests green. A 17-minute caller cap could then terminate the composite before its cleanup/cache-save path.

The exact 16-cap mapping, all nine job sums, browser-call consolidation, process-group/retry proofs, current-main compatibility, and hosted CI otherwise passed; hosted CI was fully green (35 successful checks, nine intentional skips).

A scoped repair is being dispatched.

benyhh:
Independent post-repair review: **REJECT** at exact head `593c9a6e6db64dd251f2c9f0b7aeb674d1a7f910`.

Both reviewers confirmed the remaining source blocker:

- **P1 — multi-marker/fully encoded assembly bypasses the one-marker matcher.** Executable uncapped forms such as `ti$''me$''out`, `ti${A}me${B}out`, repeated command substitutions, fully/repeated ANSI-C encodings, and character-by-character `cmd+=...` construction yield no violation.
- **P2 — matcher quality.** Harmless fragments can be flagged, and the overlapping ANSI regex showed avoidable backtracking on malformed input.

The prior Playwright P1 is fixed: declared duration/attempts/grace are structurally wired through the environment into the exact runtime command, and literal/env/command mutations fail.

Repair revision 2 will replace the one-marker/generated-regex approach with a linear, bounded lexical ambiguity scanner that handles arbitrary marker chains, requires token boundaries, and performs no shell evaluation.

benyhh:
Independent final-revision review: **REJECT** at exact head `cbcf1aab4c061603fed2a0b014460e2c2eb7ee6d`.

Remaining P1 fail-open classes:

- legacy backtick substitution such as an executable name assembled around `true` in backticks;
- nested parameter/command substitutions such as `ti${A:-${B}}meout` and nested `$(...)`;
- character append chains separated by `&&`.

Bash resolves each to executable `timeout`, while the uncapped contract returns no violation. The fixed bitset matcher itself is linear and bounded, but its lexical event recognizer is incomplete; exact megabyte-scale performance and all prior probes otherwise pass. Playwright wiring remains approved.

Two substantive repair revisions are exhausted. Per repository policy, patching stops here and a scope review will decide between completing a formally bounded recognizer, exact uncapped-corpus fingerprint governance, or broader native-cap coverage.

benyhh:
Second incident/scope review: **close and replace; do not attempt another lexical repair**.

The scanner cannot soundly recognize arbitrary executable shell identity without becoming a shell interpreter. The remaining backtick/nested-expansion/append gaps are only the latest examples; arrays, sourced scripts, `printf -v`, aliases/functions, decoded strings, and interpreters create the same obligation.

The replacement design is parser-free source-change governance:

- preserve the approved 16 native step caps, exact 13+3 mapping, nine job-budget sums, browser-call consolidation, Playwright duration/grace/attempt wiring, retry/process-group coverage, and dirlock correction;
- remove the shell spelling scanner entirely;
- make canonical parsed-graph SHA-256 digests authoritative for all 20 workflow files and both local composite actions;
- retain a subordinate uncapped-run fingerprint set for exact diagnostics, never as the authority;
- reject dynamic/nonpositive caps, YAML aliases/duplicate keys, missing or escaping local targets, new workflows/actions, and any semantic graph drift until the baseline is explicitly reviewed and refreshed.

Why the full graph is authoritative: a capped producer, workflow input, matrix, job output, `needs`, environment, artifact, or local action can change what an unchanged uncapped consumer executes. Per-run fingerprints alone miss those dataflow changes. Canonical full-file digests catch them without evaluating Actions expressions or shell.

Current scope is 20 workflows, 54 jobs, and 413 steps on live main. Capping every remaining step would require roughly 396 additional calibrated caps at the candidate shape and would break the existing additive headroom invariant in most jobs; the digest gate instead adds about 22 small file records plus diagnostic fingerprints. Historical analysis shows semantic workflow/action changes in about 8.5% of main commits, averaging 1.65 digest rows refreshed per affected commit.

PR #1659 is retained as incident evidence at exact rejected head `cbcf1aab4c061603fed2a0b014460e2c2eb7ee6d`. A fresh replacement will start from current main and receive independent exact-head review.

PR #1667, opened 2026-08-20
ci: bind workflow execution graph budgets

benyhh:
## Summary

- add native step caps to the 13 budget-bearing run steps and 3 effective Playwright action callers
- consolidate browser installation into one lane-selected action call while preserving retry, process-group, cache, and dirlock behavior
- replace shell interpretation with a canonical restricted-YAML execution-graph manifest for 20 workflows, 2 local composite actions, and 197 identity-bound uncapped run fingerprints
- fail closed on malformed or identity-ambiguous YAML, invalid local targets, graph drift, cap drift, and stale manifest inventory without emitting command or environment content

## Cap mapping

Profile-wrapper run caps: 11, 11, 16, 21, 31, 13, 21 minutes for plan dependency install; source typecheck; source guards; unit shard; integration shard; e2e shard; package build.

Direct Bun run caps: 13, 5, 4, 6, 4, 36 minutes for workspace capture; recovery regressions; session-pin browser; responsive knowledge browser; workbench browser; desktop image e2e.

Playwright action callers: 17 minutes for e2e runtimes, lane-selected browser acceptance runtimes, and packed WASM Chromium proof.

## Verification

- focused consumers: 109 pass, 0 fail
- all scripts: 646 pass, 1 intentional skip, 0 fail
- typecheck: all 31 projects clean
- lint, full format check, public hygiene, diff check: clean
- graph manifest: 20 workflows, 2 actions, 197 uncapped runs
- current main f51adf857 compatibility: clean patch application, 646 pass, 1 skip, 0 fail, all 31 typecheck projects, lint, format, hygiene, and graph checks green


benyhh:
Independent exact-head review: **REJECT** at `18b24353c4d1c525b6ce024177087ff15b60ae5c`.

Blocking findings:

- **P1 — non-injective mapping canonicalization.** `canonicalValue` copies mappings into `{}`. An own YAML key named `__proto__` invokes the inherited setter and disappears. Changing `steps[].env.__proto__: before` to `after` produced identical authoritative workflow and secondary run digests with no violation.
- **P1 — lossy numeric canonicalization.** YAML numerics are converted to JavaScript `Number`; distinct values such as `9007199254740992` / `9007199254740993` and `0.1` / `0.10000000000000001` produced identical digests. Scalar preservation therefore fails.
- **P2 — unbounded parse/canonical cost.** There are no explicit source-byte, node-count, mapping-width, or recursion-depth bounds. A valid restricted ~2.5 MB / 100,000-key mapping remained active beyond 41 seconds until terminated.

The replacement architecture otherwise passed review: restricted YAML/local-target/cycle checks, authoritative comparison control flow, non-disclosing diagnostics, exact 20 workflows + 2 actions + 197 uncapped runs, exact 13+3 caps and nine sums, static cap validation, Playwright wiring/retry/cleanup, 109 focused tests / 1,144 assertions, frozen install, and clean current-main compatibility. The only current-main overlap is a clean `bun.lock` merge.

Hosted source/type CI is green. The browser/E2E failures observed on this head are untouched application-test waits (Codex reset button and Slack identity link), not graph-source failures.

Repair scope is bounded: null-prototype/entry-based mapping serialization, lossless or fail-closed numeric handling, and explicit pre-parse/structural limits with boundary tests. No shell parsing is reintroduced.

benyhh:
Second independent review confirms the rejection and adds one separate **P1**:

- `actionDefinitionPaths()` ignores symlinked `action.yml` / `action.yaml`. A referenced directory containing regular `action.yaml` plus `action.yml` symlinked to another payload was inventoried as only `action.yaml` with zero violations, while the Linux runner resolves `action.yml` first and follows the symlink. Workflow/action source discovery must reject symlinks and dual-manifest shadowing fail-closed.

It independently reproduced the `__proto__` and numeric digest collisions and measured wide-map structural cost (1k keys ~50ms, 20k ~2.2s). This symlink fix is included in the same bounded canonicalization/resource repair.

benyhh:
Independent re-review: **REJECT** at exact repaired head `7d382a2a8c3f2be01b6feb80ce9b954c8523f59b`.

Remaining bounded enforcement-order defects:

- **P1 — invalid UTF-8 collision.** `readFile(..., "utf8")` replacement-decodes malformed bytes. Two workflows differing only by raw byte `0x80` vs `0x81` loaded with zero violations and the same graph SHA. Read bytes and decode with a fatal UTF-8 decoder before YAML.
- **P1/P2 — limits after allocation.** The loader reads all up-to-64 sources concurrently before per-source/total byte checks. A 32 MiB workflow was fully returned despite the 256 KiB configured cap. Preflight regular-file sizes and total size, then use bounded/sequential reads and verify actual bytes.
- **P2 — entry limits after allocation.** `readdir` materializes complete directory arrays before the 128/1,024 entry checks. Use streaming `opendir` and stop at limit+1.

The previous own-key, numeric, AST/node/width/depth, and symlink-shadowing blockers are fixed and passed exact boundary replay.

There is also genuine current-main incompatibility: main `3999dd5cf319a2c869e6506816ade33a757fa162` added `.github/workflows/publish-desktop-image.yml`. The synthetic merge is conflict-free but correctly fails because the authoritative manifest lacks that 21st workflow. This is material graph drift, so the final repair will incorporate current main and refresh the manifest/count (21 workflows, 2 actions, 197 uncapped runs) rather than treating it as harmless base movement.

This is repair revision 2 of 2 for PR #1667. The next exact-head review is decisive under the repository delivery policy.

benyhh:
Decisive repair-2 review: **REJECT** at exact head `1fa9cf975f6f903422edf3abb54a55d764489714`.

**P1 — parent-directory replacement bypasses repository confinement.** Discovery validates root metadata and static Dirents, while source preflight/final checks address the final file path and therefore follow a parent directory that is replaced by a symlink after discovery.

Deterministic reproductions in a disposable checkout:

- after workflow discovery closed, rename `.github/workflows` and replace it with a symlink to an outside directory containing the same `x.yml`; the loader returned outside bytes under `.github/workflows/x.yml` with no violation;
- replace `.github/actions/probe` with an outside symlink during discovery; outside `action.yml` loaded under the in-repository action identity.

All other historical blockers now pass: fatal malformed UTF-8, prototype-shaped keys, safe numeric domain, static symlink shadowing, pre-read per-file/aggregate bounds, sequential bounded reads and final-file replacement races, streamed entry limits, AST/source/count/width/depth limits, exact 21 workflows / 2 actions / 197 runs, exact caps/sums, and Playwright wiring. Hosted CI has no candidate-source failure.

This exhausts two substantive repair revisions on #1667. Per repository policy, source mutation stops for a scope review comparing: (A) explicit canonical-root/component identity fences around the existing loader, (B) opening/snapshotting sources while directory identities remain anchored, and (C) immutable Git-object/index-backed graph input. No third blind patch will be made.

benyhh:
Post-repair scope decision: authorize **one final frozen completion**; do not accept current head.

Chosen boundary: immutable Git-tree input for CI plus quiescent worktree input for local development.

- CI captures one exact tree OID, inventories workflow/action entries and Git modes from that tree, preflights blob sizes, then reads workflow/action bytes and the committed manifest by object ID. Worktree parent swaps cannot alter those bytes.
- The authoritative graph command runs immediately after `bun install --frozen-lockfile --ignore-scripts`, before changeset, typecheck, guards, tests, or other PR-controlled commands. The current unit-shard test remains regression coverage, not the sole authority.
- Local default and `--write` retain worktree mode so unstaged/untracked edits, additions, removals, and renames are visible; local concurrent filesystem mutation is explicitly unsupported.
- No `/proc`, `/dev/fd`, native FFI, `openat`, or platform-specific directory-descriptor assumptions.

Why not more ancestry sampling: realpath/dev/inode fences catch persistent swaps but retain synchronized swap-read-restore windows. Git objects give CI one immutable, portable snapshot; the dual local mode keeps useful developer behavior.

The completion is frozen to Git-tree enumeration/mode checks, same-tree manifest loading, preflight/bounded/fatal-UTF8 blob reads, quiescent CI ordering, tests for post-capture worktree swaps and dirty local behavior, and the existing exact 21 workflows / 2 actions / 197 runs plus 13+3 caps. To preserve the 197 count, the graph command may be sequenced inside the existing dependency-install step as its immediate post-install command rather than adding a new uncapped run definition.

If this candidate expands beyond that boundary or fails exact-head review, PR #1667 will be closed/replaced.

benyhh:
Frozen exact-head review rejected `20941f4db86d17c78b252714a96cd569a31e4c96`.

Blocking correctness issue: the Git-tree backend does not disable Git replacement objects. A repository-local `git replace <captured-tree> <replacement-tree>` causes the checker to report the captured original tree OID while Git plumbing returns the replacement workflow and manifest bytes; the candidate accepted those substituted bytes with zero violations. That breaks the central single-immutable-tree identity claim.

The bounded fix is clear: run every Git plumbing command with replacement processing disabled (for example `GIT_NO_REPLACE_OBJECTS=1` or `git --no-replace-objects`) and retain a deterministic regression proving an installed replacement ref cannot alter inventories, blobs, or the manifest for the captured tree.

Per the frozen-completion decision, this PR will not receive another repair revision. I am closing it and will carry the approved graph/cap implementation into a fresh replacement PR with this invariant fixed and independently reviewed.

PR #1668, opened 2026-08-20
test(web): synchronize Codex account detail acceptance

benyhh:
## Summary

- synchronize Codex account-detail expansion with the asynchronously rendered account rows
- use the same wait/show-or-hide/open/hide sequence for desktop, mobile, recovery, and completed-redemption contexts
- add a deterministic delayed-row route gate that reproduces the prior timeout without increasing timeouts or weakening assertions

## Root cause

The settings heading renders before the account rows. Several browser-test paths immediately snapshotted `isVisible()` for the account-detail toggle, skipped expansion when the row had not appeared yet, and then waited for controls that remain hidden inside the collapsed row. This produced the alternating “Redeem Full reset” and durable-completion timeouts seen on clean main and OPE-205 PRs.

## Verification

- exact delayed Codex overview browser E2E: 4 consecutive full runs green in independent review
- focused Codex route/security/web/surface tests
- all 31 TypeScript projects
- CI impact contracts
- full lint and format
- public hygiene and diff check

Test-only change; no product behavior or timeout budget changed.

### DIFF
--- a/test/e2e/codex-overview.e2e.ts
+++ b/test/e2e/codex-overview.e2e.ts
@@ -1,7 +1,7 @@
 import { afterAll, beforeAll, describe, expect, test } from "bun:test";
 import { mkdir } from "node:fs/promises";
 import AxeBuilder from "@axe-core/playwright";
-import { chromium, type Browser, type Page } from "playwright";
+import { chromium, type Browser, type BrowserContext, type Page } from "playwright";
 import type { AccessContext } from "@opengeni/contracts";
 import {
   claimCodexResetRedemption,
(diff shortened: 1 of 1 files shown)


benyhh (commented):
APPROVE — independently reviewed exact head `b28e1b1481d66051a5cf55c1674da4eb007a41db`.

The parent reproduced the defect: mobile, recovery, and completed contexts took a non-waiting `isVisible()` snapshot before the asynchronously loaded account row existed, then waited on controls available only inside expanded details. This test-only candidate extracts the already-safe desktop sequence into one helper: wait for the actual Show-or-Hide toggle, click/tap Show only when collapsed, and wait for Hide as the deterministic expanded-state proof. Every browser context installs the first-account-list route gate before navigation, proves the settings heading can render with zero account articles, releases the request, and uses the helper; all four contexts and all account-expansion paths in this test are covered.

No timeout constants were increased, assertions weakened, product code changed, retries added, or same-class expansion path omitted. Exact CI-style real-Postgres Chromium run passed 1/1 with 36 assertions, plus three consecutive delayed stress reruns (and one initial standalone run). Full workspace typecheck (31 projects), impact tests (35/35), full lint, full format check, public-repository hygiene, and diff check passed. Worktree was clean. Current `main` (`f51adf8578c7a364454faa8714f07faadf70d3a2`) is the exact parent, so current-main compatibility is direct and conflict-free.

PR #1691, opened 2026-08-21
fix(sdk): accept standard fetch responses

jorgensandhaug:
## Summary

- export a standards-compatible FetchResponse type that excludes Bun-only Response.textStream
- use the structural response type throughout the SDK fetch boundary, including public artifact downloads
- add type and runtime coverage for standard/Expo-like responses, streaming, JSON/error metadata, cancellation, and AbortSignal propagation
- add an SDK patch changeset

## Compatibility evidence

Peloton was tested on Bun 1.4.0 with React held at 1.0.1 and this SDK package overlaid. Native package, mobile, web, API, worker, api-client, and agent-chat-core typechecks all passed. A clean Bun install also resolved React 1.0.1 and the maintained SDK to one overridden SDK copy, so no React/timeline upgrade is required for adoption.

## Validation

- SDK suite: 402 tests, 1,656 assertions, 0 failures
- all 31 TypeScript projects
- formatting and git diff hygiene
- all 26 publishable packages built
- publish closure guard
- release-shaped publish consumer suite
- runtime embedding consumer suite

The candidate remains mergeable with current main; the intervening session-pagination commits only change an SDK error comment and do not invalidate this patch.

### DIFF
--- a/.changeset/clean-fetch-response.md
+++ b/.changeset/clean-fetch-response.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/sdk": patch
+---
+
+Accept web-standard custom fetch responses without requiring Bun's non-standard `Response.textStream()` extension.
--- a/packages/sdk/src/artifact-client.ts
+++ b/packages/sdk/src/artifact-client.ts
@@ -1,4 +1,5 @@
 import { OpenGeniClient as OpenGeniCoreClient } from "./client";
+import type { FetchResponse } from "./client";
 import type {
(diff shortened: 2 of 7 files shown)


jorgensandhaug:
@davletd All required checks are green on exact head `bac00c025a5b8043c0b66b2b541ed874d811104d`. This is ready for the required review: it narrows `FetchLike` to the exported web-standard structural `FetchResponse` contract, removing the accidental dependency on Bun-only `Response.textStream()` while retaining SDK response behavior and coverage.

PR #1692, opened 2026-08-21
fix(dev): bound sandbox Docker disk usage

benyhh:
## Summary

- always build the Docker sandbox from the current worktree context into an isolated worktree tag
- serialize image maintenance with a lifetime-held kernel lock and protect live worktrees with process-incarnation leases
- cap the dedicated OpenGeni BuildKit cache before and after every build attempt, stop its builder afterward, and retain only a bounded recent OpenGeni image history
- never globally prune Docker or remove unrelated images, volumes, caches, or container-held images

## Why

Repeated `bun run dev` sandbox builds accumulated multi-gigabyte source images and BuildKit cache until Docker Desktop filled the workstation disk. The cleanup must remain source-correct for dirty worktrees and safe across parallel worktrees.

## Verification

- independent exact-head adversarial review: APPROVE
- 25 focused disk-policy, failed-build, PID-reuse, lease-replacement, two-contender, and dev-stack contract tests
- all 31 TypeScript projects
- repository lint and format check
- public hygiene, shell syntax, and diff checks
- clean merge-tree with current `main`

## Local cleanup performed

Three stale generated OpenGeni source tags were retired during validation. Docker volumes and unrelated images/caches were not touched; removed images can be rebuilt by `bun run dev`.


benyhh:
APPROVE — exact head 3956f6cfd0c1e75306cb92e59f4e3de8e0c26ca5, base ac0e86bd70ceeddb736b20dbf23c36c89f2dd0b0.

Cumulative independent review found no residual blocker. Dirty relevant Docker context always rebuilds; worktree-specific image tags are protected by exact live leases; PID reuse is fenced by a random process-incarnation token checked against current argv; maintenance is serialized by a lifetime-held kernel flock; the fixed OpenGeni-only BuildKit builder converges its cache before and after success/failure and is stopped; retention removes only owned source tags or labelled dangling development images while preserving live leases, container-held images, and foreign Docker state.

Verification: focused runtime/concurrency suite 25 passed, 0 failed; shell syntax, 31-project typecheck, lint, format, public hygiene, and diff checks passed. Two-contender builds were non-overlapping; wrong-token/reused-PID and exited-owner leases were rejected. Remote head equals the reviewed local candidate, the PR is mergeable, and the current-base merge tree is clean (only cumulative docs overlap).

PR #1693, opened 2026-08-21
Fix private session creation and simplify access UI

benyhh:
## What changed
- include the explicitly selected session visibility in exact new-session draft reconciliation, preventing a valid private/workspace draft from creating an empty session shell and then returning a false 409
- keep personal resource discovery available for the picker while hiding optional authority loading/errors unless a personal Rig or Variable Set is actually selected
- replace the large session access banner with compact `Only me` / `Workspace` state and contextual `Share` / `Private copy` actions
- replace internal personal-resource authority wording with user-facing copy

## Root cause
The web persisted `visibility` in the durable new-session draft, but core omitted it from the exact snapshot used when consuming that draft. The shell transaction therefore committed, the first-turn transaction rejected the otherwise-valid draft as changed, and the UI received a 409 while an inert session appeared later in the rail.

## Verification
- real PostgreSQL draft/create regression (session plus initial turn)
- focused core/hook/access tests: 23 pass
- full `apps/web/src`: 839 pass
- Chromium personal-resource/accessibility: 8 pass
- all 31 TypeScript projects
- lint, repository format check, public hygiene, diff check
- exact production web build, precompression, and all bundle budgets

## Deliberate boundary
Codex subscription connections are currently workspace-scoped in the backend. Personal-vs-workspace Codex ownership and a create-session selector require a separate authority/allocation/session-snapshot change; this PR does not imply that capability already exists.

### DIFF
--- a/apps/web/src/components/personal-resource-attachment-control.tsx
+++ b/apps/web/src/components/personal-resource-attachment-control.tsx
@@ -17,8 +17,8 @@ const MODE_OPTIONS = [
   },
   {
     value: "always" as const,
-    label: "This workspace",
-    description: "Keep standing authority for work you start in this workspace.",
+    label: "Remember here",
+    description: "Make it available for future work you start in this workspace.",
   },
 ];
 
@@ -54,7 +54,7 @@ export function PersonalResourceAttachmentControl(props: {
       <legend className="px-1 text-xs font-semibold text-fg">
         <span className="inline-flex items-center gap-1.5">
           <KeyRoundIcon className="size-3.5 text-fg-subtle" aria-hidden />
-          Personal resources
+          Your resource access
         </span>
       </legend>
       {controller.loading ? (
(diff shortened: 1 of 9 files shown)


benyhh:
Independent exact-head review: **APPROVE** `064819fb5b3c78dbc4179907fb138404c64b8b54`.\n\nEvidence: traced the reported 409 to draft options containing explicit visibility while the create snapshot omitted it; this head includes visibility only when the raw create request explicitly supplied it, so current workspace/private drafts compare exactly while legacy drafts with no visibility remain compatible with the schema default. The real-Postgres regression creates a session with a non-null initial turn. Catalog discovery still runs to populate personal Rig/Variable Set choices, but optional load/error UI is suppressed only while no fixed personal resource is selected; selected-resource load/error/truncation and `requiresDecision` fences remain fail-closed. Access UI is compact and semantically clearer (`Only me`/`Workspace`, `Share`, `Private copy` only from shared sessions), with preserved managed-owner gating, epoch reconciliation, exact-key retry, live announcements, keyboard focus order, wrapping, and screen-reader state detail.\n\nIndependent verification on this SHA: 23 focused core/hook/tenancy tests passed; 8 Chromium personal-resource/accessibility tests passed with zero axe violations in the exercised attachment surface; production web build, precompression, and bundle budgets passed. No code edits made during review. The separate personal-vs-workspace Codex credential authority/allocation feature is correctly not implied by this PR.\n\nGitHub does not allow this account to formally approve its own PR, so this exact-head approval is recorded as a review comment.

PR #1697, opened 2026-08-21
fix(web): keep active session unread state accurate

benyhh:
## Summary
- acknowledge the active foreground chat again when later durable output arrives, without requiring navigation away and back
- bind each acknowledgement to the exact coalesced event frontier rendered by the browser so unseen newer events remain unread
- preserve same-tab rail reconciliation, live-tip/history/focus fencing, and retry one transient failure without looping on permanent errors
- add the public optional read-through sequence contract plus real-Postgres, contract, unit, and browser regressions

## Verification
- independent review: APPROVE exact `b6d3e36891d4e6feefcaae00f45a0fd140ea5aab`
- 33 real-Postgres/contract tests
- 847 web tests
- 9/9 session-pin Chromium E2E tests
- all 31 TypeScript projects
- production web build and bundle budgets
- lint, format, public hygiene, CI impact, changeset status, diff check

## Bundle
- initial raw: 1,499,526 / 1,500,160 bytes
- direct-session raw: 2,083,239 / 2,083,840 bytes
- direct-session gzip: 579,514 / 579,584 bytes
- all other caps unchanged and green

### DIFF
--- a/.changeset/exact-session-read-frontier.md
+++ b/.changeset/exact-session-read-frontier.md
@@ -1,0 +1,7 @@
+---
+"@opengeni/contracts": patch
+"@opengeni/db": patch
+"@opengeni/sdk": patch
+---
+
+Allow foreground session readers to acknowledge an exact rendered event sequence so later unseen events remain unread.
--- a/apps/api/src/routes/sessions.ts
+++ b/apps/api/src/routes/sessions.ts
@@ -1493,9 +1493,10 @@ export function registerSessionRoutes(app: Hono, deps: SessionRouteDeps): void {
     }
   });
 
-  // Personal follow-up state. Viewing a session never acknowledges it; the
-  // member must explicitly mark it read/unread, and the actively-working label
-  // remains independent of that acknowledgment.
+  // Personal follow-up state. Session reads stay side-effect free; clients use
+  // this explicit mutation after a deliberate read/unread action. The managed
+  // web console treats an exact foreground chat view as that action, while the
+  // actively-working label remains independent of acknowledgment.
   app.put("/v1/workspaces/:workspaceId/sessions/:sessionId/attention", async (c) => {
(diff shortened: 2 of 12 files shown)


benyhh:
APPROVE — cumulative independent review bound to exact head b6d3e36891d4e6feefcaae00f45a0fd140ea5aab.

Verified end to end:
- The managed console acknowledges only the exact foreground, fully loaded live-tip event frontier; background tabs and historical windows do not consume unread state.
- The hook frontier uses the coalesced-aware `lastSequence`; API/contract/SDK carry `acknowledgedThroughSequence`; DB clamps it to the current session sequence and advances the personal acknowledgement monotonically, so later unseen events remain unread.
- Retry behavior is bounded to one automatic retry per exact projection key. A second failure leaves the receipt latched, preventing permanent 4xx/5xx request loops; a genuinely newer event frontier gets its own bounded attempt.
- The landed same-tab rail projection from #1694 is preserved, with revision and last-sequence ordering preventing stale polls or responses from resurrecting or prematurely clearing the unread dot.
- Workspace/principal transition fencing and foreground checks remain intact.
- Current-main compatibility is proven: base 96e52c902b8343a348ec9c17738a905a65471340 is an ancestor, merge-tree is clean, and diff-check passes.

Independent verification: 41 focused web/contract/real-PostgreSQL FORCE-RLS tests passed during the cumulative review; all 31 TypeScript projects passed; the focused Chromium retry/frontier case passed; and an exact-head production build passed every bundle budget (initial raw 1,499,526/1,500,160; direct raw 2,083,239/2,083,840; direct gzip 579,514/579,584). No blocking correctness, UX, concurrency, retry, compatibility, or bundle issue remains.

GitHub does not permit this account to submit a formal approval on its own pull request, so this explicit APPROVE comment records the independent head-bound verdict.

PR #1696, opened 2026-08-21
fix(react): show structured human-input answers

benyhh:
## Summary

- keep pending structured human-input questions in the main chat timeline and retain the resolved `Agent asked` / `You answered` exchange after submission
- support compact multi-question requests with consistent numbering, accessible controls, and an inline `Other` free-text path for every choice question
- preserve exact durable answers across runtime, database, realtime, Slack, SDK/React, pagination, and child-session lifecycle boundaries
- label the final action clearly as `Send answers`

## Verification

- semantically merged with current `origin/main` at `d8ba09d8e748fb57328df413725da72f8905850a`; clean merge-tree at exact head
- 646 focused contract/runtime/DB/React/API tests pass on the combined tree; the review repair adds focused accessible-name and paginated multi-answer renderer regressions
- contracts/runtime/db/react/api/web typechecks pass
- production web build remeasured; bundle budget, compiled CSS, formatting, public hygiene, and `git diff --check` pass
- independent exact-head rereview: SHIP, no remaining blockers

## Visual approval

The desktop and narrow pending/resolved states were reviewed and approved before delivery. The final `Continue` → `Send answers` copy-only adjustment was explicitly approved without another visual gate.

Linear: https://linear.app/cloudgeni/issue/OPE-296/show-structured-human-input-answers-in-the-session-timeline

PR #1376, opened 2026-08-12
fix(react): unify durable composer attachments

jorgensandhaug:
## Summary

- render durable file refs through the composer attachment-card owner instead of a second generic restored-resource strip
- hydrate restored `FileAsset` metadata and signed image previews after refresh while preserving live blob previews during soft draft reconciliation
- keep exact default/custom mount identity through draft hydration, removal, and accepted-send cleanup
- add unit and Chromium coverage for upload → autosave → soft reload → hard refresh → preview failure fallback → removal
- register the browser regression in the fail-closed CI impact map and isolate public-boundary bundle proofs from suite-global loader state

## Testing

- [x] `bun run typecheck` (all 30 projects passed during the broad local check)
- [x] `bun run --cwd packages/react typecheck`
- [x] `bun run test:react:clean` (1,290 passed, 0 failed, no runtime warnings)
- [x] `bun test packages/react/test/use-file-attachments.test.tsx packages/react/test/chat-composer-uploads.test.tsx apps/web/src/lib/use-new-session-draft.test.tsx scripts/ci/impact.test.ts` (97 passed)
- [x] `bun test packages/react/test/artifact-public-boundary.test.ts packages/react/test/session-entry.test.ts` (6 passed; boundary pair also passed 10 consecutive runs)
- [x] `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH=... bun scripts/run-browser-e2e.ts ./test/e2e/composer-attachments.browser.e2e.ts`
- [x] React/web/example TypeScript checks, oxlint, oxfmt, `git diff --check`
- [x] `bun run --cwd packages/react demo:build`
- [x] Broad local `bun test`: 8,456 passed; remaining failures required unavailable Docker/PostgreSQL services. The candidate-owned CI impact and React boundary failures found by that run were repaired and their exact gates now pass.
- [x] Disposable computed merge against current-main ancestry passed targeted type/test/build/browser gates; current-base source admission remains the authoritative moving-main gate.

## Delivery integrity

- [x] Candidate/version labels represent substantive source changes, not base-only refreshes.
- [x] `main` advanced after candidate creation; the source head stayed focused and current-base mergeability/material compatibility were verified without rebasing unrelated commits.

## Notes

- Exact candidate head: `c218086db6f1123bdc0ec09046b3d11fb0edb4b7`
- No deployment is included in this change.

PR #1704, opened 2026-08-21
fix(web): stabilize message delivery interactions

benyhh:
## Outcome

- Keeps an optimistic user message continuously mounted while its canonical durable `user.message` event arrives, without showing an ordinary `Sending…` state.
- Keeps failed local deliveries visible with a red error affordance plus Retry and Remove, and surfaces their exact count through the existing production session-rail status slot—including loaded children under collapsed roots.
- Focuses the chat composer only after an explicit unmodified desktop session-row click, with guards for mobile/coarse pointers, dialogs, other focused controls, terminal/read-only sessions, approvals, and human input.
- Reorders new-session controls so tools/folder/repository configuration sits above the prompt while model selection stays beside voice/send; constrained layouts wrap without losing selected identities.

## Root cause

The optimistic row used its temporary event id as its React grouping key, while the durable SSE event used the canonical event id. Reconciliation therefore replaced the mounted group even though both records represented the same accepted `clientEventId`, creating a missing render frame. The renderer now uses that accepted client identity only as a presentation reconciliation key while preserving the canonical durable timeline id and idempotency semantics.

Local send failures remain separate from durable unread/session lifecycle state. The rail consumes the existing status geometry and only swaps in a small delivery-failure marker; the detailed error actions remain in the chat bubble.

## Verification

- `bun test packages/react/test/message-timeline-pagination.test.tsx packages/react/test/timeline-renderers.test.tsx apps/web/src/components/rail/session-list-delivery-attention.test.tsx apps/web/src/lib/session-attention.test.ts apps/web/src/lib/session-focus.test.ts apps/web/src/lib/sessions-group-channels.test.ts apps/web/src/pickers.test.tsx apps/web/src/repository-picker.test.tsx apps/web/src/session-control-surface.test.ts` — 170 pass
- `bun run --cwd packages/react typecheck`
- `bun run --cwd apps/web typecheck`
- `bun run --cwd packages/react check:css`
- `bun run --cwd apps/web build` — exact aggregate graph measurements remain below the documented next-whole-KiB envelopes
- `bun run lint`
- `git diff --check`
- clean merge-tree against current `origin/main`, PR #1696, and PR #1376

The desktop and constrained-width visual approval board was reviewed and explicitly approved before this PR. The production rail was then verified separately with the real `RailTrailingMetadata` component so no harness-only rail styling was shipped.

## Adjacent PR coordination

PR #1700 independently owns the workspace chrome redesign and its rail date-width change. It remains semantically compatible, but current main now creates expected mechanical conflicts in `scripts/check-web-bundle-budget.ts` and the sandbox-workspace surface; whichever merges second must preserve both main changes and remeasure the integrated graph.

Closes [OPE-298](https://linear.app/cloudgeni/issue/OPE-298/keep-sent-messages-continuously-visible-and-polish-composer)


benyhh (commented):
Independent review verdict: APPROVE at exact head `41668c8f5204af456bae06010ba9203ad17bf0f7`. (GitHub does not allow the authenticated PR author to submit a formal APPROVE review.)

Reviewed the full diff against current `origin/main` (`72736efa29377bfdd6f56983fbd810054c5d8da1`), including optimistic→durable identity/idempotency, failure actions and local rail attention, focus/accessibility guards, responsive setup hierarchy, repository identity, production rail fidelity, tests, and bundle-envelope rationale. The initial finding—lost message counts/loaded-child aggregation under `treeStats`—is fixed and covered by regressions.

The production rail keeps its existing metadata geometry; only the existing status slot gains the small failure icon, with Message not sent / Retry / Remove remaining in chat. Focused suite: 170 pass; React/web typechecks, CSS check, lint, and exact production web build pass. Merge-tree is clean with current main, #1696, and #1376. #1700 has the documented expected mechanical budget-file conflict and must remeasure the combined graph when merged second; its independent date-width geometry remains outside this PR. No remaining actionable findings.

PR #1710, opened 2026-08-21
fix(sessions): preserve private child authority

benyhh:
## Summary

- inherit omitted child visibility from the exact parent and reject cross-visibility child creation
- admit private children only through an exact live-attempt, same-owner PostgreSQL capability with canonical organization/workspace authority locking
- expose truthful bounded effective tool policy separately from persisted tool refs in `session_get`
- keep late child results durable without restarting settled no-goal, paused/completed-goal, or failed parents; active goals still wake
- freeze the exact private-session owner on internal-update attempts so Company Brain/profile context resolves under the correct authority

## Root causes

- `CreateSessionRequest` defaulted omitted visibility before child inheritance was resolved
- the MCP session detail bound silently dropped the already-computed effective tool policy
- every child terminal result autonomously woke an idle parent, even after its ordinary obligation had settled
- machine-input turns did not restore the private session owner as the causal initiating human

## Security and reliability

- DB trigger universally fences parent/child visibility at insertion under a parent row lock
- private-child capability is bound to parent/child IDs, live turn/attempt/generation, owner membership/subject, and allowed sandbox group
- organization advisory lock is acquired before workspace/parent rows; exact ordinary-workspace access remains locked through insert
- independent security/reliability review completed with no blocking findings

## Verification

- 34 focused real-Postgres tenancy, capability, Company Brain, and wake-policy tests
- 82 session control-plane tests
- 27 focused core/API/release tests
- API `session_create`/`session_get` integration slice
- worker child-create integration slice
- all 31 workspace typechecks
- full lint and format checks
- migration ordinal, FORCE-RLS backfill, docs-reference, and release-schema checks

Linear: [OPE-307](https://linear.app/cloudgeni/issue/OPE-307/preserve-private-child-authority-and-settled-parent-semantics-across)


benyhh:
CI follow-up `ac71646a6` addresses the exact failed leaves from the first run: the lost-wake fixture now exercises the preserved active-goal lane; the current-main sandbox file-link CSS artifact and hook dependencies are consistent; and the direct-session raw budget advances only to the next measured KiB. Exact local regressions, source guards, full typecheck, React CSS, web budget, and the previously failing E2E shard are green.

benyhh:
Rebased onto current `main` (`023b7f5be`). The landed #1709 now owns the upstream sandbox-file CSS/lint/budget repairs mentioned above; this branch retains only OPE-307 code plus the active-goal lost-wake fixture correction. Source guards are green after the rebase and CI run 32492082768 is queued.

PR #1711, opened 2026-08-21
fix(web): repair sandbox file-link build contracts

jorgensandhaug:
## Summary
- regenerate the committed React CSS for the width and ring utilities introduced by sandbox file-line links
- make the new file-open callbacks satisfy the exhaustive dependency contract
- advance only the measured direct-session raw bundle envelope to the next whole KiB

## Verification
- `bun run lint`
- `bunx oxfmt --check apps/web/src/routes/session.tsx scripts/check-web-bundle-budget.ts`
- `cd packages/react && bun run check:css && bun test --timeout 30000 test/compiled-css.test.ts`
- `cd apps/web && bun run build`

### DIFF
--- a/apps/web/src/routes/session.tsx
+++ b/apps/web/src/routes/session.tsx
@@ -152,6 +152,7 @@ export function SessionRoute({
   realtimeAutostartModel?: SessionRealtimeModel | undefined;
 }) {
   const context = useAppContext();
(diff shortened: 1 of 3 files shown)


jorgensandhaug:
Superseded by #1709, which landed the same hook and bundle-budget repair and removed the newly introduced CSS utilities so the committed compiled CSS remains current.

PR #1775, opened 2026-08-23
fix(deploy): stop single-node from stacking turns on one event loop

jorgensandhaug:
## Summary
- Single-node was `resource-based` / 256 on **one** turn worker. Temporal meters host CPU/RAM, not the Bun event loop, so one process admits tens of `runAgentTurn`s until `/readyz` misses kubelet (seen on Jorgebot: 44 inflight, 2–7s loop lag, 0/1 Ready).
- Profile is now **6 replicas × fixed 16** (96 host-wide, 16 per loop). Same density as the multi-worker default; Temporal still queues the rest.
- Docs and the Helm contract test follow.

## Test plan
- [x] `bun test scripts/helm-upgrade-contract.test.ts`
- [x] Jorgebot already running 6×16; `/readyz` 200, work spread across pollers

Made with [Cursor](https://cursor.com)

### DIFF
--- a/deploy/helm/opengeni/templates/prometheusrule.yaml
+++ b/deploy/helm/opengeni/templates/prometheusrule.yaml
@@ -112,8 +112,8 @@ spec:
             )
         {{- end }}
         # Node-relative protection cannot rely on a container memory limit: the
-        # single-node profile intentionally leaves the resource-based turn worker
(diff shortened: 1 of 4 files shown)


jorgensandhaug:
Closing. The 6×16 cap is a jorgebot overlay, not a change to the official single-node example.

PR #1788, opened 2026-08-24
chore(ci): refresh workflow execution manifest

benyhh:
## Summary
- regenerate the workflow execution graph manifest through the canonical generator
- record the current digests for three workflows and four uncapped-run steps
- unblock the source-contract dependency-install guard on current `main`

## Verification
- `bun install --frozen-lockfile --ignore-scripts`
- `bun run check:workflow-graph`
- `bun test scripts/workflow-execution-graph.test.ts`
- `bun run format:check`
- `bun scripts/workflow-execution-graph.ts --git-tree 'HEAD^{tree}'`\n- `git diff --check`

PR #1790, opened 2026-08-24
fix(ci): refresh workflow execution graph

benyhh:
## Summary
- refresh the committed execution graph after the workflow changes merged in #1661
- restore the source/type contract gate on current main and downstream PRs

## Verification
- bun run workflow-graph:refresh
- bun run check:workflow-graph
- bun test ./scripts/workflow-execution-graph.test.ts (29 pass)
- git diff --check


benyhh:
Closing as a duplicate of #1788. Independent review confirmed current main has the identical tree and manifest blob, so this PR has no remaining delta.

PR #1791, opened 2026-08-24
fix(web): restore the session creator monogram on root rail rows

benyhh:
## What

Restores the session creator monogram to the rail, on **root rows only**.

`apps/web/src/lib/creator-initials.ts` (`creatorInitials`, `creatorHue`) and its unit tests survived the chrome redesign, but the chip itself was dropped when the rail row body moved into `session-row-content.tsx` - the helpers had no caller left anywhere in the rail. This puts the chip back inside `RailTrailingMetadata`, before the status dot, at the exact prior visual: `size-4` round chip, `text-[8px] font-semibold leading-none text-white/90`, `background: oklch(0.45 0.11 <creatorHue(subjectId)>)`, `title` = the creator label.

## Root rows only

A spawned session inherits its parent's creator, so repeating the same monogram down every nested row is noise, and it costs title width on exactly the rows that have the least of it.

The predicate is `railRowCreator(session)` = `session.parentSessionId === null`, not the row's render `depth`. `parentSessionId` is the session's own lineage fact; `depth` is a placement artifact. Concretely, a collapsed channel section renders the *selected* session at `depth: 0` (`session-list.tsx`, `visibleRows`) even when that session is nested several levels down, so `depth === 0` would show a chip on a child row.

`SessionRow` is the single row renderer for every rail section (pins, channel sections, projects, browse forest, archive) and `SessionList` backs both the desktop rail and the mobile drawer, so one predicate at one call site covers every place root rows are listed.

### Flat projections say so

Search results and browse groupings are flat by design - a partial match set is not a tree, and a browse bucket is a list - so every row in them *is* top-level and must claim so, or a child renders chip-less beside chipped roots in a list where its parent is nowhere on screen.

`session-list.tsx` already normalized `parentSessionId` to `null` for `buildPinnedRailSections`, but passed the raw rows to `groupSessionsForBrowse`, so "Group by → Created/Creator" showed a ragged chip column. Both flat consumers now read one `projectRailSessions(browseSessions, hierarchyMode)` projection, which also removes the duplicated inline `map`.

## Layout

No fixed-width reservation and no placeholder for an absent chip. The title column stays `min-w-0 flex-1` and the metadata block stays `shrink-0`, so a row *without* a monogram keeps exactly today's title width and a row *with* one truncates correspondingly earlier.

One behavior change in `RailTrailingMetadata`: the chip is now its own reason to render the block. Previously it returned `null` when there was no schedule icon, status marker, or relative time; a mobile root row hits exactly that case and must still show who started the workstream.

The chip leads the block rather than sitting between the waiting duration and the dot, so a needs-attention root reads `BN 10h ● 3h` and not `10h BN ● 3h` - the waiting text and the dot are one status cluster (they share a `title`) and should not be split.

## Accessibility

The chip is `aria-hidden` with its `title` tooltip. The announcement lives in the row's **accessible name**: the row's `<Link>` carries an `aria-label`, and `aria-label` replaces name-from-content, so anything inside the link - the chip, and the pre-existing `sr-only` state line - is inert for what a screen reader actually says.

That name is now built by one exported `sessionRowAccessibleName` helper, which appends `. Created by <label>` gated on the same facts that decide whether the chip renders, so a row never announces a creator it does not show. The `sr-only` span is left exactly as it was on `main` rather than becoming a second, silently inert source of the same fact.

## Mobile

The monogram **is** shown on mobile. The trailing metadata block still renders there (schedule icon, status dot); mobile only moves the relative time down into the stacked sub-line, which leaves that block nearly empty. A 16px chip fits a 48px row comfortably, costs no vertical space, and "who owns this workstream" is if anything more useful on a narrow screen. Suppressing it would mean a mobile-only branch for no benefit.

## `priority.tsx`

The "For you" feed had its own copy of the same chip markup (`aria-hidden` + `title` + the same `oklch` expression, at `size-4.5`). Both call sites now use one shared `CreatorMonogram`; `priority.tsx` passes `className="size-4.5"` and tailwind-merge resolves that over the component's default `size-4`, so its rendering is unchanged. That leaves exactly one implementation of the chip in the tree.

The now-dead copy the redesign left behind in `session-list.tsx` is already gone from current `main` - nothing to remove there.

## Tests

New `apps/web/src/components/rail/session-row-creator-monogram.test.tsx` (20 tests, happy-dom):

- root row renders the chip with the right initials, hue, `aria-hidden`, and label in `title`; label falls back to the subject id
- creator-only metadata block renders when the row has nothing else
- no chip element for a child row, a service creator, or `unattributed-legacy`, and the block holds exactly two children in each case
- **frozen** exact `className` on the metadata container. This replaces an earlier tautological `creator={undefined}` vs `creator={null}` comparison, which was the same code path and could not fail. Verified enforceable by adding `pl-1` to the container: the test fails with the exact diff, then passes again once reverted.
- chip ordering against a needs-attention row: `["BN", "10h", "", "3h"]`
- `sessionRowAccessibleName` announces the creator for a root row, omits it for a child row and for service / legacy creators, and keeps pinned + spawned facts ahead of it
- `sr-only` state line unchanged for root, child, and service rows
- title column stays `min-w-0 flex-1` and metadata stays `shrink-0` with and without a chip
- mobile row keeps the chip

Plus `railRowCreator`, `creatorLabel`, and `creatorAnnouncement` units in `sessions-group-channels.test.ts`, and a flat-projection block in `session-browse.test.ts` covering `projectRailSessions` (lineage preserved in hierarchy mode, flattened otherwise without mutating the source rows) and both flat consumers agreeing that every row is top-level.

Results: `bun run typecheck` clean across all projects; `apps/web` 947 pass / 0 fail; `bun run check:public-hygiene` and `bun run check:docs-refs` pass; oxfmt + `oxlint --deny-warnings` clean on `apps/web/src`.

## Bundle

`bun run --cwd apps/web build` passes with the budget untouched.

`main` has advanced past this branch's base (#1785 raised the direct-session envelopes to 2,163,712 raw / 605,184 gzip / 24 files). Compatibility verified **without source mutation**, by building a throwaway merge of this frozen head into current `main` on macOS/arm64:

| graph | merged with current main | envelope |
| --- | --- | --- |
| `directSessionRaw` | 2,163,091 | 2,163,712 |
| `directSessionGzip` | 603,176 | 605,184 |
| `directSessionFiles` | 24 | 24 |

Inside every envelope with 621 raw bytes of headroom, so no budget line moves and `scripts/check-web-bundle-budget.ts` is not touched. That merged tree also passes `bun run typecheck` and the full `apps/web` suite (952 pass / 0 fail). No files in this change overlap what `main` touched.

`apps/web` is in the changeset `ignore` list, so no changeset.


benyhh:
Reopening to re-run CI against the repaired `main` (`29829c4f`). No source change.

PR #1792, opened 2026-08-24
feat(sessions): acknowledge a child when its parent turn consumes its lifecycle notice

benyhh:
## What

A parent turn that consumes a child session's lifecycle notice as machine input now advances that child's read-acknowledgment fence for the human who initiated the parent's turn, exactly as if that human had opened and viewed the child.

An orchestrator with dozens of children currently shows a blue "unread" dot on every child forever until a human opens each one by hand, even though the parent agent already consumed and acted on each result.

## Where

`claimSessionWorkForAttempt` (`packages/db/src/index.ts`) is the only place `session_system_updates` rows become `delivered` and get bound to a receiving turn plus one `session_history_items` row. Both of its delivery sites now call `acknowledgeConsumedChildLifecycleNotices` immediately after `persistDeliveredUpdateBatch`, in the same transaction:

- the internal-update path, right after the `system`/`goal` turn's batch is persisted (subject: the `initiatingHumanSubjectId` that path already resolves for the frozen turn principal);
- the queued-turn path, right after a queued human/API turn's coalesced batch is persisted (subject: `row.initiatingHumanSubjectId ?? (row.initiatorKind === "subject" ? row.initiatorSubjectId : null)`, the same expression `registerAttempt` uses for personal preference authority).

All six child lifecycle kinds carry the same `childSessionId` field, and the set is derived from `CHILD_LIFECYCLE_SYSTEM_UPDATE_KINDS` via an `Extract<>` type so a future seventh kind without that field fails to compile instead of silently dropping out. The ids come from the payloads the delivery loop already validated, so no row is parsed twice.

## Guarantees

- **Generic.** Keyed only on "a claimed turn consumed a child lifecycle update, and that turn has a frozen initiating human". No orchestrator, goal, or depth special case; a nested chain behaves identically at every level.
- **Per viewer.** Only the initiating human's fence moves. Another member still sees the child unread.
- **Removes noise only.** `unread` is nothing but `sessions.last_sequence > acknowledged_sequence`, so a child that emits one more event goes unread again with no special handling, and the `failed` (red) and `requires_action` (purple) rail indicators are derived from `sessions.status` and rank above unread in `summarizeRailNodes`. No client change is needed.
- **No OCC revision minted.** The acknowledgment never touches `attention_version`. That revision exists only to order explicit human attention mutations against each other, and this writer publishes no event, NATS invalidation, or sequence advance a browser could learn from, so bumping it would silently stale the version the rail holds and 409 the human's next mark-read click.
- **Monotone.** A single statement whose conflict path refuses to move a fence backward, so a human who already read further, or a racing claim that observed a later sequence, is never regressed.
- **Batched.** Several notices for one child in one delivered batch produce exactly one upsert reflecting the final sequence.
- **Service turns write nothing.**
- **Idempotent with recovery.** The acknowledgment commits or rolls back with the claim that made the batch durable. Recovery reuses the already-persisted batch and re-delivers nothing.

## Behaviour change worth calling out: mark-unread is not sticky

Because the fence is monotone and OpenGeni stores no durable "leave this unread" intent, an explicit mark-unread does **not** survive a later consumption: mark a child unread, let it work and report again, and the parent's next claim acknowledges it once more. That follows from the PR's premise that consumption is the read signal, and the mark still holds until the parent consumes a newer notice, but it means mark-unread is not a durable personal to-do flag on a child of an active orchestrator.

Making it sticky would need a durable explicit-unread marker the acknowledgment respects. This PR deliberately does not add one, and instead states the real behaviour in the changeset, `docs/durable-agent-inputs.md`, the docstring, and a test that pins it.

## Lock order and fencing

The claim already owns the canonical workspace-control -> workspace -> session -> turn -> attempt prefix for the **parent**.

- It takes the `session-personal-state` fence with `pg_try_advisory_xact_lock` and **skips** the acknowledgment when the fence is already held. `remove_workspace_member_command` (migration 0278) takes that fence *before* the workspace/session prefix, so blocking on it would invert the canonical order; a non-blocking probe cannot. Honouring the fence is what keeps this from being the "racing pin writer recreates personal rows after cleanup" hole 0278 exists to close, and losing the probe is a clean no-op: the child stays unread, exactly as today.
- It takes no child turn/attempt lock and no explicit child session lock. `session_pins` does carry two FKs to `sessions`, so each inserted row takes `FOR KEY SHARE` on the child's session row. That is compatible rather than an inversion: `FOR KEY SHARE` conflicts only with `FOR UPDATE`, and every canonical session writer takes `FOR NO KEY UPDATE`.
- One `INSERT ... SELECT ... ORDER BY id` covers the whole batch, so rows lock in the same UUID order the rest of this file uses and the index order 0278 deletes in. This replaced up to 100 sequential single-row inserts inside the workspace-wide critical section.
- The select is fenced on `parent_session_id`, so a payload field can never decide whose personal state is mutated on an unrelated session.
- Both `session_pins` RLS policies still apply, through the existing `withTemporarySubjectRls` scope.

There is no membership probe: the write grants nothing and discloses nothing, the subject comes from durable frozen turn provenance rather than a request, membership removal deletes these rows and is now serialized against this writer by the fence, and the personal-workspace-owner exception a bare membership probe would need is authorized by an API-layer canonical-managed-cookie stamp the worker path does not have.

Because a `session_pins` row can appear between another writer's read and its write, `setSessionPin`, `setSessionAttention`, and `setSessionArchive` upsert instead of insert. The acknowledgment row is pin-neutral and archive-neutral (exactly the absent-row projection), so their conflict path is the same transition their insert performs. `setSessionAttention` additionally writes `greatest(acknowledged_sequence, ...)` on its read branch so a stale read cannot regress a fence that committed after it; an explicit mark-unread keeps its exact lower sentinel.

`withTemporarySubjectRls` no longer masks the original SQLSTATE. Restoring the subject scope inside an already-aborted transaction raises 25P02, which replaced the real error; callers on the claim path classify the original 40P01/40001 for the idempotent-persistence retry, so that substitution silently disabled the retry. A restore failure is now suppressed only when `fn` already threw.

## Known scope limit

The mechanism fires only when the claiming turn has a frozen initiating human: queued human/API turns that coalesce child notices, goal continuations (which resolve the human through the continuation's `causalTurnId` lineage), and scheduled / xAI-user / private-owner internal turns. A bare `child_terminal_result` system turn on a `workspace_shared` parent with no goal continuation in the same batch freezes the `internal-update` service actor and acknowledges nothing. That is intentional and covered by a test; widening the subject resolution is deliberately out of scope.

## Tests

`packages/db/test/child-read-acknowledgment.test.ts` (10 tests): fence advance and `unread: false`; a different subject still `unread: true`; a further child event flips it back; a goal-continuation batch acknowledges for the causal human; a pure-service turn writes nothing; a failed child still reports `status: "failed"`; an already-ahead fence is not regressed; two notices for one child produce exactly one row; a nested three-level chain at every level; no OCC revision is minted so a subsequent `expectedVersion` attention mutation still applies; a later consumption advances past an earlier mark-unread; a notice naming a non-child session acknowledges nothing.

`packages/db/test/session-pins.test.ts` gains three tests that reach the new conflict branches **deterministically**, using this file's existing `BEFORE INSERT` advisory-barrier idiom to park each mutation after its absent-row read while an acknowledgment-shaped row is inserted without the personal-state fence. All three were verified by falsification: sabotaging each conflict `set` makes exactly those three tests fail.

`session-event-writer-audit.test.ts` does not apply (this writer inserts no `session_events`); it passes unchanged.

Verified locally: `bun run typecheck`; `oxfmt --check`; `oxlint --deny-warnings`; `bun run check:public-hygiene`; `bun run check:docs-refs`; `bun changeset status`; and `OPENGENI_REQUIRE_REAL_DB=1 bun test` on `child-read-acknowledgment`, `session-pins`, `child-lifecycle-notices`, `session-control-plane`, `queue-mutations` (159 together), `durable-goal-wake` (44 solo), and `apps/api/test/session-authorization` plus `session-event-writer-audit` (81).

## Docs

`docs/durable-agent-inputs.md` gains a "Consuming a child notice acknowledges that child" section; one clause each in `AGENTS.md`'s machine-input bullet and the `docs/architecture.md` session-pins sentence. Changeset: `@opengeni/db` minor.


benyhh:
Reopening to re-run CI against the repaired `main` (`29829c4f`). No source change.

PR #1794, opened 2026-08-24
feat(sessions): complete atomic visibility forks

benyhh:
## Summary

- add one atomic same-workspace fork contract with explicit private or workspace visibility
- require a durable warning acknowledgement when private content crosses into workspace scope
- derive fresh destination owner, authority, provenance, root, and sandbox group without copying live credentials, grants, connections, tools, turns, goals, processes, or pins
- allow any currently authorized workspace member to fork a shared source while keeping private sources owner-only
- expose the generic managed-web Fork dialog with exact retry and destination reconciliation

## Verification

- real PostgreSQL migration suite, including shared-member to private fork and later source access loss
- dedicated-schema and TEMP-shadow isolation coverage
- API/core/web/contract focused suites
- release-schema and migration-ordinal guards
- all 31 TypeScript projects
- lint, format, and diff checks

Linear: OPE-197


benyhh:
Repair checkpoint at exact head `bfd1c01c2babc125d331ea51c732c8904b1ac8bc`:

- exact applied receipts now resolve before mutable source visibility/authority checks, so the same key/body replays after a shared source becomes private; a changed body conflicts and a new key is denied;
- personal GitHub authority markers are removed from forked resources, with real-Postgres fork coverage exercising the canonical validator used by follow-up turns;
- only the stale current-main managed-self-context assertions were updated.

Local evidence: real-Postgres fork/UI 19/19, personal-workspace API 24/24, dedicated-schema replay 16/16, focused tenancy/web 55/55, release/schema activation surface 17/17, all 31 TypeScript projects, lint, format, and diff checks. Fresh hosted CI and external review are pending.

benyhh:
OPE197_CURRENT_MAIN_COMPATIBILITY_V2 — exact head bfd1c01c2babc125d331ea51c732c8904b1ac8bc integrates conflict-free with protected main 92324b57d185f6581e3d1028bf55e1f76bfc9c01 (integrated tree a4e7db75df934e3aaf14cb1dce5315d40ea2cfb1). The integrated checkout passes 114 focused real-PostgreSQL/API/core/contracts/web tests and all 31 TypeScript projects. Hosted CI is fully green: 34 pass, 9 intentional skips, 0 bad. Live OPE-197/OPE-10 comments were reread first; no newer acceptance addition was found. Fresh independent exact-head review and external APPROVED review remain mandatory.

benyhh:
Exact-head CI is now complete for `bfd1c01c2babc125d331ea51c732c8904b1ac8bc`: [run 32729367054](https://github.com/Cloudgeni-ai/opengeni/actions/runs/32729367054) has 34 passes, 9 intentional skips, and 0 bad conclusions. The orchestrator also proved conflict-free integration against protected main `92324b57d185f6581e3d1028bf55e1f76bfc9c01` at merge tree `a4e7db75df934e3aaf14cb1dce5315d40ea2cfb1`, with 114 focused tests and all 31 typechecks green. The PR remains open and mergeable; fresh independent exact-head review is still required and `davletd` remains requested.

benyhh:
INDEPENDENT_EXACT_HEAD_BLOCK — bfd1c01c2babc125d331ea51c732c8904b1ac8bc still authorizes the mutable source before reaching the repaired applied-receipt lookup. After a shared member commits a fork and the source later becomes private, an exact lost-response retry is rejected in the application layer; only the DB-helper test passes. Repair is active: exact actor/workspace/source/key/hash-bound applied replay must occur before mutable source authorization, while changed-body reuse conflicts and fresh keys deny, with real PostgreSQL core/API coverage. This head must not merge.

benyhh:
### Head updated to `352975298` after an independent exact-head review returned BLOCK

The review of `736beaad` found **two blockers inside this change's own surface**, one important doc defect, and five minor findings. All are repaired here.

**1 — a private fork bypassed the migration-0323 organization private-session decision.**
`open_private_session_create_capability` refuses a private *create* in a shared workspace when the owner/admin setting is off. The fork path minted its own visibility-write capability and never consulted it, and 0311's private-create branch requires `forked_from_session_id IS NULL`, so a fork can never consume a `private_session_create_capabilities` row — it is admitted through the alternative lifecycle-capability branch, which carries no product gate.

Reachable: an owner disables private sessions, the create surface correctly hides "Only me", and any authorized member then opens the new Fork dialog, picks Private, and lands a `user_private` session in that shared workspace. The bypass predates this PR at the API level, but this PR exposes it in product UI for the first time and widens it from source-owner-only to any authorized member on any shared session.

Repaired in the migration, in the shape the create path documents: the check sits **after** keyed-replay resolution and **before** any source read, so a fork that already committed still replays byte-identically once the setting is disabled, while a fresh key fails closed with the same SQLSTATE `55000`. Personal workspaces stay exempt. `forkSessionContent` maps that state to `SessionTenancyNotActivatedError`, and the dialog no longer offers a Private destination the database would refuse — including while the capability answer is unknown, which fails closed.

**2 — `replay_applied_session_fork` / `replayAppliedSessionFork` were not pinned** by `test/session-visibility-contract-surface.test.ts`. That routine exists precisely to answer without consulting mutable source authorization and is granted to `opengeni_app`, so a future worker/MCP/runtime/React import would have passed every guard in the tree while the docs updated by this PR assert the boundary is pinned. Both names are now in the pinned sets — and the guard immediately caught a doc comment that named the SQL function inside a web component.

**3 — `AGENTS.md` said the opposite of what ships** ("every production session stays `workspace_shared`"). Rewritten.

Minor findings applied: receipt lookups bind `account_id`; the forked row records `create_requested_visibility`; the dead `prepareSessionForkAttempt` is deleted (no caller, and it contradicted the live controller); the confirmation is destructive only for private → workspace.

Deliberately **not** applied and recorded as follow-ups: reading `authority_epoch` back from the receipt instead of the hardcoded `1` (regression risk on older receipts, no behaviour gain), and 0303's deployment-global `SHARE ROW EXCLUSIVE` locks on 17 tables including `sessions`, inherited byte-identically here — real, but it predates this candidate and belongs in its own change.

**New coverage.** Real-PostgreSQL proof that a private fork is refused while disabled, that a shared fork of the same source still succeeds, and that a fork committed while enabled still replays after disable; plus a web test that the dialog offers only Workspace and defaults a private source's fork to workspace when the capability is false.

### Migration renumbered three times

Five open PRs all claimed 0333. #1798 took 0333, #1799 took 0334, #1796 took 0335. This candidate moved 0333 → 0334 → 0335 → **0336** with `bun run migration:renumber` each time; no migration bytes changed.

### Local validation on this head

`typecheck` all 31 projects clean · `lint`, `format:check`, `check:docs-refs`, `check:migration-ordinals`, `check:migration-rls-backfills` all pass · **215 pass / 0 fail** across the focused release/core/contracts/SDK/db/API/guard suites · `apps/web` **939 pass / 0 fail**.

Needs terminal green CI on this head, a fresh independent exact-head review (the prior evidence was for `736beaad`), and an external `APPROVED`.

benyhh:
### Fresh independent exact-head review of `352975298`: PASS (with non-blocking notes)

Both previously-blocking findings are confirmed closed, and the reviewer settled the riskiest question empirically rather than by reading.

**Blocker 1 — the 0323 product gate.** Verified at `0336_atomic_session_fork_visibility.sql:280-287`: after the keyed-replay early return (249-270), before the source read (289). The personal-workspace exemption is semantically identical to the create path's `IF NOT actor_personal_workspace` and strictly *stricter* on a NULL `personal_workspace_id` (fork fails closed where the create path silently skips). Web-side, `canForkPrivately` uses the same predicate the SQL does, initialises `false`, and `.catch(() => false)`. The reviewer tried to get around it — replay needs an applied receipt bound to the calling actor, the hash binds visibility + acknowledgement so key reuse conflicts, and the legacy 9-arg overload forwards into the same gated body — and found no path.

**Blocker 2 — the replay surface** is pinned in both entry-point sets; guard suite green at 10,504 assertions.

### The FORCE-RLS production-posture trap is proven ABSENT here

This mattered: two sibling tenancy candidates were blocked the same day because a SECURITY DEFINER routine read a FORCE-RLS table without the capability its policies require and silently saw zero rows in production. This gate reads FORCE-RLS `organization_private_session_settings` from inside `fork_session_content`, so it was checked on a real production-posture database (`acquireOwnerMigratedTestDatabase`, NOSUPERUSER/NOBYPASSRLS owner, `relforcerowsecurity = true`):

| probe | result |
|---|---|
| `organization_private_sessions_enabled()`, enabled | **true** (not zero-rows) |
| same, disabled | false |
| `forkSessionContent` private, enabled | succeeds, `user_private` |
| `forkSessionContent` private, **disabled** | **refused, SQLSTATE 55000** |
| committed fork replayed after disable | same `sessionId`, `replay: true` |
| shared fork, disabled | succeeds |

Safe because the `..._scoped_read` policy keys on `opengeni.account_id` (already proven equal to `p_account_id` earlier in the routine) and the read is a plain `EXISTS(SELECT 1 …)` with **no `FOR SHARE`/`FOR UPDATE`** — so the UPDATE/ALL-policy asymmetry that broke the siblings does not apply.

The new gate test was checked for vacuity and is self-proving: one test flips only `enabled` and gets refuse → allow → replay-after-disable, and a shared fork on the same source succeeds in the same test, closing the "org isn't activated" escape.

### Non-blocking findings — the head stays frozen deliberately

Per the delivery contract, a candidate is mutated only for a source defect, a real merge conflict, or a material semantic incompatibility. None of these qualify, CI is green here, and `main` is consuming migration ordinals faster than candidates clear their gates (this migration has already been renumbered three times), so re-opening the head would trade a green reviewed candidate for another renumber race.

Filed instead:
- **OPE-352** — `transition_session_visibility` carries no 0323 gate, so fork-then-transition reaches the same end state. Reproduced under owner posture. Pre-existing, but amplified: before this PR a member could only privatise sessions they already owned; now any authorised member can fork another member's shared session and privatise the copy.
- **OPE-353** — the inherited `LOCK TABLE … SHARE ROW EXCLUSIVE` over 17 hot tables is unqualified, stalling writes for every tenant in the schema for the length of a full history copy. Not newly reachable (the shipped visibility toggle takes the same lock) but newly triggerable from a UI button, with a duration that scales with session size.

Minor, recorded but not filed: all six SQLSTATE 55000 sites map to "not activated" (five are internal invariants); `docs/organization-tenancy.md` and `docs/architecture.md` rewrote the fork contract without mentioning the new gate, though AGENTS.md documents it; and `usePrivateForkCapability` is never exercised directly since every web test injects the prop.

### CI

32 pass / 0 fail on this head, 8 intentional skips. `mergeable: MERGEABLE`. **External `APPROVED` is the only remaining gate.**

PR #1796, opened 2026-08-24
feat(slack): add workspace routing tables and resolvers

benyhh:
Steps 1 and 2 of per-channel and per-DM Slack workspace routing: the schema and a pure `packages/db` resolver library.

## Nothing reads any of this yet

No route table is consulted, no new column is written, and no existing code path changed. Single-workspace Slack behaviour is byte-identical. The routing decision, the picker, the per-message workspace line, and the web settings are all later steps.

## Migration `0333_slack_workspace_routing.sql` (rolling)

Four FORCE-RLS tables, each with a `workspace_isolation` policy carrying both `USING` and `WITH CHECK`, and full DML granted to `opengeni_app`:

- `slack_channel_routes` and `slack_user_dm_routes` are the durable ask-once memory, unique per `(connection_id, channel|user)`.
- `slack_route_prompts` and `slack_route_prompt_options` hold the pending first-use picker. `slack_interaction_action_handles` cannot carry it: its `session_id` is `NOT NULL` and composite-FK'd to an existing `slack_interactions` row, while a picker exists before any session.

Plus five additive nullable routing columns and a `route_state` CHECK on `slack_interaction_inbox`, a frozen `routed_workspace_label` on `slack_interactions`, and exactly one new `SECURITY DEFINER` routine: `opengeni_private.resolve_slack_interaction_tenancy(uuid, text)`, an ids-only tenancy probe pinned to `search_path = pg_catalog`, revoked from `PUBLIC`.

**The installation binding is deliberately untouched.** `resolve_slack_installation`, `slack_installation_bindings_active_team_uq`, and `sync_slack_installation_binding()` are unchanged, so one Slack team still installs into exactly one home workspace and one credential. Routing is a separate additive fact.

**There is no backfill.** `route_state IS NULL` means "legacy / never routed" and is exactly today's behaviour, which is what makes this rolling-safe: an old API or worker image ignores the new tables and columns entirely. `bun run check:migration-rls-backfills` stays green without touching its grandfathered lists.

## Home and target tenancy

The Slack bot credential is workspace-owned and every provider call is fenced on it, so a routed workspace can never build the bot client. Routing therefore splits one identifier in two:

| scope | owns |
|---|---|
| **home** = the installation binding's `(account_id, workspace_id)` | the connection and bot credential, the post/update/delete ledgers, `slack_interaction_inbox`, `slack_bot_user_links`, App Home, reaction-summon settings, Slack task policy, and every table added here |
| **target** = the routed `(account_id, workspace_id)` | the `slack_interactions` row, its action handles, its progress deliveries, the grant, the session, and every session event |

Both sides always sit in one organization: `target_account_id = account_id` is a table CHECK. Cross-organization routing is explicitly out of scope, and adding that CHECK later would not be a rolling migration.

The inbox keeps `account_id`/`workspace_id` `NOT NULL` as home tenancy on purpose. `workspace_rls_visible` is strict equality, so a nullable `workspace_id` would make an un-routed row invisible to every tenant including its own, and `settleSlackInteractionInbox` would update zero rows while reporting success.

## Resolvers

- `probeSlackInteractionTenancy` returns ids only, so thread continuation can cross workspaces on the connection-global `(connection_id, route_key)` unique without widening any read.
- `getSlackChannelRoute` / `getSlackUserDmRoute` and their `upsert*` / `delete*` siblings, all RLS-scoped under home tenancy. A repeat upsert re-points the one row and bumps `version`; it never forks a second answer.
- `listSlackRoutableWorkspacesForSubject` is the single implementation of "workspaces this subject may start work in", so a future shared candidate-set contract can replace it wholesale. It unions permission-filtered memberships with the subject's own personal workspace. `listWorkspacesForSubject` is not usable here: it has no permission filter and cannot see personal workspaces at all, because a managed human's personal workspace deliberately has no `workspace_memberships` row.
- `resolveSlackTargetAuthority` is the only place a Slack request may reach a managed human's personal workspace. `getWorkspaceGrant` is **not** widened, so no API-key or delegated-bearer principal gains personal-workspace access. The pointer path accepts only the subject's own `personalWorkspaceId` from an active same-organization membership, then re-checks live authority through `namedSubjectHasLiveWorkspaceAuthority`. It lives in `packages/db/src/index.ts` rather than the new module only because no sibling module in that package imports `./index` and this one should not be the first.

## Tests

`packages/db/test/migration-0333-slack-workspace-routing.test.ts` (8 tests) and `packages/db/test/slack-routing.test.ts` (11 tests) both run against real PostgreSQL. They prove tenant isolation under `opengeni_app` GUCs including a refused `WITH CHECK`, the same-organization fence, the composite target FK cascade, all four arms of the inbox `route_state` CHECK plus its legacy arm, the definer routine's `SECURITY DEFINER` and `proconfig` pin and its `PUBLIC` revoke, that the candidate set includes a personal workspace and excludes read-only memberships and other organizations, that the personal grant is minted only for its own owner and stops once the membership is no longer active, and that the probe crosses workspaces and returns ids only.

## Incidental repair

Migrations `0331` and `0332` landed on `main` without being registered in `scripts/release-schema-contract.test.ts`, which left that test failing on `main`. This PR registers them alongside `0333`, so the contract test is green again.


benyhh:
Pushed a repair for a blocker found by an independent adversarial review panel and then reproduced directly against Postgres.

**The blocker.** `opengeni_private.resolve_slack_interaction_tenancy` was a bare `SECURITY DEFINER` routine over `slack_interactions`, which carries `FORCE ROW LEVEL SECURITY`. `FORCE` binds the *table owner*, and OpenGeni migrates as a `NOSUPERUSER NOBYPASSRLS` owner, so the routine executed as a role the strict `workspace_isolation` policy still applies to. It returned **zero rows** for exactly the cross-workspace lookup it exists for, which is the lookup the whole thread-continuation design rests on.

The existing tests could not see this: `acquireSharedTestDatabase` hands out the container superuser, for whom `FORCE RLS` never engages. Reproduced on a fresh database owned by a `NOSUPERUSER NOBYPASSRLS` role with the real migration chain: a caller scoped to workspace A probing an interaction owned by workspace B got `[]`.

**The fix** adopts the private transaction-capability seam migration 0243 already established for Google Drive object ACLs: a transaction-local row in a private `opengeni_private` table opens one owner-checked permissive `SELECT` policy for the duration of the probe, deleted on both the success and the failure path. `opengeni_app` holds no privilege on that table and cannot forge the path with a custom GUC. The pre-existing RESTRICTIVE `session_visibility_isolation` policy still applies, so this never widens session visibility.

**New coverage** in `packages/db/test/migration-0333-slack-tenancy-probe-owner.test.ts` drives the real migration chain through `acquireOwnerMigratedTestDatabase` and proves:

- the probe resolves a workspace-B interaction while the caller is scoped to workspace A;
- the caller still cannot read that row directly, before or after the probe;
- an unknown route returns nothing and leaves no capability row behind;
- the runtime role is denied every direct privilege on the capability table (`42501`).

**Also from the panel:**

- The inbox routing columns had no same-organization CHECK, unlike all four new tables. Added, and proven to raise. This is the row the pump actually claims and routes from, and the one place `target_workspace_id` deliberately carries no foreign key.
- `personalWorkspaceIdForSubject` leaked `opengeni.subject_id` into a caller-owned transaction. A nested drizzle transaction is a SAVEPOINT and releasing it does not undo a transaction-local `set_config`, so the probed subject silently re-scoped the caller's remaining statements. Now restored in a `finally`, matching `namedSubjectHasLiveWorkspaceAuthority`. The regression test fails without the fix.
- The same-organization CHECK is now exercised on `slack_user_dm_routes` and `slack_route_prompt_options`, not just `slack_channel_routes`.
- The candidate-ordering assertion was vacuous on a shrunken or empty result; it now pins the set as well as the order.

25 further claimed findings were rejected under adversarial verification.

PR #1797, opened 2026-08-24
fix(release): register migrations 0331 and 0332 in the schema contract

benyhh:
`main` was red on two independent causes. PR #1800 landed the web/e2e test repairs directly on `main`, so this candidate is now reduced to the one failure #1800 deliberately left out: the release-schema contract.

## What remains here

`scripts/release-schema-contract.test.ts` fails on `main` because commit `0d75b4c69` (#1793) added `0331_managed_organization_creation.sql` and `0332_organization_shared_workspace_control_plane.sql` without adding them to the forward-migration list. The checkpoint input therefore still framed both files, producing `d4944e04...` where the pinned governed `0326` checkpoint is `c4ec5d41...`.

This registers exactly those two migrations and pins their bytes:

- `0331` sha256 `62912af210c1db18c7b58de8ad2c233f2fc6759a6ef4509c8e6352d75dd0641f`, `-- deployment-mode: rolling`
- `0332` sha256 `6de1a4fc6f0dc0e67852cef996b6d0e01326d180c67d188cb7fcb9c704bc5cdd`, `-- deployment-mode: rolling`

Both hashes were verified byte-exact against the files on disk, and both deployment modes read directly from the first line of each migration. No migration bytes, generator, or generated artifact changes.

## Merge resolution

`main` advanced with #1800, which repaired the same five test files this branch had repaired. Every conflicting file is resolved to `main`'s version, so the merged tree differs from `main` by this one file only. No product code changes and no feature is reverted.

## Verification

- `bun test scripts/release-schema-contract.test.ts`: 6 pass, 0 fail
- `bun run check:migration-ordinals`: ok, 341 migrations, next free ordinal `0333`
- `bunx oxfmt --check scripts/release-schema-contract.test.ts`: clean
- `bun run check:public-hygiene`: clean

Before the merge resolution, this branch's `103b8575` (Slack access-link e2e alignment) was independently reviewed and mutation-tested: reverting the managed-onboarding branch in `context.tsx`, dropping the `INVALID_EMAIL_OR_PASSWORD` mapping in `managed-auth-form.ts`, and disabling URL-fragment scrubbing each made the corresponding assertion fail. That content is now superseded by `main`'s equivalent from #1800.

Duplicate of #1802, which carries a byte-identical change to the same file.

Test-only change, so no changeset.

### DIFF
--- a/apps/web/src/managed-self-context-surface.test.ts
+++ b/apps/web/src/managed-self-context-surface.test.ts
@@ -53,16 +53,11 @@ describe("managed self-context surfaces", () => {
     expect(settingsSource).toContain("<MembersSection workspaceId={workspaceId}");
   });
 
-  // `get_organization_administration_overview` excludes every membership's
-  // personal workspace from the projection; the workspace-access section has to
-  // keep saying so, otherwise an administrator reads the shared-only roster as
-  // the whole organization. Keep these substrings on one source line each so an
-  // ordinary reflow does not masquerade as a removed guarantee.
   test("separates organization administration from Personal content", () => {
-    expect(organizationAdminSource).toContain("Personal workspaces stay private.");
     expect(organizationAdminSource).toContain(
       "Create shared workspaces, then choose which organization members can use each one.",
     );
+    expect(organizationAdminSource).toContain("Personal workspaces stay private.");
     expect(organizationSource).toContain("<OrganizationPeopleSection");
     expect(organizationSource).toContain("<OrganizationRetentionSection");
   });
--- a/apps/web/src/managed-self-context-surface.test.ts
+++ b/apps/web/src/managed-self-context-surface.test.ts
@@ -53,11 +53,16 @@ describe("managed self-context surfaces", () => {
     expect(settingsSource).toContain("<MembersSection workspaceId={workspaceId}");
   });
 
+  // `get_organization_administration_overview` excludes every membership's
+  // personal workspace from the projection; the workspace-access section has to
+  // keep saying so, otherwise an administrator reads the shared-only roster as
+  // the whole organization. Keep these substrings on one source line each so an
+  // ordinary reflow does not masquerade as a removed guarantee.
   test("separates organization administration from Personal content", () => {
+    expect(organizationAdminSource).toContain("Personal workspaces stay private.");
     expect(organizationAdminSource).toContain(
-      "Personal workspaces and their content are\n            never included.",
+      "Create shared workspaces, then choose which organization members can use each one.",
     );
-    expect(organizationAdminSource).toContain("Every shared workspace in this organization.");
     expect(organizationSource).toContain("<OrganizationPeopleSection");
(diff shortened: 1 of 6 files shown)


jorgensandhaug:
Added `9e8e93c9b` to complete the deterministic CI repair:

- align the browser acceptance with the intended omission of a redundant visibility chooser when Workspace is the only valid option; no timeout increase
- pin the exact SHA-256 and rolling deployment mode for migrations 0331 and 0332

Local verification: the browser suite passed 10 consecutive repetitions, then passed again from this PR worktree; the schema and component contract tests are green.

jorgensandhaug (approved):
Reviewed the exact `7ce012f92` head. The repair preserves product behavior, restores the stale unit/browser contracts, keeps published schema history intact, and adds exact rolling migration receipts. Targeted schema, component, browser, public-hygiene, impact, and type checks pass locally; the lifecycle-specific regression suite is also green.

jorgensandhaug:
Added `103b85752` after reproducing the remaining E2E shard-1 failures locally. The patch aligns the Slack access-link tests with the managed auth/onboarding UX, uses the real Better Auth error code, mocks the new invitation read, and waits for the visible continuation UI before asserting the URL fragment has been scrubbed. Both changed cases pass together locally; formatting is clean.

jorgensandhaug:
Superseded after #1800 merged the overlapping web-test repairs into main. The two remaining deltas were rebuilt cleanly on current main in #1803: release-schema registration and the reproduced Slack URL-scrub/auth-fixture fix.

benyhh:
Post-close record: `103b8575` was independently reviewed and mutation-tested.

Verdict: legitimate, test-side only, no product regression masked.

The "No workspace access" failure was not stale copy and not a regression. #1793 (`0d75b4c6`) added the `ManagedWorkspaceOnboardingPanel` branch in `apps/web/src/context.tsx`, so the bare `ProblemPanel` is now the unmanaged path only; this e2e runs `auth.mode: managedSession`. Likewise `"Sign in failed"` was a toast in `context.tsx` that #1793 replaced with the inline `Notice` titled "Couldn't sign in".

Three mutations, each making exactly the guarded assertion fail and nothing else:

- force the old `ProblemPanel` branch in `context.tsx` -> fails on "Set up your OpenGeni workspace"
- remove the `INVALID_EMAIL_OR_PASSWORD` mapping in `apps/web/src/lib/managed-auth-form.ts` -> fails on "Email or password is incorrect."
- disable URL-fragment scrubbing in `stripSlackLinkFromBrowserLocation` -> fails on the reordered hash assertion, so moving `expectVisible` earlier did not defang it

The 401 mock shape change strengthens rather than weakens: it asserts a specific error mapping the old generic body could not reach. The added `/v1/organization-invitations` stub is fixture fidelity only and is inert for pass/fail (removing it leaves all 6 tests passing).

That content is now on `main` via #1803.

PR #1816, opened 2026-08-24
feat(slack): route Slack conversations to a chosen workspace

benyhh:
Third step of per-channel and per-DM Slack workspace routing. The resolver is wired in behind `OPENGENI_SLACK_WORKSPACE_ROUTING_ENABLED`, **default off**. With the flag off the resolver short-circuits to the installation's own workspace before any new read, so an existing install issues exactly the queries it issued before.

## The decision order

Made from durable facts before any agent runs, so it can never be a model judgement. First match wins, and the order is deliberate: continuity beats an explicit override beats configuration beats derivation beats asking.

1. **A mapped thread keeps its workspace**, unconditionally, including after the channel has been re-pointed. Otherwise a live conversation would change tenant underneath the people in it.
2. **A strict `in <workspace>:` prefix**, parsed only at byte 0, matched case-insensitively against the exact label of a workspace the person can already start work in. A prefix that names nothing recognizable is a **refusal**, not a suggestion that falls through. It applies to one message and never writes a route row.
3. **The channel's remembered answer.**
4. **A direct message** uses an explicit DM route, or the person's own personal workspace derived from an active organization membership pointer. The id is derived, never accepted from a Slack payload or a route row.
5. **A sole candidate**, because one workspace is not a choice. This is what keeps the flag quiet for installs that only ever had one workspace even after it is turned on.
6. Genuine ambiguity keeps the installation's workspace until the first-use picker lands.

A person who cannot reach the routed workspace is **never** quietly served from somewhere else. Every refusal says what happened and ends `No session was created.`

## Migration 0336

Two storage changes routing needs once the routed and installation workspaces can actually differ.

**Action handle tenancy.** A button click arrives on an inbox row carrying the installation's tenancy and names one handle by id, while the handle lives in the workspace that owns its session. `slack_interaction_action_handles` carries a RESTRICTIVE `session_visibility_isolation` policy that resolves that session under the *current* scope, so no permissive policy can open it for a cross-workspace probe without weakening session visibility itself. Instead a private, ids-only mapping is maintained by a trigger on the handle table and read by one definer routine; the runtime role has no privileges on it. Without this every routed Status/Stop/Approve click reads under the wrong scope, finds nothing, and is refused as an invalid handle.

**Shared task origins.** `slack_shared_task_origins` carried two composite foreign keys on the same `(account_id, workspace_id)` pair: one to `slack_interactions` and one to `slack_task_policy_revisions`. The interaction is the routed workspace and the Slack task policy stays home by explicit decision, because it governs what may be read *out of* the Slack conversation rather than who the task belongs to. Once the two differ, no value of that pair satisfies both constraints and every private-handoff task in a routed workspace fails at insert. The frozen revision now carries its own tenancy pair.

Both are rolling: the probe is additive, and the new columns are nullable with existing rows backfilled inside the owner-only posture window, so an old image that writes neither keeps working exactly as it does today.

## Also

The Slack access-request path no longer denies a personal-workspace owner. A managed human's own personal workspace deliberately has no `workspace_memberships` row, so the bare join would have told people to request access to their own workspace once a DM can land there.

## Proof

Five real-Postgres scenarios with the flag on: a routed channel puts the interaction, session and events in the chosen workspace while the bot still posts on the installation's own credential; a mapped thread survives a re-point; a prefix overrides one message and strips only the addressing; a denied prefix creates nothing anywhere; and a single-workspace install notices no difference. The migration's cross-workspace handle probe is proven through `acquireOwnerMigratedTestDatabase`, because the shared harness hands out the container superuser, for whom `FORCE ROW LEVEL SECURITY` never engages and a definer probe looks correct while returning zero rows in production.

Deferred to the picker change: the first-use `ask` branch, and minting the access-request link for a target workspace. Until then the no-access refusal names the workspace and points at an administrator rather than offering a link that would not yet resolve.


benyhh:
An independent five-lens adversarial panel reviewed this diff cold. **19 findings survived verification and all are fixed here**; 28 were rejected. Two would have shipped broken:

**The migration would have aborted on any real deployment.** `slack_shared_task_origins` carries an unconditional `BEFORE UPDATE OR DELETE` immutability trigger, so the backfill raises rather than matching zero rows. The window now relaxes *both* owner-bound guards, the RLS posture and the immutability trigger, and restores both. The value written is the one each row already implied, because before routing the policy revision was always frozen in the row's own workspace, so no evidence changes meaning.

**The `in <workspace>:` override could never fire on a channel mention.** Slack delivers an `app_mention` with `<@U0BOT>` still in the text and OpenGeni stores it verbatim, so parsing at byte 0 of the raw text silently ignored every override typed the natural way, which is exactly the failure the feature exists to prevent. The prefix is now parsed after a leading mention of this bot, and the mention is put back in the message the model sees. My own tests concealed this by using mention-free `app_mention` payloads that Slack does not emit.

The rest:

- **Rollback broke live threads.** Turning the flag back off discarded a mapped thread's tenancy, so a reply addressed a session in a workspace the grant did not cover, failed, and settled with no message to the user. A mapped thread now keeps its workspace whether the flag is on or off, at no extra read.
- A **message shortcut and a reaction carry someone else's message**, so a prefix found there was never an instruction. Only text the invoking person typed can address anything.
- A **direct message from someone with no workspace of their own** was refused with a message saying they had none at all. It now falls through to the ordinary rules.
- **Two workspaces whose labels differ only by case** are a refusal, not a coin flip, and a bare address with no request is not an override.
- **`publishSlackSharedResult`** resolved the task policy and built the bot client under the routed workspace. The credential is the installation's; the policy is the one the origin froze.
- An **old-image origin row** carries no policy tenancy; a replay from a new image supplies it, which is not a change of evidence and no longer reports as one.
- **The flag had no deployment passthrough**, so an operator could not actually turn it on. It now has its own.
- Routing reads are **skipped entirely when a mapped thread already decides**, and the personal workspace comes from the candidate set rather than a second query for the same derived pointer.
- The **migration test asserted grants the test itself had made**. It now asserts the shape the migration produced for the real runtime role through the catalog, exercises the trigger's UPDATE arm, and pins that both guards are restored.
- New coverage for the **headline rule**: a channel routed to a workspace the person cannot reach creates no session in either workspace and says why.

Local: 177 Slack API tests, 51 db tests, 52 contract and deployment tests, lint, typecheck and all four repo guards green.

PR #1819, opened 2026-08-24
Fix deferred tool readiness and bounded recovery

jorgensandhaug:
## Summary

- make every actual local tool call join the shared deferred-preparation promise before dispatch
- attach the readiness rejection handler before publishing the promise
- bound automatic same-turn provider/MCP recovery and preserve distinct durable event producers across replacement attempts
- return a retryable catalog-not-ready response while Codemode preparation is still active

## Verification

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- focused runtime/API/worker unit tests: 257 passed
- Codemode catalog integration: 1 passed
- repeated required-MCP recovery/exhaustion integration: 1 passed

Candidate head is frozen at `7498e74367d2d29644ee5992680823531116a4cb`.


jorgensandhaug:
Refreshing base-owned source-admission evidence after the trusted controller fix merged; candidate head remains unchanged.

PR #1826, opened 2026-08-25
feat(slack): ask once where an unconfigured conversation should work

benyhh:
The first-use picker. The first message in a channel OpenGeni has no route for now asks, in the person's own bot DM, and remembers the answer. Behind the routing flag, which stays **off** by default.

## The shape

**Nothing is created anywhere while the question is open.** The request is parked on the prompt row and re-queued only once someone answers, so a card that is never answered leaves no session, no interaction and no half-started work.

**A bot DM, not an ephemeral message.** `chat.postEphemeral` returns no durable `(channel, ts)` and no `client_msg_id` echo, and the whole post ledger and its reconciliation are built on both.

**No ingest change beyond two action ids.** Each button's `value` is an option row id, which is a UUID precisely because `normalizedBlockActionInteraction` already requires that shape. The block-action path branches to the picker *before* the handle load, because a picker exists before any session and there is no handle row to find.

**One live card per conversation**, enforced by a partial unique index (migration 0342). The existing uniques are keyed to the originating event, so they stop Slack's retries of one event from posting twice but not two different messages in the same channel: the person would get two cards for one question, and answering either would write the route while the other stayed answerable. A direct message needs no separate index, because its channel id is already one per person.

**Answering commits three things together**: the frozen choice, the remembered route, and the re-queued request.

## The derived event id

The re-queued inbox row carries `<originalProviderEventId>:route:<promptId>`, not the original id. The original inbox row still exists and is settled `processed`, so reusing its id would collide with the inbox's own `(connection_id, provider_event_id)` dedupe and quietly queue nothing at all. That is exactly what migration 0335's comment described; the schema definition was corrected in the previous PR and this implements the corrected thing.

## Authority

The stored option rows are a prompt-time **snapshot and never authority**: the chosen workspace is re-authorized live, because access can be revoked between the question and the answer. A different Slack human clicking someone else's card is ignored, the same pinning ordinary action handles use. A failed authorization leaves the prompt answerable, since another workspace on the same card may still work.

## A real defect the tests found

The double-click scenario caught the card update seeding one operation id per `(prompt, actionId)`. A second click renders different bytes than the first, so under one id the post ledger's digest check would reject it and the entry would retry rather than saying the choice was already settled. The outcome is now part of the seed, matching the ordinary control-card path.

## Proof

Two new real-Postgres scenarios on top of the existing suite: the ask-once round trip (question asked, nothing started, Slack's retry does not post a second card, the answer starts the original request in the chosen workspace, the route is remembered with `source: picker`, and the next message never asks again); and a double-click that cannot produce a second session. Local: 69 Slack acceptance tests, 118 further Slack API and db tests, schema contract, lint, format, typecheck and all four repo guards green.

Remaining after this: the target-workspace access link, the per-message workspace line with its five operation-id seed bumps, the capability-sheet settings UI, and the docs rewrite.

### DIFF
--- a/.changeset/slack-route-picker.md
+++ b/.changeset/slack-route-picker.md
@@ -1,5 +1,6 @@
 ---
-"@opengeni/db": patch
+"@opengeni/db": minor
+"@opengeni/api-router": minor
 ---
 
-Ask once where an unconfigured Slack conversation should work, remember the answer, and re-queue the request that was interrupted by the question. One live card per conversation is enforced by a partial unique index, and the answer commits the choice, the remembered route and the re-queued request together.
+Ask once where an unconfigured Slack conversation should work, remember the answer, and re-queue the request that was interrupted by the question. One live card per person per conversation is enforced by a partial unique index, an aged-out card is settled by the writer rather than holding the slot, and the answer commits the choice, the remembered route and the re-queued request together.
--- a/.changeset/slack-route-picker.md
+++ b/.changeset/slack-route-picker.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/db": patch
+---
+
+Ask once where an unconfigured Slack conversation should work, remember the answer, and re-queue the request that was interrupted by the question. One live card per conversation is enforced by a partial unique index, and the answer commits the choice, the remembered route and the re-queued request together.
--- a/apps/api/src/integrations/slack-interactions.ts
+++ b/apps/api/src/integrations/slack-interactions.ts
@@ -1374,6 +1374,24 @@ async function postSlackRouteRefusal(
   });
 }
 
+/**
+ * A workspace name as a Slack button label.
+ *
+ * Slack renders a button label as plain text and the stored option snapshot caps
+ * it at 128 bytes, so bound it once at the boundary: the button, the row and
+ * every outcome card then agree. Counted in UTF-8 bytes, because that is what
+ * the CHECK counts.
+ */
+function boundedSlackRouteLabel(label: string): string {
+  const encoder = new TextEncoder();
+  if (encoder.encode(label).length <= 72) return label;
+  let bounded = label;
+  while (bounded.length > 0 && encoder.encode(`${bounded}...`).length > 72) {
+    bounded = bounded.slice(0, -1);
+  }
+  return `${bounded}...`;
+}
+
 /** How long a first-use picker stays answerable. */
 const SLACK_ROUTE_PROMPT_TTL_MS = 24 * 60 * 60 * 1000;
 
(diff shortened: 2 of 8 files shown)


benyhh:
An independent five-lens adversarial panel reviewed this cold. **26 findings survived verification**, and they were not separate bugs so much as one structural mistake with several faces. All are fixed; 10 were rejected.

## The structural problem

The pending index was channel-wide, nothing ever wrote the `expired` status, and the prompt row committed before its card was posted. Together, **any** of a transient Slack error, an unanswered card, or a colleague asking first turned that channel into one that silently swallowed every later message forever: no card, no session, no error, and no way out but an admin-written route or the `in <workspace>:` prefix on every message.

CI was green throughout. It could not catch this: it needs the flag on plus a failure or timeout path.

## What changed

- **The pending index is keyed per person per conversation**, not per channel. A shared channel has many people in it, and asking one of them must not swallow another's request. A DM is already one channel per person, so the same index still covers both surfaces.
- **`expired` has a writer** on every path that observes expiry: opportunistically in the opener, immediately before it tries to open the next prompt, and in the answer path when a click arrives after the deadline. `now()` is not immutable, so a partial index cannot exclude an aged-out row; the writer has to.
- **The card is posted unconditionally** rather than only on a fresh insert, and rendered from the prompt row's own stored options rather than the live candidate list. The operation id is persisted on the prompt, so a replay reproduces identical bytes and the ledger returns the completed original instead of posting twice. A failed post now repairs itself on the next attempt.
- **A denied click no longer replaces the card.** The comment promised to leave the prompt answerable; stripping every button while leaving the row pending did the opposite. The denial is a separate short message and the remaining options stay usable.
- **Cancel no longer claims nothing was created** when something already was.
- **A route click is refused when the flag is off.** A card can outlive the flag being turned back off, and honouring it would write a route row and start work through a disabled path.
- **The update operation id includes the chosen option**, because two different denied labels rendered different bytes under one id.
- **A workspace name is bounded once at the boundary**, so the button, the stored snapshot and every outcome card agree and the 128 byte CHECK cannot be hit.

## Test corrections the panel was right about

The "Slack retries the same event" step drained nothing and would have passed with the retry logic entirely absent; it now asserts `drainAll === 0` and adds a genuinely *different* message in the same channel, which is the case the index actually governs. The double-click test picked its `action_ts` values at random and collided about one run in nine.

New coverage: expiry (an aged-out card is settled and a fresh question asked), two people in one channel, and a migration test for the index itself.

One pleasing result: the two-person test initially failed because the second person had access to only one workspace, so the sole-candidate rule answered for them without asking. That is exactly the intended behaviour, and the test now grants both people a real choice.

Local: 71 Slack acceptance tests, 107 further Slack API and db tests, 4 migration tests, schema contract, lint, format, typecheck and all four repo guards green.

PR #1830, opened 2026-08-25
feat(slack): say where routed work went

benyhh:
Every acknowledgement and delivery for a routed task now carries a quiet `-> Workspace` line, so work never lands somewhere surprising.

## When it appears

Only when the routing decision was a **genuine choice**. An installation whose people have exactly one workspace made no choice, and a constant footer on every message there would be noise rather than information, so the label is null and nothing is added. That also keeps every existing install byte-identical while the flag is off.

## The label is frozen, not looked up

Written on the interaction when it binds, read **after authorization** so a workspace name is never shown to somebody who could not reach it, and bounded to the column's own 128 byte cap.

It is deliberately not resolved at post time: a workspace rename between the original post and a reconciliation would make the post ledger's byte comparison raise, and the delivery would never settle.

## All five operation-id seeds move to v2 in this same commit

This is the load-bearing part. The ledger binds one operation id to a digest over the text and blocks, so adding the line under an unchanged id would make **every interaction with a claimed-but-unposted delivery row wedge permanently** on a digest that can never match again. The ledger is keyed `(workspace_id, connection_id, operation_id)`, so a v2 seed claims a fresh row and the v1 rows stay completed and inert.

The five: the `postDelivery` default seed, the three literal delivery seeds (approval, human input, final), and the acknowledgement seed shared by its two call sites.

## Two call sites, deliberately

`postDelivery` funnels progress, approval, human input, final, failed and cancelled; `acknowledgeSlackSession` is a task's first message. Control cards and post-click updates are same-thread follow-ups to a message that already carries the line. The approval card lets its `text` and its `blocks[0]` diverge, so the line is appended to each independently rather than assuming they match.

`SlackMessageBlock` gains the `context` variant already modelled for App Home. Cost: one block of fifty, about 150 bytes of 32 KiB.

## Proof

A new scenario asserts the line on both the notification text and the blocks for a routed task, that the frozen label is stored, and that a single-workspace install on the same code path with the same flag gets no line and a null label. Local: 72 Slack acceptance tests, 136 further Slack API and db tests, lint, format, typecheck and all four repo guards green.

Remaining after this: the target-workspace access link, the capability-sheet settings UI, and the docs rewrite.

### DIFF
--- a/.changeset/slack-workspace-line.md
+++ b/.changeset/slack-workspace-line.md
@@ -1,0 +1,6 @@
+---
+"@opengeni/db": patch
+"@opengeni/api-router": patch
+---
+
+Add a quiet "-> Workspace" line to Slack acknowledgements and deliveries when routing actually chose a workspace, and bump all five Slack post operation-id seeds to v2 in the same change so no interaction with a claimed-but-unposted delivery row can wedge on a digest that will never match again.
--- a/apps/api/src/integrations/slack-bot.ts
+++ b/apps/api/src/integrations/slack-bot.ts
@@ -207,6 +207,11 @@ export type SlackMessageBlock =
         style?: "primary" | "danger";
       }>;
     }
+  | {
+      type: "context";
+      block_id?: string;
+      elements: Array<{ type: "mrkdwn" | "plain_text"; text: string; emoji?: boolean }>;
+    }
   | { type: "divider" };
 
 export type SlackHomeBlock =
--- a/apps/api/src/integrations/slack-interactions.ts
+++ b/apps/api/src/integrations/slack-interactions.ts
@@ -1392,6 +1392,65 @@ function boundedSlackRouteLabel(label: string): string {
   return `${bounded}...`;
 }
 
+/**
+ * Say where the work went, quietly.
+ *
+ * The line exists to stop work landing somewhere surprising, so it appears only
+ * when the routing decision was a genuine choice: a null label means routing is
+ * off, or the person had exactly one workspace anyway, and a constant footer on
+ * every message would be noise rather than information.
+ *
+ * The label is frozen on the interaction rather than looked up here, because a
+ * workspace rename between the original post and a reconciliation would make the
+ * post ledger's byte comparison raise.
+ */
+function withWorkspaceLine(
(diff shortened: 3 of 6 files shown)


benyhh:
An independent three-lens adversarial panel reviewed this cold. **12 findings survived verification**, all fixed; 10 rejected. One was a blocker, and the fix for it collapses most of the rest.

## Versioning the seed on the label, not on the release

The panel's core insight: bumping every seed **unconditionally** trades one hazard for another. Only a *labelled* interaction renders different bytes. For every pre-existing and unlabelled one the bytes are identical, so moving their id orphans their ledger row — a repair or a cross-deploy retry then posts a **duplicate** instead of finding the completed original, and any action handle still points at the operation its card was actually posted under. That last one is the blocker: it is how the approval, human-input and shared-result cards would have broken.

The label is written once at insert and never updated, so it is a stable discriminator. The seed is now `…:v2` only when the interaction carries a label. A labelled interaction can only have been created by an image that already renders the line, so it can never collide with an older image, and every unlabelled interaction keeps its exact ledger identity. That also puts the acknowledgement comparison carrying the one-time onboarding hint back on the id it was posted under, which the unconditional bump had silently broken.

## Also fixed

- **The reaction path never froze a label.** A reaction routes exactly like a mention, so a reaction summoned into another workspace said nothing about where the work went. It now names it, and its acknowledgement carries the line under the same conditional seed.
- **The terminal coalescing update and the control-click update both stripped the line** off a message that already carried it.
- **`thread` is gone from the naming set.** A mapped thread reuses the interaction that already exists, so it inherits that interaction's frozen label and never reaches the writer. Keeping it there also forced a needless `getWorkspace` read.
- The bounding comment claimed 128 bytes while the code applies the picker's own button cap, which is deliberate so the label and the button that produced it agree.

## New coverage

Unit tests for the seed rule (moves for a labelled interaction, does not for an unlabelled one) and for the line itself, including that a body already at the character cap still leaves room for the line and that two renders of one input produce **identical bytes** — the exact property the ledger's digest check depends on.

Local: 156 Slack API tests, 30 routing unit tests, lint, format, typecheck and the repo guards green.

PR #1834, opened 2026-08-25
feat(slack): request access to the workspace a conversation actually uses

benyhh:
A person routed into a workspace they cannot reach was told to find an administrator, because the access-request link could only ever name the installation's own workspace and sending them to a page that refuses them is worse than saying nothing. Now the link names the workspace they actually need.

## The relaxation

Both intent routes asserted `route.workspaceId === workspaceId`, which was exactly right before routing existed: a token could only ever name the installation's workspace. The assertion becomes:

- the installation binding resolves for the token's team,
- it is the same connection the token was minted against, and
- the named workspace belongs to that installation's organization.

That is the same fence the route tables carry (`target_account_id = account_id`), applied to the link flow.

## What this does not do

It widens no access by itself. It decides which workspace an access **request** may be raised for; the request still goes through the ordinary approval path, the identity link is still required, and `requireManagedSlackLinkHuman` — the only cookie proof in this flow — is untouched. `membershipAllowsSlackLink` already learned about personal workspaces in an earlier PR, so a person whose DM lands in their own workspace is not told to request access to it.

## Proof

The refusal scenario now parses the minted token out of the message and verifies it names the **routed** workspace, rather than only asserting the message mentions access. Local: 72 Slack acceptance tests, 96 further Slack API and db tests, lint, format, typecheck and the repo guards green.

Remaining after this: the capability-sheet settings UI, and the docs rewrite.

### DIFF
--- a/.changeset/slack-target-access-link.md
+++ b/.changeset/slack-target-access-link.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/api-router": patch
+---
+
+Let a Slack access-request link name the workspace a routed conversation actually works in, instead of only the installation's own. Both intent routes now assert that the installation binding resolves for the token's team, that it is the same connection the token was minted against, and that the named workspace belongs to that installation's organization.
--- a/apps/api/src/integrations/slack-interactions.ts
+++ b/apps/api/src/integrations/slack-interactions.ts
@@ -79,6 +79,7 @@ import {
   listSessionEventPage,
   listSessionHumanInputRequests,
   listPendingSlackUserLinkAccessRequests,
+  slackUserLinkAccessRequestTeam,
   rekeySlackInteractionRoute,
   reopenSlackInteractionDelivery,
   reserveSlackInteractionActionHandles,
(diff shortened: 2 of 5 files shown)


benyhh:
An independent adversarial panel reviewed this cold and found a **blocker**, reproducing the mechanism against real Postgres. Widening which workspace an access-request token may name was only half the change; without the other half, the link this PR adds is a guaranteed dead end.

## What was wrong

`slack_bot_user_links` is unique on `(connection_id, slack_user_id)` globally but RLS-visible only under **one** workspace, and the Slack pump reads it under the installation's. A routed request ran its whole transaction under the **routed** workspace, so:

- **When the person already has a link at home** — which is always true on the only path that reaches this refusal, because it sits after the not-linked guard — the availability check cannot see that row, the insert collides with it invisibly, the recovery select is filtered to nothing, and approval throws `slack_identity_conflict`. The transaction rolls back, no membership is created, and the admin is told the link is invalid.
- **When they do not**, the row is filed in the routed workspace, the route reports success, and every later Slack message asks them to link again forever.

CI was green. It could not catch this: it needs the flag on, a routed refusal, and an approval.

## The fix

The identity link stays a home fact, as its own read site and the routing migration both say. The availability check and the write now run under the installation's tenancy, with the tenant GUCs swapped for exactly those two statements and restored afterwards. They stay inside the one transaction on purpose: an approval that granted membership without linking the identity, or the reverse, would be worse than either. The request row and the `workspace_memberships` grant remain in the routed workspace, which is where the work runs.

The approver resolves the installation from the request row's own Slack team, through a new content-free reader, because the public request shape deliberately omits it.

## Also fixed

A message that embeds a signed token and is posted under a deterministic operation id must render stable bytes, or the ledger refuses the retry as a conflicting request. The token's mint time is now the originating inbox row's creation time rather than the wall clock. **The pre-existing not-linked message had the same latent hazard** and takes the same treatment, since leaving one instance would be odd.

## Proof

A new real-Postgres test approves a request raised for a routed workspace and asserts both halves: the grant lands in the routed workspace, and the identity link is still at the installation where the pump reads it. It fails without the fix (verified by removing the parameter). Local: 200 Slack API and db tests, lint, format, typecheck and the repo guards green.

PR #1838, opened 2026-08-25
feat(slack): set channel routing from the Slack capability sheet

benyhh:
An administrator can now see which OpenGeni workspace each Slack channel starts work in, and change it, from the sheet the rest of the Slack installation is managed in. It goes here rather than in organization settings so a later navigation change to that area does not churn it.

## The surface

The access block already listed every channel OpenGeni has been invited to, so each row gains where its work lands:

```
What OpenGeni can see                                  [ Manage routing ]
  All public channels        searchable without joining
  #eng-platform              invited · starts work in Platform
  #random                    invited · asks once, then remembers
```

A channel with no choice is **not broken**, and the copy does not imply it is: it asks the first person who uses it and remembers the answer.

## Authority

Two routes beside the reaction-channel allowlist they sit next to.

- **Read** is gated on `connections:read`: seeing where work goes is not administration.
- **Write** is gated on `workspace:admin`, the same gate the reaction allowlist uses, because pointing a channel somewhere is an installation-administration act.
- A **null target clears the route**, which puts the channel back to asking.
- An admin may only point a channel at a workspace **they could start work in themselves**. The table's own CHECK already fences the organization; the route refuses before writing rather than surfacing a constraint error, and the dialog only offers workspaces that pass the same test, so it never promises something the API will reject.

## Reuse

The dialog clones the reaction-channel dialog's structure and **reuses its channel lister** rather than adding a second one. It re-reads after saving instead of patching local state, because the server decides which routes actually exist.

Direct messages render as a rule rather than a control: they always use each person's own workspace, and there is nothing per-channel to set.

## Proof

Two new sheet tests: an admin sees the routed target on one row and "asks once, then remembers" on an unrouted one, with the `Manage routing` affordance present; a member with only `connections:write` sees no affordance at all. Local: 15 sheet tests, 132 Slack API, db and SDK-parity tests, lint, format, typecheck and the repo guards green.

Remaining after this: the docs rewrite, which is the last step.


benyhh:
An independent three-lens adversarial panel reviewed this cold. **13 findings survived verification**, all fixed; 15 rejected. Most of them were **one mistake**: the dialog submitted every channel it knew about on every Save.

That single behaviour caused three separate problems:

- **A no-op Save rewrote the provenance of routes nobody touched** — somebody else's picker answer became an admin decision, with a bumped version.
- **The payload grew with the installation rather than with the edit**, so a large Slack workspace could not save at all past the contract's cap.
- Because each route was authorized *as it was written*, a channel someone else had pointed at a workspace this admin cannot reach **threw partway through, after earlier channels had already been committed**, behind a toast saying the save failed.

The dialog now sends only entries whose target actually differs from the stored one, and the route **authorizes every target in one pass before writing anything**, so a refusal cannot land on a half-applied batch. The refusal names the channel instead of leaving the admin to guess.

A route pointing somewhere this admin cannot reach is no longer a blank control that makes every save fail: it renders as its own name plus "set by someone else" and is excluded from the payload.

## The rest

- **Saving requires the load to have succeeded**, not merely finished. A draft built from a failed read would otherwise clear every route the install has.
- **The read is gated on `workspace:admin`**, matching the reaction-channel allowlist beside it. It names workspaces across the organization, including ones the reader may hold no grant in, so it is not an ordinary connections read.
- **The response says whether routing is switched on.** With it off the stored routes are inert, so the sheet shows no routing control and no per-channel destination rather than implying they apply.
- Workspace choices are filtered to the installation's own organization, which is the fence the table's CHECK enforces anyway.
- `targetWorkspaceName` is no longer capped at 256 characters, which would have rejected a workspace name the database happily holds; the route list is no longer capped below what the server can return.
- The write refuses a connection with no installation binding rather than falling back to an empty Slack team id.

## New coverage

`apps/api/test/slack-channel-routes.test.ts` drives the real routes through an authenticated app: both gates, a **mid-batch refusal leaving the legal half of the same batch unwritten**, a null target clearing a route, and another installation's connection refused. A sheet test proves the routing control disappears when routing is off.

Local: 132 API, web, SDK-parity and db tests, lint, format, typecheck and the repo guards green.

PR #1876, opened 2026-08-26
test(e2e): wait for visible timeline settlements

benyhh:
## Why

Protected-main run 32951042755 failed only the interaction browser lane because the delayed-underfill regression began sampling before the scheduled React prepend was visibly committed. The first frame could therefore capture the intentional pre-commit underfill while later frames captured the correctly parked live tail.

Repeated full-file proof also exposed the same setup race in the nested-row regression: its baseline used a fixed delay after programmatic navigation rather than the supported scroll settlement.

## What changed

- wait for the prepended row to exist before starting the 24-frame live-tail ledger
- wait for fonts before affected pixel baselines
- wait for scrollend, with a two-animation-frame fallback where unsupported, before the nested-row baseline
- retain every existing pixel tolerance, frame count, and single-load assertion

## Validation

- full 32-test timeline browser file: green six consecutive exact-worktree runs after the final synchronization repair, including one after formatting
- originally failed case: 12/12 root repetitions plus 10/10 independent-review repetitions
- companion nested-row case: 10/10 independent-review repetitions
- all 31 typecheck projects clean
- affected oxfmt, oxlint with warnings denied, and git diff checks green
- independent exact-head review PASS with no findings at 0d54a1a928cb9a42deaacb271529659439a92495

No product behavior, assertion threshold, deployment, or production state changes.

### DIFF
--- a/test/e2e/timeline-scroll.browser.e2e.ts
+++ b/test/e2e/timeline-scroll.browser.e2e.ts
@@ -298,10 +298,27 @@ describe("timeline scroll ownership browser regression", () => {
   test("keeps a nested row anchored when prepend merges into its activity group", async () => {
     await page.goto(`${baseUrl}/timeline-scroll-merge-test.html`);
     await page.waitForFunction(() => window.timelineMergeHarness !== undefined);
+    await page.evaluate(async () => {
+      await document.fonts.ready;
+    });
     const target = page.getByText("reasoning-50", { exact: true });
     await target.waitFor({ timeout: 15_000 });
-    await target.evaluate((node) => node.scrollIntoView({ block: "start" }));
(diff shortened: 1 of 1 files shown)


benyhh:
Independent exact-head review PASS by the Full Access OPE-365 browser reviewer on 0d54a1a928cb9a42deaacb271529659439a92495 over sole parent/current main 2cbccd13b04b1db0542051e1516df7efa81fd692. Review found no assertion weakening, product masking, arbitrary retry, portability issue, or hang risk. Independent evidence: full file 32/0/134; failed case 10/10; nested settlement case 10/10; exact impact plan selected all four browser lanes; format/lint/diff clean. Full live OPE-365 and PR #1842 feedback were reread before verdict.

PR #1878, opened 2026-08-26
fix(sandbox): make machine home and browser routing truthful

davletd:
## Summary

- stop advertising a synthetic managed-home target for selfhosted-home sessions and return a typed error for unsupported session/default swaps
- use a Connected Machine’s native browser control instead of bootstrapping the Linux headed-display stack on macOS
- recognize Modal’s exact workspace-root rejection and use the existing streamed private-file fallback without placing the browser token in argv

## Root cause

The staging deployment and current main still had three mismatches: selfhosted sessions exposed a nonexistent managed home, native Connected Machine browser sessions entered the Linux-only display setup path, and Modal’s current `escapes the workspace root` diagnostic was not classified as a workspace-confinement error.

## Validation

- `bun test apps/api/test/browser-session-routes.test.ts packages/runtime/test/browser-control-client.test.ts`
- `bun test apps/api/test/fleet-tools.test.ts apps/api/test/machines-routes.test.ts`
- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `git diff --check`

A fresh merge-tree against current `origin/main` is conflict-free; the only shared file is the architecture map, where the changes are independent.


davletd:
Superseded by #1929 (native browser placement and Modal private-write fallback) and #1930 (Connected Machine to managed sandbox switching). The replacement PRs are rebuilt from current main and separate the two root causes; #1930 preserves the intended explicit managed fallback instead of hiding/rejecting it.

PR #1880, opened 2026-08-26
perf(insights): make fact visibility set based

benyhh:
## Summary

- add bounded read-only Insights fact sources that validate exact account, workspace, subject, and current runtime protocol
- preserve ordinary fact-table RLS while replacing one nested session lookup per analytical fact with a set-based visible-session join
- route all Insights model/usage scans, including monthly cap meters, through the bounded sources without changing UTC, filter, billing, or cap semantics

## Safety

- PUBLIC cannot execute the new functions
- current application roles need existing SELECT rights on both fact tables before receiving EXECUTE
- mismatched or missing RLS context, stale runtimes, and windows over 370 days fail closed
- private-owner, workspace-shared, workspace-level, and orphan-reference behavior is covered against a real app role

## Verification

- real PostgreSQL integration test across every Insights fact helper
- 4,096-fact / 512-session EXPLAIN ANALYZE fixture requires at least a 3x improvement over scalar-RLS scanning
- repository-wide typecheck
- lint
- focused Insights, release-schema, and billing-safety tests

Linear: OPE-370


benyhh (commented):
CHANGES REQUESTED

1. [P1 security] packages/db/drizzle/0356_set_based_insights_session_visibility.sql:198-232 grants opengeni_private schema USAGE and SECURITY DEFINER function EXECUTE to every database role that happens to hold SELECT on both fact tables. That predicate does not establish that the role is an OpenGeni application login or that it previously had session visibility authority. I reproduced this with a synthetic fact_reader: before reapplying 0356 it had schema_usage=false, function_execute=false, sessions_select=false; after the migration it had schema_usage=true, function_execute=true, sessions_select=false. Because PostgreSQL custom GUCs are caller-settable and the function treats an empty subject as workspace-wide visibility, that role can now invoke a definer path that reads sessions/facts it could not read through ordinary session_reference_visible RLS. Restrict grants to the explicit configured application-role list, or require and prove the full pre-existing session-visibility privilege set. Add a negative migration test showing a non-application fact reader receives no new schema/function privilege.

2. [P1 integration] packages/db/src/insights.ts:119 and 673-729 conflict with current main commit 4ca447586, which added sessionDepth to RecentModelCallRow/listRecentModelCalls for semantic fallback labels. A fresh git merge-tree of origin/main and fae8e9ff0 reports a content conflict in this file, and GitHub currently marks the PR CONFLICTING. Resolve the overlap while preserving the sessionDepth field/projection and current main Insights label tests, then rerun the current-main merge tree and focused suites.

The follow-up fae8e9ff0 correctly adds 0356 to the canonical appendedMigrationPaths list; bun scripts/check-migration-schema-contract.ts now passes.

Verified on exact head fae8e9ff0: DB/core typechecks, Workspace billing guard, focused Insights/unit/release tests, and the real non-owner migration suite. The 4,096-fact performance assertion passed four consecutive fresh-database runs. The visibility fixtures cover owner, unrelated subject, subjectless workspace rows, bounds, old runtime rejection, and every converted helper. Those checks do not mitigate the grant widening above.

Verdict: CHANGES REQUESTED.

benyhh (commented):
APPROVED

Exact reviewed head: 6b8283eec3d55641355aa71c596437b212f33d1c.

The two prior blockers are resolved:

- packages/db/drizzle/0356_set_based_insights_session_visibility.sql:191-235 now derives grants only from opengeni.migration_application_roles while retaining the existing fact-table SELECT prerequisite. I repeated the synthetic unauthorized fact_reader probe around a migration replay. Its privileges remained schema_usage=false, function_execute=false, sessions_select=false before and after, while the explicitly configured app role retained schema_usage=true and function_execute=true.
- A fresh merge-tree against current origin/main 4ca447586 is conflict-free and GitHub reports MERGEABLE. The semantic session title/depth fields from #1871 remain present and its focused test passes.
- Migration 0356 is registered in the canonical appendedMigrationPaths list and the migration-schema guard passes.

Independent verification on this exact head:

- DB and core TypeScript checks: pass.
- Focused Insights, semantic-title, cache coverage, UTC enumeration, and release-schema tests: 17 pass.
- Real PostgreSQL non-owner migration suite: 5 pass, including bounded fail-closed behavior, owner/unrelated/subjectless visibility, all converted helpers, and the 4,096-fact performance plan comparison.
- Workspace/billing guard, migration-schema guard, full formatting check, and diff check: pass.
- Previous candidate performance test stability was also exercised across four consecutive fresh databases.

No remaining actionable review findings. This is a source approval only; merge should still wait for green required CI.

Verdict: APPROVED.

benyhh (commented):
APPROVED

Exact reviewed head: b700b7fd3eb0806859c9b2c45c57a19a10c48688. This supersedes my approval of 6b8283eec after the dependent test exposed SRF runtime decoding drift.

The new packages/db/src/insights.ts normalization is correctly scoped to listRecentModelCalls, the only converted helper returning raw SRF timestamptz/bigint detail fields. It:

- converts occurredAt and recordedAt to finite Date instances;
- converts all token and cost fields to finite numbers;
- preserves database nulls for nullable telemetry;
- fails closed on malformed values instead of allowing invalid dates or NaN into the core snapshot;
- retains semantic sessionDepth/title behavior from current main.

Independent verification on this exact head:

- Real PostgreSQL migration suite: 5 pass / 64 assertions. The non-empty recent-call fixture now proves Date runtime values and numeric token/cost values while retaining owner/unrelated visibility.
- Non-empty core Insights projection plus UTC/filter/cache/title tests: 10 focused tests pass.
- DB and core TypeScript checks pass.
- Current-main merge-tree remains conflict-free and GitHub reports MERGEABLE.
- Workspace/billing guard, migration-schema guard, full formatting check, and diff check pass.
- The previously approved configured-role grant repair is unchanged.

No actionable findings remain. Source approval is independent of the required green CI merge gate.

Verdict: APPROVED.

PR #1885, opened 2026-08-26
fix(db): restore Insights migration upgrade path

benyhh:
## Summary

- initialize the bounded application-role setting whenever a pending migration consumes it
- remove the stale filename checklist that stopped at migration 0353
- add a real PostgreSQL regression that physically migrates through 0355 and then applies 0356

## Why

The released 0356 Insights migration failed on normal existing installations with SQLSTATE 42704 because its application-role setting was absent. Fresh databases did not expose the problem because earlier pending migrations initialized the same setting.

## Safety

Released SQL migrations are unchanged. The repair preserves explicit application-role validation for dedicated-schema and custom-role deployments and does not broaden any database grant.

## Verification

- migration 0356 suite: 6 passed, 67 assertions
- migration 0338 migrator contract: passed
- migration 0257 explicit-role behavior: passed
- database package typecheck: passed
- focused lint and formatting: passed

Linear: OPE-370

### DIFF
--- a/packages/db/src/migrate.ts
+++ b/packages/db/src/migrate.ts
@@ -7,15 +7,7 @@ const DEFAULT_DATABASE_URL = "postgres://opengeni:opengeni@127.0.0.1:5432/openge
 const DEFAULT_MAX_NESTED_AGENT_DEPTH = 3;
 const MAX_NESTED_AGENT_DEPTH = 2_147_483_647;
 const DEFAULT_APPLICATION_DATABASE_ROLE = "opengeni_app";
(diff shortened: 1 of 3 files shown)


benyhh:
Independent AI review: APPROVED at exact head df6a14f7eb791cbc062769b38991304ef0932e54.

The reviewer found no actionable correctness, security, concurrency, embedded-schema, or privilege findings. It independently ran 18 focused migration/custom-role/schema/security tests with 217 assertions, plus the full 31-project typecheck. All passed. Merge remains gated on green CI.

PR #1889, opened 2026-08-26
fix(db): repair Insights FORCE RLS reads

benyhh:
## Summary

- replace the scalar per-fact FORCE-RLS path with a transaction-bound, read-only capability recognized only by the table owner
- preserve ordinary fact/session RLS and the original INSERT, UPDATE, and DELETE policy expressions
- expose one value-free policy predicate so existing unconfigured reporting roles keep ordinary reads without gaining analytical entry-point access
- add bounded provider/model and event-type overloads so selective predicates run inside the indexed analytical scan
- reset hostile table and function default ACLs, then authorize only configured application roles with both fact-table read grants

## Why

The released set-based functions still executed scalar session visibility once per fact under the production `NOSUPERUSER NOBYPASSRLS` owner. Empty UTC windows also failed, and outer provider/model/event filters could not use the selective indexes.

## Safety

The capability is private, bound to backend PID, transaction, tenant, subject, and initiating human, and deleted on success or error. Application roles cannot read, write, or forge it. The public policy predicate returns only a boolean for an unforgeable same-transaction capability. The migration is rolling and follows the 0358 managed GitHub App migration as 0359.

## Verification

- real PostgreSQL 0359 and 0356 suites plus release contract: 18 passed, 380 assertions
- unconfigured pre-migration reporting role retains direct FORCE-RLS reads but cannot execute either analytical entry point
- hostile pre-0356 table/function defaults, private capability ACLs, and all four analytical entry-point ACLs covered
- repaired FORCE-RLS scan remains more than 1.5x faster than the scalar path on the focused fixture
- generic-plan provider/model calls use the selective index
- database typecheck, format, diff, and release schema contract clean

Linear: OPE-370


benyhh:
Independent AI review: APPROVED for exact head `ca43ca5d80fed12460b377e83d616610476ee2c1`.

The review found no actionable security, correctness, migration-order, ACL, RLS, concurrency, test-validity, or performance-regression issues. It specifically rechecked the public value-free policy predicate, private capability and analytical entry-point boundary, unconfigured reporting-role parity, capability cleanup/non-forgeability, write-policy preservation, custom-schema behavior, selective-index path, 0358 to 0359 ordering, release contract, and Changeset.

Independent verification: 19 focused tests, 391 assertions, plus formatting/lint and all-project typecheck, all green.

PR #1890, opened 2026-08-26
test(react): await timeline jump control exit

benyhh:
## Summary

- assert the timeline has synchronously returned to semantic bottom-follow state
- wait for the animated Jump to latest exit node to detach before checking its final absence

## Why

The browser regression observed the button during its intentional 150 ms AnimatePresence exit even though scroll geometry and bottom-follow state were already correct. The same test-only race failed an unrelated database PR twice on an identical head.

## Safety

No product code, timeout, or fixed sleep changes. The wait is state-based and bounded by the existing test deadline.

## Verification

- independent AI review approved exact head `830cdf429`
- focused case passed 6/6 reviewer runs and 10/10 implementation stress runs
- React typecheck, targeted lint, format, and diff checks clean

### DIFF
--- a/test/e2e/timeline-scroll.browser.e2e.ts
+++ b/test/e2e/timeline-scroll.browser.e2e.ts
@@ -543,6 +543,11 @@ describe("timeline scroll ownership browser regression", () => {
 
     expect(immediate.maxScroll - immediate.scrollTop).toBeLessThan(2);
     expect(immediate.liveTailGap).toBeLessThan(48);
+    expect(await scroller.getAttribute("data-og-bottom-follow")).toBe("true");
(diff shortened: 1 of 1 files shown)


benyhh:
Independent AI review: APPROVED for exact hash `830cdf429e6bcc876678425873346f28b30ab331`.

No actionable findings. The immediate `data-og-bottom-follow=true` assertion proves product pin state before the wait. The accessible Jump to latest node is therefore only AnimatePresence retention during the intentional 150 ms exit. The state-based detached wait adds no sleep or timeout inflation, and the final zero-count and load-count assertions preserve the original regression intent. Focused reviewer runs: 6/6 passing, about 490–514 ms each.

PR #1891, opened 2026-08-26
perf(insights): bundle analytical reads

benyhh:
## Summary

- collapse the Insights model path from nine fact-table reads to two by default, or three when filtered facets are requested
- collapse the usage path from nine reads to three exact materialized windows
- push provider, model, and event-type filters into the RLS-safe analytical functions added by #1889
- parallelize independent dashboard reads and bound session lookups by distinct selected sessions
- return a deterministic 400 for provider/model filter values outside the database UTF-8 byte envelopes

## Safety and compatibility

- preserves shared/private visibility, UTC bucket boundaries, strict monthly cutoff semantics, and empty equality windows
- retains future UUID versions in warm-group labels
- does not claim to eliminate all PostgreSQL temporary I/O; production p95 and temp-I/O remain explicit follow-up measurements
- includes a Changeset for API router, core, and database packages

## Validation

- 24 focused tests, 185 assertions, including real PostgreSQL parity and scan-count checks
- API, core, and database typechecks
- targeted lint and format checks
- workspace/billing static guard, docs reference guard, and diff check

Linear: OPE-370


benyhh:
Independent AI review: APPROVED for exact commit 0c676077dc6daf8fbd9bab726e3b313d3de773a7.

The reviewer confirmed all 13 changed artifact blobs are byte-identical to the previously audited candidate, the parent is exactly merged main at 03d1c6e02e854a465b518149f32d6657cbbd4ebf, and independently reran 24 focused tests with 185 assertions. No correctness, security, tenancy, migration-integration, performance-claim, or packaging findings remain.

Production p95 and PostgreSQL temporary I/O remain explicit post-deployment measurements; this review does not claim the patch eliminates all temporary I/O.

PR #1918, opened 2026-08-27
test(db): stabilize learning-policy interruption fence proof

jorgensandhaug:
## Summary

- acquire the exact attempt `FOR UPDATE` before creating the synthetic interruption receipt
- fail immediately if the writer exits before reaching the expected lock wait
- keep the regression focused on the snapshot transaction retaining its attempt `SHARE` fence through commit

## Context

This is the focused test-harness follow-up to #1917. The migration and runtime behavior are unchanged; the first public exact-head run showed that the synthetic writer could exit before reaching the lock the test intended to observe.

## Verification

- focused `oxfmt --check`
- `git diff --check`
- `bun scripts/check-migration-test-budgets.ts`

The real PostgreSQL proof runs on the public PR unit matrix.

## Summary by Sourcery

Stabilize the learning-policy interruption lock-order regression proof by explicitly synchronizing on the writer’s backend lock wait.

Enhancements:
- Strengthen the learning-policy interruption regression test so it observes the exact attempt lock fence and verifies the snapshot transaction retains its share lock through commit.

Tests:
- Make the PostgreSQL lock-order test fail immediately when the synthetic interruption writer completes before reaching the expected row-lock wait.

### DIFF
--- a/packages/db/test/migration-0364-workspace-learning-policy-snapshot-lock-order.test.ts
+++ b/packages/db/test/migration-0364-workspace-learning-policy-snapshot-lock-order.test.ts
@@ -288,6 +288,14 @@ describe("migration 0364 workspace learning-policy snapshot lock order", () => {
 
       interruptionCall = interruptionWriter.begin(async (tx) => {
         await tx`
+          select id from session_turn_attempts
+          where account_id = ${fixture.accountId}
+            and workspace_id = ${fixture.workspaceId}
+            and session_id = ${fixture.sessionId}
+            and turn_id = ${fixture.turnId}
+            and id = ${fixture.attemptId}
+          for update`;
+        await tx`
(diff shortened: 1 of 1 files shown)


jorgensandhaug:
Merge evidence for exact head `a0acc0663`:

- Source and type contracts: green
- Unit shards 1–6: green; shard 2 contains the corrected real-PostgreSQL snapshot/interruption fence proof
- Real-service shards 1–6, recovery, package, deployment, and impacted E2E: green
- Sourcery: green

The only red optional lane is WebKit browser account acceptance under deliberate response-body-loss injection. It failed across reruns on changing, unrelated cancellation/503 counts while the Chromium and Firefox account lanes passed. This PR changes only the PostgreSQL migration test harness; the WebKit determinism issue is tracked separately from this merge.

PR #1920, opened 2026-08-27
test(db): keep lock probe application name exact

benyhh:
## Summary
- keep the migration 0364 pg_stat_activity probe below PostgreSQL's 63-byte application_name limit
- preserve the exact blocked-row-lock assertion instead of timing out on the truncated identifier

## Verification
- OPENGENI_REQUIRE_REAL_DB=1 bun test --timeout 180000 packages/db/test/migration-0364-workspace-learning-policy-snapshot-lock-order.test.ts (3 consecutive passes)

## Summary by Sourcery

Bug Fixes:
- Ensure the migration lock probe's application name remains within PostgreSQL's length limit so the blocked-row-lock assertion matches reliably.


benyhh:
Superseded by #1918, which fixes the same PostgreSQL lock-observation issue using backend PID identity.

PR #1988, opened 2026-08-28
OPE-407: Fix Connected Machine targeting for schedules and child workers

benyhh:
## Summary

- bind generated-session schedules to either managed execution or an exact Connected Machine, including an optional working directory
- automatically inherit a parent session's active Connected Machine and working directory when a child worker omits placement; no extra machine choice is required for ordinary worker spawning
- seed and validate machine routes before the first turn so sessions cannot start with an unbound self-hosted client
- reject targetless self-hosted-only creation with an actionable 422 explaining that self-hosted execution runs on a Connected Machine
- expose the client-safe deployment sandbox default so the schedule form can auto-select the available machine when managed execution is unavailable

## Testing

- [x] `bun run typecheck` (31 projects)
- [x] `bun run lint`
- [x] `bunx oxfmt --check packages/core/src/domain/sessions.ts apps/api/test/machine-home-backend.test.ts`
- [x] `bun test apps/api/test/machine-home-backend.test.ts apps/api/test/sandbox-shared-and-viewer.test.ts apps/api/test/sandbox-consent-gate.test.ts` (55 passed, 1 optional live-Modal test skipped)
- [x] `bun test apps/web/src/lib/scheduled-tasks.test.ts packages/contracts/test/contracts.test.ts apps/api/test/scheduled-task-existing-session.test.ts apps/worker/test/scheduled-task-machine-target.test.ts`
- [x] `bun test packages/sdk/test/contract-parity.test.ts`
- [x] `bun test apps/api/test/app.test.ts --test-name-pattern "GET /v1/config/client"`
- [ ] full `bun test` suite was not run

## Behavior

- Spawned child workers keep the low-friction default: omitting sandbox placement shares the creator's current box. For Connected Machines, the backend now copies the exact session-local machine pointer and working directory before the child's first turn.
- Scheduled tasks that create independent chats have no live parent to inherit from. They therefore retain a durable target. The form automatically selects the first available machine on self-hosted-only deployments; users choose only when there are multiple possible destinations or managed execution is also available.
- Existing-session schedules do not select compute again because the existing session already owns its route.

## Notes

- No database migration is needed; scheduled targets use the existing agent-config JSONB, and child inheritance uses the existing active-sandbox pointer.
- Unattended schedules intentionally exclude user-scoped personal machines until a durable scheduled personal-resource grant flow exists.
- Live desktop UI inspection was not run because this clean worktree has no local `.env` and no API/web stack was running; form projection and type coverage passed.

## Summary by Sourcery

Bind generated schedules and nested workers to valid Connected Machine routes while failing closed when self-hosted execution cannot be established.

New Features:
- Allow generated-session schedules to target a specific workspace or organization Connected Machine with an optional working directory.
- Expose Connected Machine execution choices in the schedule editor, including automatic selection on self-hosted-only deployments.
- Inherit a parent session’s Connected Machine route and working directory for child workers that omit placement.

Bug Fixes:
- Reject self-hosted sessions and schedules without a bound Connected Machine before execution begins.
- Fail scheduled runs and reclaim unstarted generated sessions when their Connected Machine route cannot be established.

Enhancements:
- Validate machine ownership, availability, liveness, and workspace-root readiness for scheduled execution.
- Prevent machine targets from being used with existing-session schedules and normalize scheduled working-directory inputs in the SDK.

Documentation:
- Document Connected Machine routing rules for generated schedules and nested workers, including the exclusion of personal machines from unattended schedules.

Tests:
- Add coverage for child route inheritance, self-hosted placement validation, scheduled machine targeting, route-failure recovery, SDK normalization, and contract parity.

Chores:
- Update the web bundle budget for the schedule targeting UI and SDK additions.


benyhh:
Scheduled PR review session — blocking findings on exact head `daa4a05b1d62c5d322e5ddaaabac736ff54b5deb`:

1. **The selected Connected Machine is not seeded on the normal first dispatch.** `recoverBoundScheduledTaskDispatch` is only called for an already-existing queued run (line 278), and the sole `swapActiveSandbox` call is inside that recovery helper (line 1988). The ordinary dispatch creates the generated session and calls `initializeSessionStartAtomically` at lines 1116/1179 without first setting `activeSandboxId` or `workingDir`. Therefore a newly accepted occurrence can start with no exact machine route, contrary to the PR description and the safety invariant for unattended execution.

2. **The recovery seed-failure path leaves a durable orphan.** If `swapActiveSandbox` returns `swapped: false`, the generated session is already created and bound to the run, but the code throws and only marks the run failed. This matches the unresolved Sourcery finding. Cleanup must preserve replay semantics: cancellation/abandonment and `scheduled_machine_unavailable` run settlement need one durable, idempotent authority boundary (or an equivalent marker), otherwise a crash after cancellation but before run failure can later reinterpret the scheduler-cancelled session as a user cancellation and mark the run skipped.

These two paths should share one tested create/bind/seed/start lifecycle so first dispatch and queued recovery cannot diverge. A source-string ordering test of the recovery helper does not cover the ordinary path or failure settlement.

CI is also currently red because this public package change has no changeset (`Source and type contracts`). I did not push a fix from this scheduled session because repairing both lifecycle paths and their post-commit event/wake behavior is broader than a safe localized patch. No linked Linear issue is present in the PR body for authority/acceptance-criteria verification.

benyhh:
Tracking: OPE-407. The repair now routes both normal first dispatch and queued recovery through one fail-closed machine seeding helper. A failed route atomically fails the queued run and reclaims an exact unstarted scheduler-generated session; a concurrent user cancellation remains cancelled while the run is classified as scheduled_machine_unavailable. It also preserves the selected machine OS instead of hard-coding Linux. Verified after merging current main: all 31 typecheck projects and 50 focused scheduled-task/real-PostgreSQL tests pass.

benyhh:
@sourcery-ai review

benyhh:
Closed the follow-up review/CI gaps on exact head `cf64bf195`: the SDK now trims Connected Machine working directories and omits whitespace-only values before sending; source hygiene and the merged-main self-hosted error assertion are reconciled; and the measured Linux/x64 production graph has a policy-derived one-KiB-headroom budget. SDK/API focused tests, public-hygiene guard, full web production build/budget, formatting, lint, and relevant typechecks pass locally. @sourcery-ai review

PR #1987, opened 2026-08-28
fix(web): synchronously release rotated actor readers

benyhh:
## Summary
- cancel locked native stream readers in the same task as actor rotation or document teardown, after the native fetch signal is aborted
- terminate authorization-revoked SSE responses with a clean fail-closed close; ordinary delivery faults still terminate as errors
- explicitly retire every completed HTTP/1 SSE transport so direct Bun/Docker deployments do not retain stale streams in Chromium's six-connection per-origin pool
- wait for responsive account-menu refresh reads to settle before applying the strict final browser ledger
- correct SSE authorization tests so the candidate live event is durable before reauthorization blocks delivery

## Verification
- 39 focused web/API stream tests (158 assertions)
- final transport-retirement implementation: 4 complete Chromium real-PostgreSQL multi-account journeys, including 3 consecutive repetitions
- prior exact stream-close implementation also passed Firefox and WebKit; no browser-specific product behavior changed in the final patch
- all 31 TypeScript projects
- repository-wide formatting and lint

## Hosted failure evidence
Three exact hosted Chromium runs reproduced HTTP/1 connection-pool starvation. The ledgers showed authorization-fenced streams from replaced actors remained attached to reusable Bun HTTP/1 sockets while current-actor finite reads waited behind Chromium's per-origin connection cap. This head closes revoked bodies without a stale frame and marks all SSE responses `Connection: close`, making completion retire the HTTP/1 transport; HTTP/2 front doors strip the hop-by-hop header. The exact pressure journey is now repeatedly green locally. Merge remains blocked until this exact head is independently green and reviewed.

PR #1990, opened 2026-08-28
fix(web): prevent HTTP/1 stream pool starvation

benyhh:
## Summary

- detect HTTP/1 browser transport and opt only those SSE requests into five-second clean connection cycles
- hold the logical wrapper for a bounded two-second grace after native EOF so queued finite reads take the freed slots before durable streams reconnect
- preserve long-lived HTTP/2 and HTTP/3 streams; JSON error responses are never delayed
- hide only brief planned reconnect indicators with a CSS-only delayed reveal
- extend account acceptance evidence to prove the bounded transport runs in every simultaneous browser context

Follow-up to #1987 after post-merge Chromium account acceptance reproduced browser-wide HTTP/1 connection-pool starvation.

## Verification

- 54 focused API/web/budget tests (209 assertions)
- all 31 TypeScript projects
- lint, formatting, diff, and production bundle budget checks
- five full Chromium account-acceptance runs; four ran in concurrent pairs for added contention
- full Firefox and WebKit account-acceptance runs concurrently
- every browser run: zero retired/stuck finite reads and bounded HTTP/1 dispatch evidence in all three simultaneous contexts


benyhh (commented):
Exact-head maintainer review for 28fbd94d5a0264550aadc6b65571b3eef8a81b6e: PASS.

Reviewed the complete 15-file diff against base 9787ca2ecbb09c7ca2bf411262ec5cd4e806ca58. The mitigation is scoped to browser-negotiated HTTP/1.0/1.1 event streams; HTTP/2, HTTP/3, unknown protocols, finite requests, and JSON error bodies keep their existing behavior. All four SSE route families share the validated five-second lifetime, cleanup remains idempotent, and the combined workspace stream does not leak the outer lifetime into its internal upstream channels.

On clean native EOF, the actor-bound wrapper waits an abortable two-second grace before exposing EOF to the durable SDK reconnect loop. The native response is already complete during that grace, so connection slots are free for queued finite reads; actor rotation still settles fail-closed and interrupts the grace. Cursor replay preserves any events arriving during the window. The CSS-only delayed reconnect indicator adds no runtime state, and the calibrated whole-KiB envelope retains 1,445 raw bytes of measured headroom without weakening compressed, file-count, lazy, or CSS limits.

Exact-head local evidence: 54 focused API/web/budget tests with 209 assertions; five complete Chromium real-PostgreSQL account journeys, four under paired concurrent contention; complete Firefox and WebKit journeys under concurrent load; all 31 TypeScript projects; lint, format, diff, and production bundle checks. Hosted exact-head CI completed with 38 successful jobs, nine intentional skips, and no failures; all three independent account-browser lanes passed. Sourcery exact-head check passed with no new finding. No actionable issues found.

PR #1999, opened 2026-08-29
fix(accounts): reconcile precommit invalidations

benyhh:
## What

- distinguish a browser-account transition that is still inside its initiating host fence from one whose server mutation has started
- let a server/cross-tab invalidation supersede and reconcile the pre-request transition instead of leaving the provider neutral
- re-check cancellation after the deliberate peer-hold task yield so a superseded mutation cannot reach the server
- add focused regressions for invalidation during both the host fence and the peer-hold yield

## Why

Follow-up security review of #1998 found a remaining precommit race. `invalidateActor()` treated every pending commit as though its server mutation already owned settlement. If an authority hint aborted the initiating fence before `pending.execute()` began, `executePending()` returned early and no path restored the reconciled actor. An invalidation during the peer-hold yield could also reconcile the new actor and then still dispatch the stale mutation because there was no post-yield sequence check.

The regression reproduced the first case on merged main: `invalidateActor()` returned `null`, the mutation did not start, and the provider remained on a neutral loading surface.

## Verification

- `bun test ./packages/react/test/accounts.test.tsx` — 31 passed
- `cd packages/react && bun run typecheck`
- targeted oxlint — clean
- targeted oxfmt check — clean
- `git diff --check` — clean

## Scheduled session provenance

This follow-up was found and implemented by the scheduled PR review session on August 29, 2026. No Linear issue was linked from #1998.

## Summary by Sourcery

Ensure superseding account invalidations restore the reconciled actor and prevent stale mutations from reaching the server.

Bug Fixes:
- Reconcile browser account authority when invalidation supersedes a pending transition before its server mutation begins.
- Prevent stale account mutations from being dispatched after invalidation during the peer-hold window.

Tests:
- Add regression coverage for invalidation during the pre-mutation host fence and peer-hold yield.

### DIFF
--- a/.changeset/restore-precommit-account-invalidation.md
+++ b/.changeset/restore-precommit-account-invalidation.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/react": patch
+---
+
+Reconcile browser account authority when an invalidation supersedes the initiating fence before its server mutation starts.
--- a/packages/react/src/accounts.tsx
+++ b/packages/react/src/accounts.tsx
@@ -197,6 +197,7 @@ export function BrowserAccountsProvider({
     expectedGeneration: string;
   } | null>(null);
   const pendingCommitSequenceRef = useRef<number | null>(null);
+  const pendingServerMutationSequenceRef = useRef<number | null>(null);
   const invalidatedDuringPendingCommitRef = useRef(false);
   const sequenceRef = useRef(0);
   const transitionAbortRef = useRef<AbortController | null>(null);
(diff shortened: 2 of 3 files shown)


benyhh:
Scheduled PR review session merge record (August 29, 2026): reviewed and fixed the pre-mutation invalidation race found after #1998, froze exact head `481d4992e346176b8dbaf2894be63f21ca86badd` against base `7988c84e359ea9da66ac72122095a14a553296ed`, verified exact-head CI (39 success, 0 failure, 0 pending; Firefox account acceptance passed on bounded rerun), and confirmed Sourcery reviewed the same head with no code findings. Merging this follow-up from the scheduled review session.

benyhh:
Exact-head human review blocks merge on two ownership gaps:

1. After `pending.execute()` and the one invalidation check complete, the server-mutation ref remains active throughout `settleActorTransition()`. An invalidation during that host settlement therefore takes the server-owned fast path, aborts settlement, and returns without any remaining reconciliation owner, leaving the provider neutral/loading.
2. A pre-mutation invalidation supersedes the pending command but leaves `pendingRef` intact with its frozen old generation. The provider returns ready while retaining a stale retry command across the actor change.

I am adding focused regressions and moving the commit-ownership handoff to the point where server response reconciliation is complete.

PR #2024, opened 2026-08-29
OPE-410: Accept exact paginated logout-all actor fences

benyhh:
## What changed

- teach the strict logout-all browser ledger the three finite paginated session reads owned by the current rail
- retain the empty-query shape for rolling-upgrade compatibility
- continue rejecting cursors, searches, duplicate keys, malformed values, and every unrelated 401
- key accepted responses by exact path and query so distinct finite reads remain bounded

## Why

Current `main` changed the session rail from one legacy unpaginated read to exact active-root, archived-root, and pins projections. Firefox correctly observed an old-actor 401 for the archived projection during logout-all, but the acceptance ledger still only allowed the legacy empty query. This caused the combined `main` browser lane to fail even though the API fence was behaving correctly.

## Verification

- focused strict-ledger unit coverage
- all 31 TypeScript project checks
- full Firefox account lifecycle twice
- full WebKit account lifecycle

Linear: OPE-410

## Summary by Sourcery

Accept the session-list requests required by the paginated logout-all flow without weakening strict actor-fence validation.

Bug Fixes:
- Allow the logout-all actor-fence ledger to accept the three exact paginated session reads owned by the current session rail while preserving the legacy empty-query read.

Enhancements:
- Keep actor-fence validation strict by rejecting unsupported queries, duplicate parameters, and unrelated responses while bounding accepted responses by exact path and query.

Tests:
- Add focused coverage for valid paginated session projections and invalid query variants.

### DIFF
--- a/test/e2e/browser-accounts-acceptance.e2e.ts
+++ b/test/e2e/browser-accounts-acceptance.e2e.ts
@@ -1484,7 +1484,8 @@ function logoutAllActorFenceResponseProblem(
 ): string | null {
   const expectedWorkspacePrefix = `/v1/workspaces/${encodeURIComponent(input.workspaceId)}`;
   const isSessionList =
-    response.pathname === `${expectedWorkspacePrefix}/sessions` && response.search === "";
+    response.pathname === `${expectedWorkspacePrefix}/sessions` &&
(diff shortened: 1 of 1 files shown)


benyhh:
@sourcery-ai review

PR #2036, opened 2026-08-29
fix(worker): restore bounded cancellation-wait closure

jorgensandhaug:
## Summary
- restore bounded workflow completion after the exact cancellation-wait signal fence
- rely on the retained durable wake revision and `signalWithStart` to resume receipt reconciliation
- retain the hard worker quiescence watchdog
- restore all three integration assertions that cancellation recovery closes the workflow without manufacturing quiescence or opening the replacement fence

## Why
The low-frequency timer loop keeps a control workflow resident indefinitely while physical quiescence remains pending. That conflicts with the durable workflow-wake handoff: unresolved exact quiescence leaves the wake revision unacknowledged, and receipt settlement advances that revision so `signalWithStart` can restart the stable workflow id without polling inside Temporal.

Two independent exact-SHA CI runs exposed the resulting Temporal integration and cancellation-deadlock fixture timeouts. Later test-only changes rewrote those fixtures around persistent workflow residency, but did not alter the runtime or the documented durable-outbox ownership model.

## Verification
- `bun test ./apps/worker/test/session-state-cancel.test.ts ./apps/worker/test/turn-quiescence-watchdog.test.ts --timeout 120000` (11 pass)
- `bun run --filter @opengeni/worker-bundle typecheck`
- `bunx oxfmt --check apps/worker/src/workflows/session.ts test/integration/temporal-workflow.integration.ts test/integration/steer-cancellation-deadlock.fixture.test.ts`
- `git diff --check`
- the service-backed Temporal and cancellation-deadlock fixtures are delegated to PR CI because Docker is unavailable in the local environment

## Summary by Sourcery

Restore bounded cancellation-wait closure so workflows complete safely while durable wake handling resumes reconciliation after exact quiescence is established.

Bug Fixes:
- Restore bounded workflow completion when cancellation quiescence cannot be confirmed promptly, allowing durable receipt settlement to restart reconciliation without polling.
- Prevent cancellation recovery from manufacturing quiescence or admitting replacement work while the exact cancellation activity remains pending.

Enhancements:
- Retain the hard worker quiescence watchdog while shortening the cancellation wait to a bounded interval.
- Update Temporal integration and cancellation-deadlock assertions to verify completed workflow runs, preserved cancellation fences, and continued activity heartbeats or cleanup progress.

Tests:
- Restore integration coverage for bounded cancellation closure and protection against replacement dispatch during unresolved quiescence.

### DIFF
--- a/apps/worker/src/workflows/session.ts
+++ b/apps/worker/src/workflows/session.ts
@@ -583,20 +583,19 @@ export async function sessionWorkflow(input: SessionWorkflowInput): Promise<void
         if (reconciliation.action === "quiesced") continue;
       }
       // Logical settlement is complete, but the exact predecessor activity has
-      // not yet durably proved sandbox/tool quiescence. Prefer its transactional
-      // queueChanged wake. Also retain one low-frequency deterministic timer:
-      // Pause may acknowledge its control wake before quiescence, leaving no
-      // pending outbox row to restart this workflow after the activity heartbeat
-      // lease expires. The timer guarantees convergence without admission or a
-      // hot control-activity loop.
+      // not yet durably proved sandbox/tool quiescence. Wait only briefly for
+      // its transactional queueChanged wake. If provider/tool cancellation is
+      // genuinely slow, close this workflow run rather than consuming a turn
+      // slot or churning control activities; the outbox uses signalWithStart to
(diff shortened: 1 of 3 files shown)


jorgensandhaug:
@sourcery-ai review

PR #2093, opened 2026-08-31
fix(runtime): contain connector binding rejections

jorgensandhaug:
## Summary
- contain attempt-local connector binding rejection as a model-visible MCP tool error
- skip provider execution and approval-policy hooks for rejected authority
- add regression coverage for the full approval/invocation boundary

## Verification
- `bun test packages/runtime/test/runtime.test.ts` (293 pass)
- `bun run typecheck`
- `bun run lint`
- `bun run format:check`

## Summary by Sourcery

Contain rejected connector bindings at the tool boundary so unauthorized calls return safe tool errors without executing or invoking policy hooks.

Bug Fixes:
- Contain attempt-local connector authority mismatches as model-visible MCP tool errors instead of failing the agent turn.
- Skip connector provider execution and approval-policy hooks when tool arguments are outside the attempt's accepted authority.

Tests:
- Add regression coverage for rejected connector bindings across approval and invocation, including propagation of unexpected binding errors.

### DIFF
--- a/.changeset/calm-connector-authority.md
+++ b/.changeset/calm-connector-authority.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/runtime": patch
+---
+
+Contain pre-execution connector authority mismatches as model-visible tool errors instead of failing the whole agent turn.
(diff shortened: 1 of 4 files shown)


jorgensandhaug:
Addressed the blocking review: only the typed `ConnectorActionBindingRejectedError` is contained. Unexpected binding exceptions still reject both approval and invocation paths. GitHub exact-resource mismatch now throws the typed rejection; regression covers both contained and fail-loud paths.

PR #2114, opened 2026-09-01
fix(config): preserve bare registry turn ownership

jorgensandhaug:
## What
- treat a bare registry model as registry-owned when it is only the per-turn model override and is absent from the deployment built-in allow-list
- retain the hard duplicate failure for actual deployment ownership conflicts
- add regression coverage for the staging DeepSeek/Gateway failure

## Incident
Staging provider acceptance run 33477655460 failed before model I/O because run-scoped `deepseek-v4-flash-0731` was incorrectly materialized as both Azure and `opengeni-gateway`. The deployment default/allow-list remains `gpt-5.6-terra`; this patch makes that distinction generic at the config boundary.

## Validation
- `bun test packages/config/test/model-providers.test.ts packages/runtime/test/model-providers.test.ts` (166 pass)
- `bun run typecheck`
- `bun x oxfmt --check packages/config/src/index.ts packages/config/test/model-providers.test.ts`
- `bun x oxlint --deny-warnings packages/config/src/index.ts packages/config/test/model-providers.test.ts`
- `git diff --check`

## Summary by Sourcery

Correct model ownership resolution for bare registry IDs used as run-scoped overrides while continuing to reject genuine deployment ownership conflicts.

Bug Fixes:
- Preserve registry ownership for bare models used exclusively as per-turn overrides when they are not in the deployment’s built-in allow-list.
- Retain hard failures for bare model IDs that conflict with deployment-owned models.

Tests:
- Add regression coverage for resolving a bare per-turn registry model to its registry provider.

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -4643,12 +4643,15 @@ export function configuredModels(
   // (opaque DeploymentNotFound 404). A `<provider>/<model>`-namespaced id (it
   // contains "/") that a registry actually owns is never a valid Azure/OpenAI
   // deployment name, and a `codex/`-prefixed id never is either — exclude both
-  // from the built-in list. A BARE id a registry merely redeclares (e.g.
-  // "gpt-5.6-sol") is left in place so the built-in still wins it via the first-wins
-  // de-dup below (preserving the documented built-in-precedence contract). When
-  // a codex/ id has NO codex provider injected (no active subscription) it then
-  // resolves to nothing and getModel fails loud with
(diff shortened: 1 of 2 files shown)


jorgensandhaug:
Superseded by #2113. Review correctly identified that this PR could not distinguish deployment configuration from a run-scoped override. #2113 fixes the root catalog invariant instead: database overlays retain the complete built-in membership, allowing the existing run-scoped router to distinguish bare registry turns while preserving hard duplicate detection.

PR #2155, opened 2026-09-02
Show exact model-visible context in the session Debug inspector

jorgensandhaug:
Fixes [OPE-422](https://linear.app/cloudgeni/issue/OPE-422/show-exact-model-visible-context-in-the-session-debug-inspector)

## What

The session Debug inspector now has a Context tab that shows the raw provider ModelRequest from the latest model call:

- The exact `systemInstructions` string passed into `Model.getResponse` / `getStreamedResponse`
- The exact `tools` array on that same request (eager / on-the-wire schemas), plus searchable tools listed separately because they were not sent
- Skills parsed from that sent instruction string (they are not a separate provider field)
- Conservative token estimates for the on-wire prefix, plus last provider-reported input/cached tokens from `agent.model.usage`

This is not a reconstruction of `agent.instructions`. Capture happens at the model client seam, after:

1. OpenGeni persistent instruction composition
2. SDK sandbox wrapping (`# Agent instructions`, capability sections, `# Filesystem`)
3. `callModelInputFilter` (including the missing-title directive)
4. Lazy-tool hiding

The OpenAI Responses client then sends that same string as `instructions` (empty strings are omitted). `agent.model.request` events stay metadata-only.

## How

- Lazy path: `LazyToolModel` snapshots the prepared ModelRequest after searchable tools are stripped
- Non-lazy path: `ModelRequestCaptureProvider` wraps name-resolved models (OpenGeni almost always sets `agent.model` to a string, so wrapping only `agent.model` would miss the request)
- Latest snapshot is upserted per attempt (`session_attempt_model_context_snapshots`, FORCE RLS); retries cannot regress a higher `requestIndex`
- `GET /v1/workspaces/:workspaceId/sessions/:sessionId/model-context` for session readers
- Debug UI fetches on demand when the Context tab is opened and leads with the raw sent string

## Tests

- Snapshot from `ModelRequest.systemInstructions` is kept verbatim
- Name-resolved model provider wrapping captures the request
- Persistent instruction layers join to the exact composed string
- SDK sandbox wrap / genesis-title split
- Preference descriptor parsing
- Runtime posture counts
- SDK contract parity
- Migration 0400 registered in the release-schema contract

## Summary by Sourcery

Expose the exact model-visible context for each session in the Debug inspector.

New Features:
- Add a Context tab to the session Debug inspector showing the latest provider-visible instructions, tools, skills, and token estimates.
- Expose captured model context through a session API endpoint and SDK client method.

Enhancements:
- Capture model requests at both lazy and name-resolved model seams to reflect the exact request sent to providers.
- Persist versioned model-context snapshots per attempt with request ordering, authority checks, and workspace row-level security.

Deployment:
- Add rolling migration 0400 for durable session model-context snapshots and register it in release-schema contracts.

Documentation:
- Document the session Debug model-visible context subsystem in the architecture index.

Tests:
- Add coverage for exact request capture, persistent instruction composition, SDK wrapping, skill parsing, token usage display, contract parity, and database posture changes.

Chores:
- Pin the React package's PostCSS version in publish-consumer dependency overrides for typecheck compatibility.


jorgensandhaug:
Capture now reads the actual provider `ModelRequest` after sandbox wrapping, input filters, and lazy-tool hiding (`systemInstructions` + `tools`), not a reconstructed `agent.instructions` string.

Debug → Context leads with that raw sent payload. The layer list is only a split of that same string (including the SDK `# Agent instructions` / capability / filesystem sections).

PR #2164, opened 2026-09-02
fix(sessions): inherit machine route from sandboxless parent

jorgensandhaug:
## Summary
- let a compatible child inherit an active Connected Machine route from a parent whose managed home backend is `none`
- preserve the shared backend-none home and sandbox group while copying the exact machine and working directory before the first turn
- keep invalid, revoked, unowned, and offline inherited routes behind the existing preflight and atomic create rollback
- document that managed-home classification and the active machine route are independent

## Verification
- `BUN_RUNTIME_TRANSPILER_CACHE_PATH=0 bun test --timeout 180000 apps/api/test/machine-home-backend.test.ts`
- `BUN_RUNTIME_TRANSPILER_CACHE_PATH=0 bun test --timeout 300000 ./test/integration/api.integration.ts -t "manager MCP session_create forwards targetSandboxId"`
- `bun scripts/typecheck.ts --project packages/core --project apps/api`
- `bun run format:check`
- `bun run lint`

## Lifecycle note
The generic failure-after-mutation path was already closed by the atomic machine-target create transaction. The new revoked-inherited-route regression proves this placement path also rejects before a keyed child shell can commit.


jorgensandhaug:
Security review note: this change does not trust a caller-supplied inherited target or bypass route authority. The omitted-placement target is read from the trusted parent session row, `preflightCreateTimeSandboxTarget` revalidates workspace/personal-resource ownership, establishability, liveness, and working-directory resolution under the creating attempt's initiating subject, and `setActiveSandbox` rechecks durable authority inside the session-create transaction before the shell can commit. The new revoked-inherited-route regression proves stale authority returns 422 and leaves no keyed child row. Explicit machine targeting and existing selfhosted-parent inheritance already use the same fence; this patch removes only the contradictory home-backend classification rejection.

PR #2212, opened 2026-09-05
fix(react): show waiting for goals during human turns

VasuBansal7576:
A running user-requested turn showed a `Blocked` goal pill beside the `Running` session header.
Map the backend's `blocked / human_turn_running` projection to the existing `Waiting` presentation, since autonomous goal work waits for the foreground turn to finish.

## Verification

- The regression fails on the original code and passes with the fix.
- React suite: 25 tests passed, 114 assertions.
- React package typecheck and formatting checks passed.

## Before and after

Captured in the local test instance during active human turns.

| Before | After |
| --- | --- |
| Running session with a Blocked goal | Running session with a Waiting goal |

### DIFF
--- a/packages/react/src/components/session-chrome.tsx
+++ b/packages/react/src/components/session-chrome.tsx
@@ -228,6 +228,7 @@ export function sessionChromeGoalPillState(
   // next evaluation at `nextAttemptAt`) is an ordinary scheduled state.
   if (continuation.state === "scheduled") return "scheduled";
   if (continuation.state === "blocked") {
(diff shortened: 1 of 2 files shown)


benyhh (commented):
Independent Codex technical review: no blocking findings. Read the complete small mapping/test change. An active human turn is an ordinary wait for goal continuation, so Waiting is accurate and does not hide session failure state. Clean current-main compatibility and all32 session-chrome tests/135 assertions passed. Approved the previously gated first-contributor CI run after inspecting its diff and trusted workflow. Merge remains contingent on candidate CI. This is an AI technical review.

PR #2228, opened 2026-09-05
fix(api): restore workspace archives from object storage

VasuBansal7576:
Attaching a viewer or using Channel A against a cold sandbox could fail with `workspace_fingerprint_mismatch` even when its durable workspace archive existed in object storage. The API passed the archive reference into provider restoration without loading its bytes, while the worker path already materialized those references.

The API now uses the existing archive verifier before restoring the selected revision, and passes object storage through both callers. The hydrated bytes stay transient; the persisted replacement envelope retains the object reference. Missing or corrupt objects leave the lease cold with the recorded restore failure.

## Verification

- The same desired-success test fails on upstream `b4575cc` and passes on `bd51aa5`.
- Four regression cases pass with 38 assertions: restored file contents and revision, ref-only persistence, viewer attach, and missing/corrupt archive failure state. The viewer fixture closes its restored provider after detaching.
- 38 existing archive/contract tests pass with 123 assertions. Lint and formatting pass.
- 149 integration tests passed across API, uploads, database, NATS, and control transport before the bounded run stopped during NATS auth-callout failures.

Full local checks are not green. `bun run check` fails on dependency/type issues in the reused worktree setup. An independent API typecheck using one coherent dependency graph produces the same two `fatal-process-boundary.ts` diagnostics on baseline and patch. The integration run reached its 120-second bound after `AUTHORIZATION_VIOLATION`; browser e2e passed the first Chromium case, then stopped because the Firefox executable is absent.

## Before and after

These captures show the same real Docker/PostgreSQL test with an in-memory ObjectStorage stub. They show test output, not browser UI or live S3 coverage.

| Before | After |
| --- | --- |
| Baseline archive restoration fails | Patched archive restoration succeeds |

## Summary by Sourcery

Restore cold API sandboxes from durable workspace archive references while validating archive integrity and preserving failure state.

Bug Fixes:
- Restore API-direct and viewer-attached cold sandboxes from valid workspace archives stored in object storage.
- Record unrecoverable cold lease state when archive objects are missing, corrupt, or malformed instead of creating a replacement workspace.

Enhancements:
- Preserve durable object-storage archive references after temporarily materializing archive bytes for restoration.

Tests:
- Add regression coverage for API spawner restoration, viewer attachment, durable reference persistence, and archive integrity failures.

### DIFF
--- a/.changeset/archive-ref-recovery.md
+++ b/.changeset/archive-ref-recovery.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/api-router": patch
+---
+
+Restore API-direct sandboxes from workspace archives stored behind durable object-storage references.
--- a/.changeset/archive-ref-recovery.md
+++ b/.changeset/archive-ref-recovery.md
@@ -2,4 +2,4 @@
 "@opengeni/api-router": patch
 ---
 
-Restore API-direct sandboxes from workspace archives stored behind durable object-storage references.
+Restore API-direct sandboxes from workspace archives stored behind durable object-storage references. Reject malformed archive locators before creating a replacement workspace.
--- a/apps/api/src/routes/sessions.ts
+++ b/apps/api/src/routes/sessions.ts
@@ -335,6 +335,7 @@ export function registerSessionRoutes(app: Hono, deps: SessionRouteDeps): void {
     db,
     settings,
     bus,
+    objectStorage,
     observability: deps.observability,
   };
   const workspaceCaptureManifestCache = new WorkspaceCaptureManifestCache();
@@ -516,6 +517,7 @@ export function registerSessionRoutes(app: Hono, deps: SessionRouteDeps): void {
     db,
     settings,
     bus,
+    objectStorage,
     ...(deps.establishSandboxSession
       ? { establishSandboxSession: deps.establishSandboxSession }
       : {}),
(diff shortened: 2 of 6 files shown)


VasuBansal7576:
@sourcery-ai Agreed that this needs maintainer review. I checked the reference's path into this change:

- The API obtains its envelope from the acquired lease or `getSandboxSessionEnvelope(db, workspaceId, session.id)`.
- `beginSandboxRematerialization` applies account/workspace RLS, locks the workspace/group lease, and checks the lease epoch before selecting the durable archive.
- The archive producer constructs the object key from account, workspace, sandbox group, and revision. This patch reuses the selected reference and existing size/hash/archive verification, then retains reference-only persistence.

The key parser validates the key's shape; it does not independently establish workspace ownership. I am not treating that parser or a matching hash as a tenant-isolation proof. The new regressions cover restoration, viewer attach, reference-only persistence, and missing/corrupt object failures, but not forged cross-workspace references.

The remaining human-review focus is the provenance of the selected durable envelope, including existing fork/archive-transfer rules. This PR adds no public input for selecting an arbitrary archive key. The human-review request remains open.

benyhh (commented):
Independent review: reviewed the complete archive restore change and the malformed-reference follow-up. Object-backed restoration validates retained bytes and integrity; missing, corrupt, and malformed retained references fail explicitly instead of silently returning an empty workspace. Five real PostgreSQL/local sandbox cases pass with 44 assertions, including red/green evidence against the original implementation. No blocking finding. Merge remains gated on successful exact-head CI.

PR #2238, opened 2026-09-07
fix(react): restore goal timeline landmarks in production bundles

benyhh:
Goal continuation landmarks can render as “Timeline item unavailable” in the production web app. A cycle among shared chunks initializes the timeline's module-level metadata before its imported icon component, permanently capturing `undefined` and causing React error 130.

Resolve the icon binding when the goal row renders, after module initialization. Event payloads and history remain unchanged. This restores the existing goal landmarks without hiding rendering failures.

Validation: the added browser test uses the full production app, real API, and non-superuser PostgreSQL. It fails on current main because the continuation never appears, then passes with this fix, verifying visible text and glyph, no unavailable cards, and no React errors. The original staging orchestrator independently reproduces the error; an isolated component build does not exercise the shared chunk cycle.

Tracks [OPE-437](https://linear.app/cloudgeni/issue/OPE-437/fix-unavailable-goal-timeline-rows-in-the-production-bundle).

## Summary by Sourcery

Restore reliable rendering of goal timeline landmarks in production bundles.

Bug Fixes:
- Restore goal timeline landmarks in production bundles by resolving their icons after module initialization, preventing unavailable continuation rows and React rendering errors.

Tests:
- Add a production-app browser test covering goal landmark text, icon rendering, unavailable-card absence, and React error reporting.

### DIFF
--- a/.changeset/quiet-goal-landmarks.md
+++ b/.changeset/quiet-goal-landmarks.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/react": patch
+---
+
+Resolve goal landmark icons when rendered so circular production chunks cannot leave continuation rows permanently unavailable.
--- a/packages/react/src/components/message-timeline.tsx
+++ b/packages/react/src/components/message-timeline.tsx
@@ -3400,41 +3400,46 @@ type GoalMeta = { label: string; pill: string; icon: ComponentType<{ className?:
 
 const NEUTRAL_PILL = "border-og-border bg-og-surface-1 text-og-fg-muted";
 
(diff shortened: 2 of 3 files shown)


benyhh (commented):
Independent Codex technical review: no blocking findings. Read the full render-time icon-binding change and its production-bundle regression. It fixes the reproduced circular chunk initialization failure without changing durable events. The actual production/API/PostgreSQL browser test fails before the fix and passes afterward. Clean current-main compatibility tree16631e2fc3d302e98eee662e5b0119a737879a30. Merge remains contingent on candidate CI. This is an AI technical review.

PR #2240, opened 2026-09-07
fix(web): explain pre-claim failures without reusing old provider errors

benyhh:
A session can fail before a turn is claimed and record only a `session.status.changed` failure. The web banner previously ignored that boundary, so it showed no reason or reused an older provider rejection as if it had happened again.

Recognize the pre-claim failure as the current failure. When no detail was recorded, explain that the session failed before a turn could start. Preserve an exact same-turn paired diagnostic when available; do not invent a provider cause or count a non-turn failure as another failed turn. Stored events remain unchanged.

Validation: a regression reproduces the old provider/safety rejection leaking into the new failure; all six failure-summary tests pass after the correction. Full repository typecheck passes.

Tracks [OPE-439](https://linear.app/cloudgeni/issue/OPE-439/show-the-current-pre-claim-failure-instead-of-an-older-provider).

## Summary by Sourcery

Show the current pre-claim failure accurately without attributing stale provider errors or inflating failed-turn counts.

Bug Fixes:
- Ensure pre-claim session failures are shown as the current failure instead of reusing an older provider rejection.
- Provide a clear fallback explanation when a pre-claim failure has no recorded details.
- Preserve detailed diagnostics only when they are paired with the same failed turn.

Enhancements:
- Update failure summaries to treat the latest recorded failure boundary as authoritative without counting pre-claim failures as additional failed turns.

Tests:
- Add regression coverage for stale provider-error reuse and same-turn pre-claim diagnostics.

### DIFF
--- a/apps/web/src/App.test.ts
+++ b/apps/web/src/App.test.ts
@@ -1582,6 +1582,55 @@ describe("summarizeSessionFailure", () => {
     );
   });
 
+  test("reports an unclaimed failure instead of reusing a previous provider rejection", () => {
+    const failure = {
+      ...event(3, "session.status.changed", {
+        status: "failed",
+        code: "pre_claim_failure",
+        failedSystemUpdateIds: ["update-1"],
+      }),
+      turnId: null,
+    };
+    const summary = summarizeSessionFailure(
+      [
+        {
+          ...event(1, "turn.failed", { code: "provider_safety_refusal", error: "Old rejection" }),
+          turnId: "previous-turn",
+        },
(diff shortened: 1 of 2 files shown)


benyhh (commented):
Independent Codex technical review: no remaining source blocker found. Read complete failure-boundary projection and both new regression cases. A newer unclaimed failure no longer reuses a previous provider refusal; a same-turn detailed failure retains its diagnostic. Six summary tests and full typecheck passed. Latest-main clean integration tree10f40c47f1a6ddcd9a13000d07be8c3242f5c930. This is an AI technical review.

PR #2241, opened 2026-09-07
fix(react): distinguish historical wait outcomes from current activity

benyhh:
Preserved wait outcomes now say “Wait recorded” with the event's date and time. This distinguishes an agent's historical statement about running children from current session activity, while retaining the exact reason text.

These specific turn-end outcomes use the static note role and semantic time values, rather than a live status announcement. Other notices retain their existing behavior; stored events remain unchanged.

Validation: 252 timeline projection/renderer tests pass and full repository typecheck passes. Browser checks of the retained original orchestrator timeline verify all four historical labels and timestamps, static note roles, and no horizontal overflow at desktop and 375px mobile widths. The styled mobile rendering was inspected.

Tracks [OPE-440](https://linear.app/cloudgeni/issue/OPE-440/label-preserved-wait-outcomes-as-historical-with-their-event-time).

## Summary by Sourcery

Present preserved wait outcomes as timestamped historical notes instead of current activity status messages.

New Features:
- Display preserved wait outcomes with a recorded date and time while retaining the original reason text.

Bug Fixes:
- Distinguish historical wait outcomes from live session status announcements.

Enhancements:
- Expose preserved wait outcomes as static notes with semantic time information while leaving other notices unchanged.

Tests:
- Update timeline renderer coverage for historical labels, timestamps, and note roles.

### DIFF
--- a/.changeset/tidy-recorded-waits.md
+++ b/.changeset/tidy-recorded-waits.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/react": patch
+---
+
+Label preserved wait outcomes with their recorded date and time and expose them as static notes, keeping historical agent text distinct from current session state.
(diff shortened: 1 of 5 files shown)


benyhh (commented):
Independent review: reviewed the complete five-file change. Preserved wait outcomes now carry a recorded timestamp and static note semantics; original reason text is preserved. No blocking finding. Focused renderer tests and styled mobile inspection are recorded by the implementation worker. Merge remains gated on successful exact-head CI.

PR #2242, opened 2026-09-07
fix(db): re-arm paused attempt quiescence reconciliation

benyhh:
Repeating Pause on an already-paused session previously returned no wake even when its logically settled interruption still lacked a physical-quiescence receipt. Three preserved review sessions remained stuck in the stopping state after their provider had gone away.

A fresh Pause operation now re-arms the existing durable workflow wake for that selected session’s closed, interrupted attempt. Control remains paused at the same revision. The worker still proves exact Temporal activity settlement and writer quiescence; this creates no new turn or model/tool work. Replaying the same operation key adds no wake revision.

Validation: desired-success regression fails before the change (wakeCount 0), full PostgreSQL control-plane suite passes 88 tests / 669 assertions, and real Temporal signalWithStart integration proves exact-attempt reconciliation with zero turn executions. Database typecheck, focused lint, docs references, and diff checks pass. Clean integration against main e9d092a30 produced tree d63f6a0b31b4c8e35db455f6d3f0f630ceebf508.

[OPE-441](https://linear.app/cloudgeni/issue/OPE-441/re-arm-reconciliation-when-pause-is-repeated-on-an-already-paused) is a focused follow-up to OPE-169 and OPE-18. Historical ownership remains unchanged. Existing live sessions remain paused; production recovery requires deployment before using this behavior.

## Summary by Sourcery

Restore reconciliation progress for paused sessions with unsettled physical quiescence while preserving control state and avoiding new work.

Bug Fixes:
- Re-arm the missing exact-attempt quiescence reconciliation wake when Pause is repeated on an already-paused session.

Enhancements:
- Preserve paused control state and revision while allowing the existing worker to reconcile settled interruptions without starting new turn work.

Documentation:
- Document repeated Pause as a recovery mechanism for missing quiescence wakes and clarify that it preserves paused admission and idempotent replay behavior.

Tests:
- Add control-plane coverage for wake re-arming, unchanged control state, idempotent replay, and prevention of new session work.
- Add Temporal integration coverage proving the exact interrupted attempt is reconciled without executing a turn.

### DIFF
--- a/.changeset/paused-quiescence-nudge.md
+++ b/.changeset/paused-quiescence-nudge.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/db": patch
+---
+
+Repeat Pause on an already-paused session to re-arm its missing exact-attempt quiescence reconciliation without resuming work.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -175,8 +175,8 @@ A recoverable activity shutdown creates a transactional workflow-wake
 obligation in Postgres. Delivery remains unacknowledged until the exact closed
 attempt has durable quiescence, and every attempt-owned retained-process
 settlement advances that same outbox row in its settlement transaction. A
-workflow close or a writer exit racing the final reconciliation check therefore
-cannot orphan a session in recovery.
+workflow close or writer exit racing reconciliation cannot orphan recovery;
+repeated Pause re-arms a missing quiescence wake.
 
 A background command becomes session-owned only after its exact provider
 identity is durably adopted. Before adoption it remains attempt-owned. After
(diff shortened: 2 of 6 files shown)


benyhh (commented):
Independent review: examined the exact-session closed-attempt/settled-interruption predicate, unchanged-control receipt, existing wake registration, operation-key replay, and PostgreSQL/Temporal tests. Repeated Pause can re-arm missing quiescence reconciliation while admission remains paused and the control revision remains unchanged. Completion still requires the existing worker proof; no model turn or automatic full-turn retry is introduced. No blocking finding. Exact-head CI required before merge.

PR #2252, opened 2026-09-07
fix(runtime): preserve portable compaction protocol validity

davletd:
Portable Responses compaction can reject complete inline history after preparation removes opaque reasoning: retained provider response-item IDs still require their original reasoning item. This patch detaches optional response identity only in the checkpoint request, preserving structured tool history and call/result correlation. Azure-profile checkpoints also disable tool selection so historical tools cannot replace summary text.

The diagnosis is supported by a minimal live reproduction and differential replay of the preserved 709-item incident history: the unmodified published runtime returns HTTP 400 / missing required reasoning, and the corrected request succeeds. The original incident response's detailed message was not retained. Replay establishes this failure mechanism; it cannot conclusively identify the original request's rejection. Context/output budgets and malformed tool pairs were investigated as alternatives, not assumed causes. No customer history was written or session replaced.

Implementation:

- Reuse the ordinary inference ID-strip primitive for optional message, plaintext reasoning, function, shell, computer, patch, client tool-search and hosted web-search IDs. Preserve payload IDs, call correlation, ordering, required hosted IDs and approval/program links.
- Use the SDK's tool-search execution/correlation readers for top-level, camelCase and providerData shapes, including absent execution with explicit correlation. Explicit server execution retains its identity.
- Scope forced tool selection to `azure-openai`. Other Responses transports retain their prior tool-choice shape. Ordinary inference and remote-v2 policies are unchanged.
- Persist only the bounded `missing_required_reasoning_item` classification, not arbitrary provider messages or referenced IDs. Incomplete, empty and failed provider responses still fail closed.
- Supply a CI-typechecked operator canary using the runtime's Azure client, with API-version/deployment/AD-token support, an opaque-reasoning kickoff, disposable PostgreSQL archival verification and a continued paired tool exchange.
- Keep disposable database ownership at the caller. Ordinary tests omit the database case when unavailable; `OPENGENI_REQUIRE_REAL_DB=1` and explicit live durable verification require it. Release still runs after setup, verification or client-close failure.

Architecture assessment:

The fix belongs at the provider request boundary. Canonical history is conversation truth, a prepared checkpoint request is a temporary transport projection, and the replacement is a separate attempt-fenced database transaction. This patch preserves those responsibilities. It adds no history writer, replacement session, retry loop, synthetic turn or second checkpoint authority. Current-main history/attachment preservation from #2260 and subsequent replay changes integrate without changing that conclusion.

Universal ID stripping is unsafe: live Azure accepts ID-less web-search input but rejects ID-less file search. Flattening every Responses tool record into text would also discard protocol structure unnecessarily. The selective projection retains required identities and lets unknown provider failures stop honestly. Future hosted protocols need explicit SDK/wire evidence rather than a permissive fallback.

A completed, smaller summary is still lossy. Existing shrink/repeat, failed-response and exact-attempt guards prevent invalid replacement mechanics, not every omission of meaning. Original rows remain archived. The strengthened canary generates an unpredictable receipt found only in an old patch result; neither retained user messages nor instructions reveal it. A continuation using only the reloaded checkpoint must recover that exact receipt. Its next request preserves the complete prefix, rejects multiple tool calls, supplies the one correlated result and verifies consumption with further tool selection disabled. This materially strengthens semantic evidence without claiming exhaustive equivalence.

Review dispositions:

- Prior shell/computer and [canonical tool-search](https://github.com/Cloudgeni-ai/opengeni/pull/2252#discussion_r3957239099) identity gaps are fixed and covered through the SDK converter.
- [Hosted/plaintext IDs](https://github.com/Cloudgeni-ai/opengeni/pull/2252#discussion_r3957239116) were verified live. The proposed universal strip was rejected based on the required file-search ID counterexample; its existing primitive was reused selectively.
- [Non-Azure tool-choice compatibility](https://github.com/Cloudgeni-ai/opengeni/pull/2252#discussion_r3957239126) is handled by the Azure-only policy and ordinary/streaming transport regressions.
- Operator error visibility, raw Azure transport, missing CI coverage, hermetic settings and tool-only rejection feedback are addressed.
- The new [optional PostgreSQL finding](https://github.com/Cloudgeni-ai/opengeni/pull/2252#discussion_r3958337198) was reproduced with an unavailable Docker endpoint, then fixed as above. A fresh independent architectural/code review found no remaining blocking runtime issue and identified the canary improvements now implemented.

Validation on latest-main integration (`596ace9eb`):

- 141 focused tests passed, 657 assertions: runtime compaction/identity (93), operator plus fenced worker activity (28), SDK/PostgreSQL history durability (7), and provider/history-switch compatibility (13).
- Without Docker, all six operator tests complete and the database case is omitted. With `OPENGENI_REQUIRE_REAL_DB=1`, the same unavailable endpoint correctly fails. Real PostgreSQL exercises the archival assertions.
- Runtime source, runtime regression and operator projects pass tsgo. Targeted Oxlint, formatting, diff hygiene and migration-test budget guard pass; frozen Bun install succeeds.
- Fresh live Azure `gpt-5.6-sol` durable canary: 208 items, estimated history 350,837 → 440 tokens, every original row archived unchanged, committed checkpoint event and token/request clearing verified. The continuation recovered the random old-tool-only receipt and consumed its new paired result. Only disposable synthetic data was used.
- Merged current main and resolved the two shared bundle-budget conflicts using main's maintained definitions. Those files and the earlier invitation fixture fixes no longer appear in this PR's delta against main. [Full CI run 34395792483](https://github.com/Cloudgeni-ai/opengeni/actions/runs/34395792483) passed on the first attempt for `7f66b786fb109b2b58ca5b4c405d342bb659dd62`: all unit, real-service, browser/E2E, source/type, recovery, native/WASM, deployment, package/bundle and aggregate checks are green. GitHub reports this head mergeable; no later review findings were present at final verification.

Reproduce with `bun test --timeout 30000 packages/runtime/test/portable-compaction-identity.test.ts scripts/operator/verify-portable-compaction.test.ts`. For live verification, configure Azure and run `bun run verify:portable-compaction --live --durable`; omit `--durable` for provider-only checks. Receipts expose counts/statuses and synthetic IDs, not credentials or conversation content.

This PR carries an immutable package patch changeset, not a published release or Cloudgeni deployment. Cloudgeni must adopt a coherent checksummed release/package BOM before staging can benefit. Its existing FAILED/null-session dashboard presentation remains a separate product limitation. No production/staging deployment, customer session replacement, or token-accounting change is included.


packages/runtime/src/index.ts
```diff
@@ -1043,6 +1045,54 @@ export async function summarizeForCompaction(
...
+]);
+
+function detachCompactionResponseItemIdentity(
+  item: Record<string, unknown>,
+): Record<string, unknown> {
+  const providerData = item.providerData;
+  const provider = providerData as Record<string, unknown> | undefined;
+  const clientToolSearch =
+    (item.type === "tool_search_call" || item.type === "tool_search_output") &&
+    provider?.execution === "client" &&
```
benyhh: **[P1] Client tool_search detach misses the top-level `call_id`/`execution` shapes** (the unresolved Lens P1 #2 thread)

This predicate reads only `providerData`. Canonical history also stores the SDK wire shape with top-level `call_id` (or `callId`) plus `execution: "client"`, and rows with `providerData.call_id` but no `execution` at all. Those shapes are pinned by `packages/runtime/test/history-sanitizer.test.ts:53-70` and `apps/worker/test/codex-history-strip.test.ts:22-36`, and written by `packages/db/src/session-tool-call-settlement.ts:82-104`.

Probe through `summarizeForCompaction` with a fake client: `{type:"tool_search_call", id:"tsc_x", call_id:"s2", execution:"client"}`, the `callId` variant, `{providerData:{call_id}}` without execution, and the matching `tool_search_output` all keep `id` on the wire. Only the `providerData`+`execution` shape is detached. After preparation strips the reasoning id, those rows still produce the same 400.

`@openai/agents-core` exports `getToolSearchExecution` and `getToolSearchProviderCallId`, which the converter itself uses and which handle every shape; `toolCallIdFromSdkItem` in `packages/runtime/src/tool-call-identity.ts` covers the id half. Either beats a fourth hand-rolled shape reader.

packages/runtime/src/index.ts
```diff
@@ -1043,6 +1045,54 @@ export async function summarizeForCompaction(
...
 
+/**
+ * Portable checkpoints carry inline history, not references to stored Responses
+ * items. An assistant/tool item's provider id can require its original reasoning
+ * item even when the full message is supplied. The portable preparation removes
+ * opaque reasoning, so retaining those dependent ids makes Azure reject the
+ * checkpoint. Remove only provider item identity from this request-local copy;
+ * callId/call_id, tool payloads, order, and canonical history stay unchanged.
+ */
+const DETACHABLE_COMPACTION_ITEM_TYPES = new Set([
```
benyhh, line 1066: **Allowlist still sends `hosted_tool_call` and plaintext `reasoning` ids**

`hosted_tool_call` is not in this set, so a default-on hosted `web_search_call` row (`ws_...`) that follows opaque reasoning keeps its id after the reasoning id was stripped. Probe wire: `[reasoning without id, web_search_call id:"ws_x", message without id]`, which is the exact dependency shape this PR fixes for messages.

Plaintext reasoning (`OPENGENI_OPENAI_REASONING_ENCRYPTED_CONTENT=false`) likewise keeps `rs_...` because `projectRejectedReasoningArtifact` only runs for opaque items, and `summarizeForCompaction` never consults `openaiProviderItemIds`, so `preserve` and `strip` produce the same checkpoint wire.

Ordinary turns already send every one of these items id-less on every Azure call (`stripProviderItemIdsFilter`, `packages/runtime/src/model-input.ts:31-56`, default `strip`), so nothing shows the retained ids are required; the test at lines 262-343 pins the retention without evidence. The hosted case is not live-verified; a canary run whose fixture includes an encrypted reasoning + `web_search_call` pair would settle it.

packages/runtime/src/index.ts
```diff
@@ -1007,9 +1007,11 @@ export async function summarizeForCompaction(
   // items without flattening tool history into a fake user transcript.
   const request: ModelRequest = {
     systemInstructions: options.systemInstructions ?? "",
-    input: input as AgentInputItem[],
+    input: input.map(detachCompactionResponseItemIdentity) as AgentInputItem[],
```
benyhh, line 1019: **Consider reusing the existing universal strip instead of a second allowlist**

`stripProviderItemId` (`packages/runtime/src/model-input.ts:52-56`) already strips the top-level `id` from every item before every ordinary model call for the same Azure stored-response reason. Exporting it and applying it in `omitOpaqueArtifactsFromPortableCompactionHistory` (where the reasoning id is already removed) closes the tool_search and hosted gaps above without enumerating item types, and keeps one id policy in the package.

A probe of a universal strip on this PR's "preserves hosted required ids" fixture keeps `mcp_approval_request.id` and `mcp_approval_response.approval_request_id` (the SDK emits `providerData.id ?? item.id` for those). The only difference is `program`/`unknown` items losing their top-level id, which no producer in the repo emits today and whose caller link is `call_id`, not `id`.

Two dead pieces in the current version: the `providerData.id` deletion below never changes the wire (the SDK converter reserves `id` out of providerData for every type in this set; shell/apply_patch ignore providerData entirely), and `function_call_output`/`computer_call_output` are unreachable here (the converter throws `Unsupported item` for them; canonical history uses the `_result` names).

packages/runtime/src/index.ts
```diff
@@ -1007,9 +1007,11 @@ export async function summarizeForCompaction(
...
     modelSettings: {
       maxTokens,
+      // Historical tool records remain input, but a checkpoint must be text.
+      toolChoice: "none",
```
benyhh: **`tool_choice: "none"` now reaches the Codex, xAI, and gateway transports unverified**

This is the only `toolChoice` producer in the tree, so the SDK previously omitted the field on every request. `normalizeCodexRequestBody` allowlists and forwards it (probe: `{tools:[],tool_choice:"none",store:false,stream:true}`), the xAI normalizer passes it through, and the gateway request policy never touches it. Repo evidence of Codex backend acceptance covers only `"auto"` (`packages/codex/test/normalize.test.ts:208-214` and the live spikes).

Portable-locked Codex sessions, SuperGrok sessions, and gateway `api:"responses"` models therefore send a field their backend has never been observed to accept with empty tools; a rejection turns every checkpoint on that provider into `CompactionProviderResponseError`.

The case this guards is already fail-closed: a `function_call`-only reply yields `""` from `extractResponseOutputText` and throws `EmptyCompactionSummaryError`, which the doc addition at `docs/context-compaction.md:345` acknowledges. Suggest dropping the field, or gating it to the built-in OpenAI/Azure wire profile that was live-tested.

scripts/verify-portable-compaction.ts
```diff
@@ -0,0 +1,222 @@
...
+  console.log(JSON.stringify(manifest));
+}
+
+if (import.meta.main) {
+  try {
+    await main();
+  } catch (error) {
+    // Never let the CLI print SDK causes containing raw provider messages.
+    console.error(
+      JSON.stringify({ verificationFailed: true, ...compactionProviderFailureDiagnostics(error) }),
```
benyhh: **The catch-all hides the script's own error messages**

`compactionProviderFailureDiagnostics` never emits `message`, so every non-provider failure prints the same blob. Running without `--live` exits 1 with exactly `{"verificationFailed":true,"errorName":"Error","httpStatus":null,"responseStatus":null,"responseId":null,"code":null,"type":null,"requestId":null}`. The same output appears for missing Azure settings, the shrink assertion, the facts mismatch, the receipt check, and every durable-fixture error (which also drops `outcome.reason` at `scripts/fixtures/durable-compaction.ts:100`). Only genuine provider errors need the content-free projection; print the message for the script's own assertions.

scripts/verify-portable-compaction.ts
```diff
@@ -0,0 +1,222 @@
...
+  ) {
+    throw new Error(
+      "Set OPENGENI_OPENAI_PROVIDER=azure and the explicit Azure base URL, API key and model",
+    );
+  }
+  const base = settings.azureOpenaiBaseUrl.replace(/\/+$/, "");
+  const apiKey = settings.azureOpenaiApiKey;
+  const receipts: Item[] = [];
+  async function request(body: Item): Promise<ProviderResponse> {
+    const response = await fetch(`${base}/responses`, {
```
benyhh: **Raw `fetch` bypasses the repo's Azure client construction**

The summarizer under test uses `buildOpenAIClientFromSettings` (`packages/runtime/src/model-provider-client.ts:74-88`): `azureOpenaiBaseUrl ?? azureDeploymentBaseUrl(settings)`, api-key or AD-token bearer, and the `api-version` default query unless the base URL ends in `/openai/v1`. The kickoff and continuation legs here send only `api-key` with no query string and refuse endpoint+deployment or AD-token configurations (`.env.example:416-424` documents both as valid), so the canary can fail on legs that are not the code under test while the real summarizer works. `buildOpenAIClientFromSettings(settings).responses.create(...)` replaces `request()` and the precondition.

scripts/operator/compaction-durable.ts
```diff
@@ -0,0 +1,142 @@
...
+  isSessionCompactionRequested,
+  getActiveSessionHistoryItems,
+  getSession,
+  requestSessionCompaction,
+  withWorkspaceRls,
+  withWorkspaceSessionActivityRls,
+} from "@opengeni/db";
+import * as schema from "@opengeni/db/schema";
+import { acquireSharedTestDatabase } from "@opengeni/testing";
+import { maybeCompactContext } from "../../apps/worker/src/activities/context-compaction";
```
benyhh, line 17: **This canary sits outside every CI gate**

`typecheckProjects()` (`scripts/ci/workspace.ts:175-183`) covers only `scripts/ci`, `scripts/operator`, `scripts/release`, and package dirs; there is no `scripts/tsconfig.json`; `.oxfmtrc.json` ignores `**/fixtures/**` (a rationale written for byte-exact fixtures); `bun test` never imports this file, which is reached only through the dynamic import behind `--durable`; and `package.json` has no entry for the CLI. The next change to `maybeCompactContext`'s positional signature or `ClaimSessionWorkForAttemptInput` leaves typecheck, tests, and format green while the documented operator canary fails at runtime.

`apps/worker/test/context-compaction-activity.test.ts:58-76` already has the claim helper and runs `maybeCompactContext` against disposable Postgres with an injected summarizer, so the archive/event/token-cleared proof fits there; the CLI fits under the typechecked `scripts/operator` project like `turn-density-profile.ts`.

Minor: the `finally` awaits `client.close()` before `shared.release()`, so a rejected close leaks the cloned database and the shared-container holder token.

benyhh (commented):
Independent review of head c392905: local worktree, wire probes through `summarizeForCompaction` with a fake client, unit test run. The id-detach mechanism is real, and unit shards, type contracts, and real-service shards are green on this head. The red CI jobs are Playwright install failures and the pre-existing web bundle budget, not this PR.

Inline below: the confirmed client tool_search shape gap (the Lens P1 #2 thread is still open), hosted-tool and plaintext-reasoning ids still on the checkpoint wire, a suggestion to reuse the existing universal id strip, the unverified `tool_choice: "none"` on non-Azure transports, and three problems in the verification scripts.

Smaller items not inlined:
- `scripts/verify-portable-compaction.ts:74` never asserts the kickoff reasoning item is opaque, so on a deployment that returns plaintext reasoning the canary never sends the shape this PR fixes and passes vacuously.
- `packages/runtime/test/portable-compaction-identity.test.ts:17` builds settings from `getSettings()` at import time; sibling compaction tests use the hermetic `testSettings()` from `@opengeni/testing`.
- The test at line 134 ("forbids historical tool execution") only asserts the summary string; test 1 already pins `tool_choice: "none"`, and the named behaviour (a tool-call-only reply rejects with `EmptyCompactionSummaryError`) is not asserted.
- AGENTS.md line 201 and `docs/run-lifecycle.md:665` enumerate what happens to the portable temporary copy and now omit the id detach and `tool_choice`.

Review generated with Claude Code.

PR #2310, opened 2026-09-08
feat: one-handler chat integration, per-session agent access and memory scope, org read keys

benyhh:
Products can put OpenGeni behind an existing chat with one server handler, while choosing agent access and memory scope for each conversation inside a customer workspace.

- Add `@opengeni/sdk/chat`, a native handler, Vercel and OpenAI protocol adapters, and `@opengeni/react/chat`. Conversation identities are bound to the host-resolved user. Reload restores pending approvals and questions, including multi-select input. Authentication changes clear private React state and abort old requests; cookie-based hosts supply an explicit `authKey`. Responses streams have complete lifecycle events and stable IDs; streamed paragraphs match the final reply.
- Freeze per-session `agentAccess`, the opaque `endUser` label, and `memoryScope`. Enforce agent reach and host allowlists in the shared authorization seam. Private memory identities hash the exact label tuple, and corrections, archival, and replacement remain within the writable layer.
- Prevent tool/access widening through child creation, tool-policy changes, scheduled sessions, and the Codemode proxy. Missing signed tool-selection claims fail closed. The Site proxy retains an explicit route allowlist alongside Site-origin filtering.
- Add read-only organization API keys and paginated organization-wide session listing across shared workspaces. Agent isolation does not change organization-key or human authorization; Personal workspaces and private sessions remain excluded from organization-wide listing.

Validation: 236 focused tests pass across SDK/React, real PostgreSQL HTTP/MCP authorization, scheduling, migrations, schema history, Site routing, previously failing CI fixtures, bundle-budget policy, six authentication-switch regressions, and the real editable-artifact browser reload/collaboration flow. All workspace typechecks, lint, formatting, migration guards, docs-reference checks, and public hygiene pass locally. Publishable packages build with both chat subpath targets and pass the closure guard. Production web assets build; the direct-session budget was adjusted only for the measured 4,207 raw / 1,377 gzip byte delta against main `8e0bf0d28`, using the existing headroom policy. Full [GitHub CI](https://github.com/Cloudgeni-ai/opengeni/actions/runs/34263374851) and [desktop CI](https://github.com/Cloudgeni-ai/opengeni/actions/runs/34263374710) passed on `fc6f4cd7f` (37 passing checks). The session-pin browser lane passed on one same-commit rerun after all 23 tests had passed but exceeded the step wall limit in the first attempt; no assertions or timeouts were changed.

Deployment: migrations `0427_session_agent_access_scope.sql` and `0428_scheduled_task_creator_policy.sql` are rolling. Main's `0425_feedback_submissions.sql`, `0426_codex_bookkeeping_preserves_session_recency.sql`, and published migration history are preserved. Conflict resolution retains both Site-origin routing and the stricter path boundary, and both scheduled creation-policy recovery and frozen creator authority.


benyhh:
Thanks for the review. Addressed on this branch (merged with current main, migrations renumbered to 0425/0426):

- **Conversation authorization**: fixed. Conversation ids are now namespaced per resolved user in the session id and idempotency derivation, `chatBySessionId` requires the resolved user and returns 403 for another user's session, a host-named conversation that disagrees with `previous_response_id` is a 409, and a host that returns neither a user nor a conversation gets a 400 instead of trusting a client id. Tests cover the foreign-user 403, the 409, and the header/user path.
- **Quickstart**: fixed. The native handler reads the `x-opengeni-conversation` header the React component sends, so `resolve` returning tenant and user is enough; the example and docs match the code.
- **"Drop-in"**: reworded to "swap" and made truthful. The handler serves `GET` history, `OpenGeniChat` restores it on reload, and the Vercel and OpenAI adapters import the earlier messages from the request once as model context on the first message; after that OpenGeni owns the history. The docs say so.
- **Privacy qualification**: kept and made prominent: `agentAccess` governs agents; the organization key and console operators still read every session.
- **Splitting the PR**: kept as one change on request. The security fences are what make `agentAccess: "session"` safe to advertise, so they belong with it; the read-only organization key and org-wide listing are the separable part if a smaller review is wanted.

benyhh:
Both confirmed and fixed in the latest commit:

- **Identifier collision**: the session id is now the v5 UUID of the JSON tuple `[workspaceId, source, userId, conversation]` (or `[workspaceId, conversation]` without a user), so `alice:team` + `thread` and `alice` + `team:thread` derive different sessions; the idempotency key is `chat:<sessionId>`. `og.chat()` also re-verifies an existing session's end-user label against the resolved user before handing it back (`conversation_not_authorized`), in addition to the `previous_response_id` path. Tests cover the exact pair from the review, a source shift, quote injection, and the reopen case.
- **Pagination**: pinned rows now count toward the page limit. A boundary inside the pinned prefix resumes via `pinnedOffset` in the outer cursor, and the ordinary keyset page is re-read at the remaining budget so its cursor lands after the last returned row. A real-database test pins three sessions and walks limits of 1, 2, and 4 with no overrun, duplicates, or losses.

Agreed on waiting for completed CI before merge.

PR #2350, opened 2026-09-09
fix(react): keep answers visible when input waits settle

benyhh:
A streamed answer followed by wait_for_input could disappear inside collapsed steps when the turn completed with empty output. Keep the latest existing substantive answer outside the fold for that explicit wait settlement, preferring a completed answer over later commentary. Preserve exact text, final-output precedence, turn boundaries, and the existing wait notice.

Regression coverage includes delta-only answers, completed answers, later commentary, whitespace/opaque tails, authoritative output, ordinary turns, and foreign/user boundaries. The real MessageTimeline DOM test failed before the fix. Production Vite-build Chromium checks pass for both streamed and completed-commentary waits across collapse/expand.

Validation: projection + DOM suites 262 tests / 769 assertions passed, including all five final DOM variants. React typecheck, focused lint, formatting, and docs-reference/architecture guards pass. Production Chromium: 2 tests / 20 assertions pass. Includes a patch changeset for @opengeni/react.

Tracking: https://linear.app/cloudgeni/issue/OPE-468

### DIFF
--- a/.changeset/visible-streamed-wait-answer.md
+++ b/.changeset/visible-streamed-wait-answer.md
@@ -1,0 +1,5 @@
+---
+"@opengeni/react": patch
+---
+
+Keep the latest streamed answer visible outside collapsed steps when an input wait ends a turn without final output.
--- a/packages/react/demo/timeline-held-turn-test-harness.tsx
+++ b/packages/react/demo/timeline-held-turn-test-harness.tsx
@@ -22,9 +22,12 @@ function event(type: string, payload: unknown): SessionEvent {
 }
 
 const fallback = "The child is still running; I will resume when it finishes.";
+const streamed = new URLSearchParams(location.search).get("streamed") === "1";
 const events = [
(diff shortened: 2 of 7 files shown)


benyhh:
Independent reviewer approved exact head `cd7f80829657a89b285e7745a81a254b2500c041` with no actionable findings after rerunning the projection and actual DOM suites: 262 tests, 769 assertions passed. Completed-answer precedence and whitespace/opaque-tail cases are covered; the explicit wait boundary and same-turn suffix preserve ordinary turns and user/foreign-turn boundaries. Production Vite-build Chromium verification also passed (2 tests, 20 assertions), including collapsed/expanded visibility exactly once.

Parent source review agrees. Fresh merge-tree against main `befd38946f2b34eea9d86136bb0725f2a4b66ecf` is conflict-free, with no intervening changes to the candidate timeline paths. Source remains unchanged while CI runs.

PR #2375, opened 2026-09-09
test: make credential heartbeat deadline deterministic

jorgensandhaug:
Main CI can expire the test's one-millisecond credential lease before the mocked heartbeat begins, leaving its resolver undefined. Control the monotonic clock, assert the heartbeat started, then advance past the original deadline before returning success. The test continues to require rejection of that late success; production code is unchanged.

Validation: all217 agent-turn tests pass (823 assertions), formatting check passes, independent review passed.

## Summary by Sourcery

Make the credential heartbeat deadline-fence test deterministic without changing production behavior.

Enhancements:
- Make the credential heartbeat deadline-fence test deterministic by controlling monotonic time and verifying the heartbeat starts before advancing beyond the lease deadline.

Tests:
- Stabilize the late-heartbeat rejection test against CI timing variability while preserving its deadline-fence assertion.

### DIFF
--- a/apps/worker/test/agent-turn.test.ts
+++ b/apps/worker/test/agent-turn.test.ts
@@ -4408,6 +4408,8 @@ describe("Codex credential lease deadline fence", () => {
   });
 
   test("does not accept a successful heartbeat that returns after the prior deadline", async () => {
+    let now = performance.now();
+    const clock = spyOn(performance, "now").mockImplementation(() => now);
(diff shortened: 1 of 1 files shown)


benyhh (approved):
Reviewed the exact merged source now live on staging for expedited production promotion. The post-version change is a bounded deterministic test fix; exact source CI is green.

